PR #1, opened 2016-08-18
Rename mageia pubkey to RPM-GPG-KEY-Mageia

Conan-Kudo:
This is consistent with the file name in mageia-repos-keys package and the one previously used.

This PR depends on xsuchy/distribution-gpg-keys#1


xsuchy:
But we need to specify version in Requires in spec too. But I will do that in separate commit.

PR #4, opened 2016-09-19
fix Line-too-long

RadimS97:


scop:
What value did you use as the max line length? Note that line length messages are deliberately being ignored at the moment in various tools, see setup.cfg.

py/mock.py
```diff
@@ -23,7 +23,8 @@
 from __future__ import print_function
 """
 usage:
-       mock [options] {--init|--clean|--scrub=[all,chroot,cache,root-cache,c-cache,yum-cache,dnf-cache,lvm]}
+       mock [options] {--init|--clean|
+       --scrub=[all,chroot,cache,root-cache,c-cache,yum-cache,dnf-cache,lvm]}
```
scop: This does not look desirable as it will also affect line lengths of output.

py/mock.py
```diff
@@ -26,7 +26,8 @@
        mock [options] {--init|--clean|
        --scrub=[all,chroot,cache,root-cache,c-cache,yum-cache,dnf-cache,lvm]}
        mock [options] [--rebuild] /path/to/srpm(s)
-       mock [options] --buildsrpm {--spec /path/to/spec --sources /path/to/src|--scm-enable [--scm-option key=value]}
+       mock [options] --buildsrpm {--spec /path/to/spec 
+       --sources /path/to/src|--scm-enable [--scm-option key=value]}
```
scop: Again, this does not look desirable as it will also affect line lengths of output.

py/mock.py
```diff
@@ -126,7 +126,8 @@ def command_parse():
     parser.add_option("--clean", action="store_const", const="clean",
                       dest="mode",
                       help="completely remove the specified chroot")
-    scrub_choices = ('chroot', 'cache', 'root-cache', 'c-cache', 'yum-cache', 'dnf-cache', 'lvm', 'all')
+    scrub_choices = ('chroot', 'cache', 'root-cache', 'c-cache', 'yum-cache',\
+    'dnf-cache', 'lvm', 'all')
```
scop: The 2nd line should be indented, and I don't think the backslash is needed.

py/mock.py
```diff
@@ -270,7 +270,8 @@ def command_parse():
                       "Implies --no-clean. Valid options: build, install, binary")
     parser.add_option("--rpmbuild-opts", action="store",
                       help="Pass additional options to rpmbuild")
-    parser.add_option("--enablerepo", action="callback", type="string", dest="enable_disable_repos", default=[],
+    parser.add_option("--enablerepo", action="callback", type="string",\
+    dest="enable_disable_repos", default=[],
```
scop: The 2nd line should be indented, and I don't think the backslash is needed.

py/mock.py
```diff
@@ -274,7 +274,8 @@ def command_parse():
     dest="enable_disable_repos", default=[],
                       help="Pass enablerepo option to yum/dnf", metavar='[repo]',
                       callback=repo_callback)
-    parser.add_option("--disablerepo", action="callback", type="string", dest="enable_disable_repos", default=[],
+    parser.add_option("--disablerepo", action="callback", type="string",\
+    dest="enable_disable_repos", default=[],
```
scop: The 2nd line should be indented, and I don't think the backslash is needed.

py/mock.py
```diff
@@ -369,7 +369,8 @@ def setup_logging(config_path, config_opts, options):
     try:
         if not os.path.exists(log_ini):
             if os.path.normpath('/etc/mock') != os.path.normpath(config_path):
-                log.warning("Could not find required logging config file: %s. Using default...", log_ini)
+                log.warning("Could not find required logging config file: "\
+                "%s. Using default...", log_ini)
```
scop: The 2nd line should be indented, and I don't think the backslash is needed. Also, I think a better place to split would be the comma before log_ini.

py/mock.py
```diff
@@ -390,7 +390,8 @@ def setup_logging(config_path, config_opts, options):
 
     try:
         # set up logging format strings
-        config_opts['build_log_fmt_str'] = log_cfg.get("formatter_%s" % config_opts['build_log_fmt_name'], "format", raw=1)
+        config_opts['build_log_fmt_str'] = log_cfg.get("formatter_%s"\
+        % config_opts['build_log_fmt_name'], "format", raw=1)
```
scop: The 2nd line should be indented, and I don't think the backslash is needed. At least when splitting on the comma before "format" which would look like a better place for the split.

xsuchy:
> What value did you use as the max line length?

I created pylintrc (it is in git root) And I set the line length to 120 (raised from the default 100). I made survey between several colleagues and 120 characters should fit most modern terminals run in DM. We can discuss if we can lower the limit, but anything longer than 120 is definitely too much.

scop:
> I created pylintrc (it is in git root) And I set the line length to 120

Fine with me, setup.cfg updated accordingly for flake8/pep8/pycodestyle.

xsuchy:
All issues are resolved.
Merged. Thank you.

PR #9, opened 2016-10-11
add config option `build_as_root`

pkit:
new config option that does not drop to unprivileged user at build time
helps to build some legacy packages that expect to run build tests as root

### DIFF
--- a/py/mockbuild/buildroot.py
+++ b/py/mockbuild/buildroot.py
@@ -186,17 +186,17 @@ class Buildroot(object):
         env = dict(self.env)
         if nosync and self.nosync_path:
             env['LD_PRELOAD'] = self.nosync_path
-        if util.USE_NSPAWN:
+        if util.USE_NSPAWN or kargs.get('uid', -1) == 0:
             if 'uid' not in kargs:
                 kargs['uid'] = uid.getresuid()[1]
             if 'gid' not in kargs:
                 kargs['gid'] = uid.getresgid()[1]
             if 'user' not in kargs:
-                kargs['gid'] = pwd.getpwuid(kargs['uid'])[0]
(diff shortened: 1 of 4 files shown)


py/mockbuild/buildroot.py
```diff
@@ -175,17 +175,17 @@ def doChroot(self, command, shell=False, nosync=False, *args, **kargs):
         env = dict(self.env)
         if nosync and self.nosync_path:
             env['LD_PRELOAD'] = self.nosync_path
-        if util.USE_NSPAWN:
+        if util.USE_NSPAWN or kargs.get('uid', 1000) == 0:
```
xsuchy: Why is this here?

xsuchy: I meant - why are you testing for uid == 1000? The uid of user can be different. E.g. 1005. I understand the change in backend.py. But this one I cannot comprehend.

xsuchy:
Can you add this option to etc/mock/site-config.cfg (as comment) - with comment what it does? And add code to utils.py which will set it to defaults.

py/mockbuild/buildroot.py
pkit: It's just a placeholder. I could probably use `-1` here as well.
If we elevated to root `kargs['uid'] == 0` otherwise kargs may not even have 'uid' key (yes, it does happen).
But if `-1` is better, I can change it.

pkit:
Fixed the following:

- placeholder uid
- utils
- config example and comments

xsuchy:
This has been obsoleted by https://github.com/rpm-software-management/mock/pull/193 which allows you to build as root now. Thank you for this PR anyway.

PR #10, opened 2016-10-11
add config option `hostname` to set hostname

pkit:
adds ability to set a hostname for a build
hostname is a part of the resulting rpm in most of the cases
makes builds more reproducible

### DIFF
--- a/py/mock.py
+++ b/py/mock.py
@@ -655,6 +655,9 @@ def main():
     # New namespace starting from here
     unshare_namespace()
 
+    if config_opts['hostname']:
+        util.sethostname(config_opts['hostname'])
+
     # set personality (ie. setarch)
     if config_opts['internal_setarch']:
         util.condPersonality(config_opts['target_arch'])
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -47,6 +47,8 @@ _libc.personality.argtypes = [ctypes.c_ulong]
 _libc.personality.restype = ctypes.c_int
 _libc.unshare.argtypes = [ctypes.c_int]
 _libc.unshare.restype = ctypes.c_int
+_libc.sethostname.argtypes = [ctypes.c_char_p, ctypes.c_int]
+_libc.sethostname.restype = ctypes.c_int
 
 # See linux/include/sched.h
 CLONE_NEWNS = 0x00020000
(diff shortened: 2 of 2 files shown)


xsuchy:
I am not sure that changing the actual hostname is the way to go. I would rather see passing it to rpmbuild using variable. I add example in commit e8c3b32.
See https://bugzilla.redhat.com/show_bug.cgi?id=1302040 and https://bugzilla.redhat.com/show_bug.cgi?id=1309367 for more details.

pkit:
Are you sure that that rpmbuild option exists on all the centuries old RH Linux versions that we use mock for?

xsuchy:
It is not passed to rpmbuild by command line parameters, but by variable. So it will not produce error on E.g. EL5, however only F25+ recognize varibale %_buildhost.

pkit:
Maybe I do not understand the answer but what should I do in case where:
1. I have rpm for EL5
2. It's legacy RPM, i.e. I cannot change it.
3. I want to build it using mock (because the whole point of mock is to not use "real" EL5 machines for build)
4. How fo I run mock? How do I pass BUILDHOST to EL5 rpmbuild?

scop:
IIRC I managed to set the hostname the way I wanted a long time ago by tweaking /etc/hosts in the buildroot. The exact details escape me though. Something like this in site-defaults.cfg? Doesn't seem to work as is, but it's a start...?

```python
config_opts['use_host_resolv'] = False
config_opts['files']['etc/hosts'] = """
127.0.0.1 buildhost.localdomain
"""
```

xsuchy:
> but what should I do in case where:

With --old-chroot you can change your hostname using `hostname` command prior running mock. With --new-chroot there will be machine_id of the container. And yes right now there is not way to change it (but the @scop approach) on EL5.

My only point is that your code will change the hostname of the *host*. For all users on that machine. This is something which people will likely not expect.

IMHO it is just fair to say that for F24- and EL5, El6, EL7 it is not possible to change hostname and that you can only change it for F25+ targets.

pkit:
I beg your pardon?
`unshare_namespace()` that runs directly before that will unshare UTS (domainname, hostname, etc. from the rest of the system).

pkit:
Bump. Guys, we currently use wrapper script for `mock` to set hostname (it just does `unshare(CLONE_NEWUTS); sethostname();` ). And it works like a charm.
Considering that mock already unshares its context from the host and already does it for uts why not to incorporate that directly instead of using workarounds?

xsuchy:
OK. I am willing to merge it. But can you please add example with comments to etc/mock/site-defaults.cfg. It would be nice to have it close to use_container_host_hostname.

Frodox:
Come here to write same idea - how about using 'UTS namespace' for changing hostname inside mock ? 
I just test it also, and it works fine independently of rpmbuild's version (and IMHO it is like it should be).

Frodox:
+1 for examples and docs

Frodox:
bump @pkit

xsuchy:
I merged this and added the example there myself. Thank you for contributing.

PR #13, opened 2016-10-24
Delay mounting of user-defined mountpoints (rhbz#1386544)

msimacek:
Splits mounts into two categories:
- mock managed mountpoints, which are treated the same as before
- user managed mountpoints, which are mounted as the last initialization step
  (even after hooks), to make sure that they won't be affected by cleanup
  routines

bind_mount plugin and mount plugin now use user mounts, everything else uses
managed mounts


xsuchy:
Looks good to me. Thank you. Merged.

PR #18, opened 2016-11-24
macros: don't let host's %_install_langs affect guest chroots

praiskup:
Background story:  Fedora Cloud (and Docker) kickstarts put
basically only en_US.utf8 into %_install_langs, which then affects
mock's chroots -> the configuration is inherited from host's
config.

This affected Internal Copr builders which are built on top of
official Fedora Cloud images, but I believe this should be fixed
for everybody.

### DIFF
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -879,6 +879,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['macros'] = {
         '%_topdir': '%s/build' % config_opts['chroothome'],
         '%_rpmfilename': '%%{NAME}-%%{VERSION}-%%{RELEASE}.%%{ARCH}.rpm',
+        '%_install_langs': 'all',
     }
     # security config
     config_opts['no_root_shells'] = False


xsuchy:
After face2face disqussion with @praiskup: This does not work as intendend and it needs some work. So putting on back burner for now.

james-antill:
Should just be able to set override_install_langs=all in the yum configuration.

praiskup:
@james-antill, is that what we want in mock's configuration ("yum.conf" part)?
@xsuchy raised a good point that fixing (workaround) this in mock doesn't sound like natural fix -- some fix on lower level would look much better.

The point is that the host's rpm "configuration" (macros in this case) is not completely isolated from guest's (mock chroot) rpm configuration.  Haven't had a deeper look at it yet, but could this be guaranteed by RPM itself?

james-antill:
> @james-antill, is that what we want in mock's configuration ("yum.conf" part)?

Yes, you can put that in mock's config. to override the host changing the rpm macros.

Should you put it in the default mock config. files? Not sure, it's better than doing anything to try to override rpm changes in the mock code and it's harmless if people don't change the host rpm macros ... but in general people shouldn't alter the host rpm macros IMO.

praiskup:
ATM I'm not interested in fixing this, feel free to use the patch arbitrarily when needed.

PR #20, opened 2016-11-24
Add bash completion for .cfg files outside /etc/mock

jwakely:
The -r option accepts pathnames of files with .cfg extensions,
so make the bash completion also expand suitable files and
directories.

### DIFF
--- a/etc/bash_completion.d/mock
+++ b/etc/bash_completion.d/mock
@@ -4,6 +4,7 @@ _mock_root()
 {
     COMPREPLY+=( $( compgen -W "$( command ls ${1:-/etc/mock} 2>/dev/null | \
         sed -ne 's/\.cfg$//p' )" -X site-defaults -- "$cur" ) )
+    COMPREPLY+=( $( compgen -f -o plusdirs -X '!*.cfg' -- "$cur" ) )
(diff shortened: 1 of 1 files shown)


etc/bash_completion.d/mock
```diff
@@ -4,6 +4,8 @@ _mock_root()
 {
     COMPREPLY+=( $( compgen -W "$( command ls ${1:-/etc/mock} 2>/dev/null | \
         sed -ne 's/\.cfg$//p' )" -X site-defaults -- "$cur" ) )
+    COMPREPLY+=( $( compgen -W "$( command ls -d "$cur"*{.cfg,/} 2>/dev/null | \
+        sed 's:/$::' )" -X site-defaults -- "$cur" ) )
```
scop: I wonder if this addition could be simplified to just:

```bash
COMPREPLY+=( $( compgen -f -o plusdirs -X '!*.cfg' -- "$cur" ) )
```

Or am I missing something here?

jwakely: You're probably right, I've never used compgen before. I'll test it and re-push.

scop:
Thanks!

PR #23, opened 2016-11-28
fix logic to remove all tmpdirs on mock exit

Frodox:
Currently there appears extra one mock's dir in /tmp after every build.
```bash
$ ls -l /tmp
drwxrwxr-x. 3 builder builder 4096 nov 28 17:13 tito
drwx------. 2 root    root      20 nov 28 14:25 tmux-0

$ mock -r epel-7-x86_64 --resultdir RPMS/ acl-2.2.51-12.el7.src.rpm
...
Start: clean chroot
Finish: clean chroot
Finish: run

$ ls -l /tmp
drwxrwxr-x. 3 builder builder 4096 nov 28 17:13 tito
drwx------. 2 root    root      20 nov 28 14:25 tmux-0
drwxrwxrwx. 2 root    root    4096 nov 19 11:07 tmpzZkwvm
```

and it is because [this line](https://github.com/rpm-software-management/mock/blob/devel/py/mockbuild/buildroot.py#L535) is always false at the end (dir already removed). And it is also miss-sync with comments in code.

### DIFF
--- a/py/mockbuild/buildroot.py
+++ b/py/mockbuild/buildroot.py
@@ -494,7 +494,7 @@ class Buildroot(object):
     @traceLog()
     def _setup_nosync(self):
         multilib = ('x86_64', 's390x')
-        self.tmpdir = tempfile.mkdtemp()
+        self.tmpdir = tempfile.mkdtemp(prefix="tmp.mock.")
         os.chmod(self.tmpdir, 0o777)
         tmp_libdir = os.path.join(self.tmpdir, '$LIB')
         mock_libdir = self.make_chroot_path(tmp_libdir)
@@ -532,7 +532,7 @@ class Buildroot(object):
         """
         Do the cleanup if this is the last process working with the buildroot.
         """
-        if os.path.exists(self.make_chroot_path()):
+        if not os.path.exists(self.make_chroot_path()):
             try:
                 if self.tmpdir:
                     for d in self.tmpdir, self.make_chroot_path(self.tmpdir):


xsuchy:
Hmm, this part of code is a bit magic to me. finalize() was added in e3c5267a0fc6827520bcb5a91c53fe2bdf15a363 by @msimacek 

I am trying to read the code and I do not understand why the line:
  `if os.path.exists(self.make_chroot_path()):`
is there at all. The comment
  `Do the cleanup if this is the last process working with the buildroot.`
does not help at all. Under which condition is should NOT be last process. There is only one process.
@msimacek do you want to comment it?

msimacek:
There are commands which do not require exclusive lock of the buildroot - for example, you can use `--shell` while a build is running - and either of them can exit independently, but only last one of them should do the cleanup.

xsuchy:
But in such situatio, the `self.tmpdir` should be deleted always, because each instance has its own copy. Right?

msimacek:
Yes, that's true, the `tmpdir` is deleted regardless of whether it's the last process or not, so the docstring is not acurate.
The correct solution to the problem should be to move the `if self.tmpdir` block outside of `if os.path.exists(self.make_chroot_path())`, but the condition there should not be changed - it's not always false - it's false only when you execute commands that delete the buildroot.

Frodox:
@msimacek could you fix it, please? I'm not clear, what should I do to fix the issue with un-deleted `self.tmpdir`

msimacek:
Ok, I'll submit a PR

xsuchy:
I manually applied first part of your commit as that still makes sense. Commit 827f980

PR #24, opened 2016-11-28
fix more_buildreq logic a bit

Frodox:
Clarify docs-examples and fix logic in case of just one dependency.

`isinstance(type(more_reqs), basestring)` is always `False`
```python
>>> isinstance(type("test"), basestring)
False
```
So, if I have lines like `config_opts['more_buildreqs']['acl'] = 'ruby-devel'`, I will get
```bash
$ mock -r epel-7-x86_64 --resultdir RPMS/ acl-2.2.51-12.el7.src.rpm
...
Start: clean chroot
Finish: clean chroot
ERROR: Error: No packages found for:
  -
  b
  d
  e
  l
  r
  u
  v
  y
```
:D 

Also, it is not clear from site-defaults, how to use 2+ deps.

### DIFF
--- a/etc/mock/site-defaults.cfg
+++ b/etc/mock/site-defaults.cfg
@@ -279,7 +279,8 @@
 # config_opts['chroot_setup_cmd'] = 'install buildsys-build'
 # config_opts['chroot_additional_packages'] = ''
 # config_opts['log_config_file'] = 'logging.ini'
-# config_opts['more_buildreqs']['srpm_name-version-release'] = 'dependencies'
+# config_opts['more_buildreqs']['srpm_name-version-release'] = 'dependency'
+# config_opts['more_buildreqs']['srpm_name-version-release'] = ['dependency1', 'dependency2']
 # config_opts['macros']['%Add_your_macro_name_here'] = "add macro value here"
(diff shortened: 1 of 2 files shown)


xsuchy:
Nice. I did not know we had such option. Note that if you want to use some buildreq for whole chroot it is better to use config_opts['chroot_additional_packages']. See https://github.com/rpm-software-management/mock/wiki/Release-Notes-1.2.19

xsuchy:
Merged. Thank you.

Frodox:
> Nice. I did not know we had such option. 

I'm using it time to time. It is useful when some maintainer miss some important dependency :)

> Note that if you want to use some buildreq for whole chroot it is better to use  config_opts['chroot_additional_packages']`

What is the different with `config_opts['chroot_setup_cmd'] += 'some_package other_package'` ?
I'm using it now like this and it works fine :-)

xsuchy:
>What is the different with config_opts['chroot_setup_cmd'] += 'some_package other_package' ?
>I'm using it now like this and it works fine :-)

Funcrionaly? None difference. It is just better for automated tools, which injects code in config on the fly.

PR #29, opened 2017-01-02
Fix local repo URLs in configs

mizdebsk:
- Update local repo URLs for rawhide - since Fedora 26, secondary architectures were merged into primary Koji.
- Switch kojipkgs URLs to https - since 12 Dec 2016 kojipkgs.fedoraproject.org has a valid TLS certificate, so there is no longer any reason not to use https.


xsuchy:
Merged. Thank you.

PR #30, opened 2017-01-04
Fix debuginfo repo naming

msimacek:
dnf debuginfo-install needs the debug repo to be named
${repo}-debuginfo, where ${repo} is the name of the main repo.
Otherwise, it doesn't find it.


xsuchy:
Merged. Thank you.

PR #35, opened 2017-01-27
Support nspawn_args

cgwalters:
This is a generalization of the `--rpmbuild-networking` code; for projects which
want to use mock as a generic container tool, this allows configuring nspawn,
which has most of the useful container features exposed already.

### DIFF
--- a/etc/mock/site-defaults.cfg
+++ b/etc/mock/site-defaults.cfg
@@ -64,6 +64,8 @@
 # If you're using nspawn, then by default networking will be turned off
 # for rpmbuild.  This helps ensure more reproducible builds.
 # config_opts['rpmbuild_networking'] = False
+# Additional args for nspawn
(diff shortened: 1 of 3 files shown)


cgwalters:
Side note: it took me a while to figure out that `--shell` has a special case that needed bridging via `doshell()`, not just `doChroot()`.; when testing I was doing `capsh --print` and being confused why it wasn't working.

Down the line, I think it would make more sense for the chroot code to be able to read the config; then we wouldn't need to schlep so much config data around.

xsuchy:
Merged. Thank you.

Frodox:
@cgwalters yup, I also noticed that `--shell` with `nspawn` mixes stdout and stderr of command into one stdout, which is very uncomfortable for `generic container tool` (I used it for that too).

Would `doshell()` fix it..?

PR #36, opened 2017-01-28
systemd-nspawn: run as PID2

ignatenkobrain:
Since we don't run systemd or any process which implements sysvinit
compatible signal handling (specifically: it needs to reboot on SIGINT,
reexecute on SIGTERM, reload configuration on SIGHUP, and so on) we
should use -a/--as-pid2 to run STUBINIT as PID1 and our process as PID2.

rpmbuild doesn't implement sysvinit compatible signal handling (e.g.
it doesn't kill child processes, so builds can stuck because of this).

Reported-by: Vít Ondruch <vondruch@redhat.com>
References: https://bugzilla.redhat.com/show_bug.cgi?id=1372234
Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>

### DIFF
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -631,7 +631,7 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=[], env=None, cwd
             raise exception.Error('Internal Error: command must be list or shell=True.')
     elif not cmd_is_list:
         cmd = [cmd]
-    nspawn_argv = ['/usr/bin/systemd-nspawn', '-q', '-M', uuid.uuid4().hex, '-D', chrootPath]
+    nspawn_argv = ['/usr/bin/systemd-nspawn', '-q', '-a', '-M', uuid.uuid4().hex, '-D', chrootPath]
     nspawn_argv.extend(nspawn_args)
     if cwd:
         nspawn_argv.append('--chdir={0}'.format(cwd))


ignatenkobrain:
Related issue in RPM: https://github.com/rpm-software-management/rpm/issues/134

xsuchy:
Very nice work. Thank you.
However there is one problem. On RHEL7 (which mock support) there is systemd-219. And it does not have this option. :(

ignatenkobrain:
@xsuchy probably we shouldn't use sd-nspawn on such ancient distros like RHEL7?

xsuchy:
I can imagine that we add 
config_opts['systemd_higher_than_X'] = True
Set this in config only on Fedora.
And in this is set we
  config_opts['nspawn_args']  += ['-a']

ignatenkobrain:
@xsuchy but this problem will not be fixed on EL7 anyway...

xsuchy:
Yes, this particulart problem will not be fixed on EL7, but it will does not block you from using nspawn on EL7. And EL7 is latest available enterprise platform. There is nothing newer. So I care about it. And we should. So I wanto to support nspawn on EL7.

lnykryn:
If you want we can check how easily would be to backport --as-pid2 to el7. Nspawn is quite isolated piece of code so it should not be that hard.
Or just write your own init :-)

ignatenkobrain:
@lnykryn would be nice to backport ;)

Conan-Kudo:
@lnykryn Please just backport it, for all of our sakes...

xsuchy:
Reported to systemd @ EL7
https://bugzilla.redhat.com/show_bug.cgi?id=1417387

voxik:
Backup plan could be to go with dumb-init, which is available in Fedora (but not in RHEL/EPEL) ...

lnykryn:
Or maybe just run /bin/bash -c "rpmbuild" in the container. Bash works as a pid1 just fine.

voxik:
@lnykryn It seems that the issue I am facing with mongodb is due to signal handling. Bash does not help with this issue ....

cgwalters:
This patch makes sense to me.  One thing we could do is have it be detected at build time - if mock is built for Fedora, use it.

cgwalters:
Side fact, this would also fix https://pagure.io/releng/issue/6602#comment-72010 - we're likely to handle it by injecting it using the new `nspawn_opts` argument.

py/mockbuild/util.py
```diff
@@ -631,7 +631,7 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=[], env=None, cwd
             raise exception.Error('Internal Error: command must be list or shell=True.')
     elif not cmd_is_list:
         cmd = [cmd]
-    nspawn_argv = ['/usr/bin/systemd-nspawn', '-q', '-M', uuid.uuid4().hex, '-D', chrootPath]
+    nspawn_argv = ['/usr/bin/systemd-nspawn', '-q', '-a', '-M', uuid.uuid4().hex, '-D', chrootPath]
```
dustymabe: would it be worth adding a comment about why we are running as pid2? could be useful for future contributors to the code.

dustymabe: there is the commit message as well, so that will suffice if the comment is unwanted

ignatenkobrain: yes, it's written in commit message

xsuchy:
Merged as c572428
On rhel and centos this option is not added. This can be revisited when BZ 1417387 is resolved.

PR #38, opened 2017-02-14
enable package_state plugin by default (again) [RHBZ#1277187]

Frodox:
since it wasn't fully enabled in 264597b8a17826362a839715fd7502cbee1ea9a0

### DIFF
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -813,7 +813,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
             'keep_mounted': False},
         'selinux_enable': True,
         'selinux_opts': {},
-        'package_state_enable': False,
+        'package_state_enable': True,
         'package_state_opts': {
             'available_pkgs': False,
             'installed_pkgs': True,


Frodox:
How about renaming also `installed_pkgs` to `installed_pkgs.log` ? To be more consistent

xsuchy:
+1 to rename
-1 to enable it be default. Any reason for that?

Frodox:
Ok, I'll propose a PR to rename, ok?

No reason to enable it by default, since every user can enable it manually. Docs (wiki, site-defaults.cfg) says it is enabled by default. When I wanted to fix docs I see, that you already "enable it" for BZ#1277187 so I do this.

In constrast, it should be package_state=False in site-defaults.cfg


P.s. also as I noticed, 'Plugin Mount's not in `PLUGIN_LIST` - is it important? There are no any comments about neither plugin-list nor plugin-mount..

xsuchy:
> Ok, I'll propose a PR to rename, ok?

That would be great.

Frodox:
Ok, what plans for this now?
There is 'out of sync' between
* [changelog](https://github.com/rpm-software-management/mock/blob/ee184b861fc2b65c3b61cbca7b39d4a378af21d9/mock.spec#L331)
* [site-defaults](https://github.com/rpm-software-management/mock/blob/ee184b861fc2b65c3b61cbca7b39d4a378af21d9/etc/mock/site-defaults.cfg#L130)
* [actual code defaults](https://github.com/rpm-software-management/mock/blob/ee184b861fc2b65c3b61cbca7b39d4a378af21d9/py/mockbuild/util.py#L816)

Should I revert 264597b instead..?

xsuchy:
Reading again throu this. I suppose that what I wrote in https://bugzilla.redhat.com/show_bug.cgi?id=1277187#c6 should be the correct state. And I see that in that original 264597b I forgot to enable it. So the "again" in commit description is little bit off. But that is detail. Merged. Thank you.

PR #41, opened 2017-02-16
add dist-git support into mock-scm

clime:
This pull request implements support for building from a dist-git repo into the mock-scm plugin.

Example usage:

    mock -r fedora-25-x86_64 --scm-enable --scm-option method=distgit --scm-option package=prunerepo --scm-option distgit_get='rpkg clone prunerepo' --scm-option distgit_src_get='rpkg sources'

Example config:

    config_opts['scm'] = True
    config_opts['scm_opts']['method'] = 'distgit'
    config_opts['scm_opts']['distgit_get'] = 'rpkg clone prunerepo'
    config_opts['scm_opts']['distgit_src_get'] = 'rpkg sources'
    config_opts['scm_opts']['package'] = 'prunerepo'
    config_opts['scm_opts']['branch'] = 'master'

### DIFF
--- a/etc/mock/site-defaults.cfg
+++ b/etc/mock/site-defaults.cfg
@@ -254,6 +254,8 @@
 # config_opts['scm_opts']['cvs_get'] = 'cvs -d /srv/cvs co SCM_BRN SCM_PKG'
 # config_opts['scm_opts']['git_get'] = 'git clone SCM_BRN git://localhost/SCM_PKG.git SCM_PKG'
 # config_opts['scm_opts']['svn_get'] = 'svn co file:///srv/svn/SCM_PKG/SCM_BRN SCM_PKG'
+# config_opts['scm_opts']['distgit_get'] = 'rpkg clone -a --branch SCM_BRN SCM_PKG SCM_PKG'
+# config_opts['scm_opts']['distgit_src_get'] = 'rpkg sources
 # config_opts['scm_opts']['spec'] = 'SCM_PKG.spec'
 # config_opts['scm_opts']['ext_src_dir'] = '/dev/null'
 # config_opts['scm_opts']['write_tar'] = True
--- a/etc/mock/site-defaults.cfg
+++ b/etc/mock/site-defaults.cfg
@@ -255,7 +255,7 @@
 # config_opts['scm_opts']['git_get'] = 'git clone SCM_BRN git://localhost/SCM_PKG.git SCM_PKG'
 # config_opts['scm_opts']['svn_get'] = 'svn co file:///srv/svn/SCM_PKG/SCM_BRN SCM_PKG'
 # config_opts['scm_opts']['distgit_get'] = 'rpkg clone -a --branch SCM_BRN SCM_PKG SCM_PKG'
-# config_opts['scm_opts']['distgit_src_get'] = 'rpkg sources
+# config_opts['scm_opts']['distgit_src_get'] = 'rpkg sources'
 # config_opts['scm_opts']['spec'] = 'SCM_PKG.spec'
 # config_opts['scm_opts']['ext_src_dir'] = '/dev/null'
 # config_opts['scm_opts']['write_tar'] = True
--- a/py/mockbuild/scm.py
+++ b/py/mockbuild/scm.py
@@ -90,9 +90,21 @@ class scmWorker(object):
         self.wrk_dir = tempfile.mkdtemp(".mock-scm." + os.path.basename(self.pkg))
         self.src_dir = self.wrk_dir + "/" + os.path.basename(self.pkg)
         self.log.debug("SCM checkout directory: %s", self.wrk_dir)
-        util.do(shlex.split(self.get), shell=False, cwd=self.wrk_dir, env=os.environ)
+        try:
+            util.do(shlex.split(self.get), shell=False, cwd=self.wrk_dir, env=os.environ)
+        except PermissionError as e:
+            self.log.error("{} does not exist or cannot be executed due permissions."
+                           .format(shlex.split(self.get)[0]));
+            sys.exit(5)
+
         for command in self.postget:
-            util.do(shlex.split(command), shell=False, cwd=self.src_dir, env=os.environ)
+            try:
+                util.do(shlex.split(command), shell=False, cwd=self.src_dir, env=os.environ)
+            except PermissionError as e:
+                self.log.error("{} does not exist or cannot be executed due permissions."
+                               .format(shlex.split(command)[0]));
+                sys.exit(5)
+
         self.log.debug("Fetched sources from SCM")
 
     @traceLog()
--- a/py/mockbuild/scm.py
+++ b/py/mockbuild/scm.py
@@ -28,6 +28,8 @@ class scmWorker(object):
         self.wrk_dir = self.src_dir = None
         self.sources = []
         self.macros = macros
+        self.branch = None
+        self.postget = []
 
         self.method = opts['method']
         if self.method == "cvs":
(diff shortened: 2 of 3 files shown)


xsuchy:
I installed package with this commit (`tito build --test --rpm -i`) and it gives me:
```
$ mock -r fedora-25-x86_64 --scm-enable --scm-option method=distgit --scm-option package=prunerepo --scm-option distgit_get='rpkg clone prunerepo' --scm-option distgit_src_get='rpkg sources'
INFO: mock.py version 1.3.3 starting (python version = 3.5.2)...
Start: init plugins
INFO: selinux disabled
Finish: init plugins
Start: run
  File "/usr/libexec/mock/mock", line 885, in <module>
    main()
  File "/usr/lib/python3.5/site-packages/mockbuild/trace_decorator.py", line 89, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 701, in main
    run_command(options, args, config_opts, commands, buildroot, state)
  File "/usr/lib/python3.5/site-packages/mockbuild/trace_decorator.py", line 89, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 720, in run_command
    scmWorker.get_sources()
  File "/usr/lib/python3.5/site-packages/mockbuild/trace_decorator.py", line 89, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.5/site-packages/mockbuild/scm.py", line 93, in get_sources
    util.do(shlex.split(self.get), shell=False, cwd=self.wrk_dir, env=os.environ)
  File "/usr/lib/python3.5/site-packages/mockbuild/trace_decorator.py", line 89, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.5/site-packages/mockbuild/util.py", line 530, in do
    preexec_fn=preexec,
  File "/usr/lib64/python3.5/subprocess.py", line 947, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.5/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg)
PermissionError: [Errno 13] Permission denied
```

xsuchy:
Can you put example in site-defaults.cfg under `# config_opts['scm_opts']['svn_get'] = 'svn co file:///srv/svn/SCM_PKG/SCM_BRN SCM_PKG'`?
And BTW `rpkg clone -a` would be better as that will work for everybody.

clime:
Add. the error: That error can also pop up if you don't have rpkg installed. Could you please look if rpkg client is installed on your system?

I will, meanwhile, work on the other two issues.

xsuchy:
Indeed I  do not have rpkg installed. Can you catch that exception and print user friendly error, that command $NAMEOFCOMMAND does not exist or cannot be executed due permissions?

clime:
That's one option but also we could just add rpkg to Requires for mock-scm as it is already done for cvs and git. Would that be also acceptable?

xsuchy:
But you can use `fedpkg` tooo.. And possibly not Requires but Suggests, and that is causing problems on el6...

clime:
I added the error message and also addressed the other two issues.

ausil:
I am kinda against the feature, at least calling it just distgit, fedora, centos inside Red Hat all have distgit's maybe fedgit fedora-distgit or something clearly defining it as fedora, adding a dep on fedpkg which would be needed also seams heavy to me.

clime:
> I am kinda against the feature, at least calling it just distgit, fedora, centos inside Red Hat all have distgit's maybe fedgit fedora-distgit or something clearly defining it as fedora, adding a dep on fedpkg which would be needed also seams heavy to me.

You can use different dist-git client. fedpkg dep wasn't added.

clime:
I will elaborate: instead of default rpkg, you can use fedpkg or rhpkg or something else. We just need to let user know that the tool needs to be installed on the system, which is now done by the error friendly message when the client tool is not installed. We can also mention it in wiki.

praiskup:
> instead of default rpkg, you can use fedpkg or rhpkg or something else. 

At least the configuration file looks pretty bound to one dist-git instance;  consider that people want to build against multiple dist-git instances (I typically build against copr dist-git, rhel's dist-git and fedora's dist-git) ..  Can I easily switch among them?  (somehow related is rhbz#1391899).

xsuchy:
> Can I easily switch among them?

Yes. Just pass: `--scm-option distgit_get='fedpkg clone -a SOMEPACKAGE'`

praiskup:
In config?

clime:
Note that you can use just rpkg for all Fedora DistGit, RedHat DistGit, and also for Fedora Copr DistGit with regard to the mock-scm plugin. You just need to set correct `lookaside` URL and `annongiturl` (or `gitbaseurl` and override the `-a` default option) in `rpkg.conf`. 

As a usability improvement, you can have more `rpkg` configs in `/etc/rpkg` and pass the appropriate one using `--config` option for `rpkg`. You need to override `config_opts['scm_opts']['distgit_get']` and `config_opts['scm_opts']['distgit_src_get']` for that of course. Example:

    config_opts['scm_opts']['distgit_get']` ='rpkg --config /etc/rpkg/rpkg-rh.conf clone -a --branch SCM_BRN SCM_PKG SCM_PKG'

For Fedora Copr DistGit, you additionally need to use argument `--module-name` for `'distgit_src_get'` to specify a triplet like `@copr/copr/copr-dist-git`.

But in the end, you can use just rpkg with mock-scm for all three DistGits!

xsuchy:
@Conan-Kudo any reason for that?

xsuchy:
My reasoning for being part of scm plugin is that it actually share a lot of code. If this will be another plugin, we will have to copy lots of code. Which will make maintenance harder.

Conan-Kudo:
@xsuchy If it's going to be part of the SCM plugin, then it shouldn't be restricted to just Git. There are plenty of systems like what Fedora, Mageia, and others use, and it would make more sense to make this so that it's more configurable.

For example, if I could use this with Koji or COPR to be able to check out from Mageia SVN the spec and sources then append the changelog generated from SVN to the spec, that would make this much more useful.

As sad as it sounds, other distro projects do still use SVN or even CVS. Maybe they use something different like Mercurial. The point is, this should be set up so that this is more generic.

xsuchy:
> As sad as it sounds, other distro projects do still use SVN or even CVS. Maybe they use something different like Mercurial. The point is, this should be set up so that this is more generic.

I don't get it. This plugin handle SVN and CVS for ages. See: https://github.com/rpm-software-management/mock/wiki/Plugin-Scm

Plugin-scm recognize 3 methods. This just add 4th method.

Conan-Kudo:
@xsuchy the SCM plugin currently doesn't handle dist-git or dist-svn (Fedora/Mageia style), only repos that have sources merged with spec (like Debian style package VCS).

This proposed addition enables Fedora-style dist-git, but isn't generic enough to support any kind of dist-scm style system.

xsuchy:
>, but isn't generic enough to support any kind of dist-scm style system.

How it comes? You just use different method in  distgit_get.

Conan-Kudo:
@xsuchy Oh? It already doesn't care whether it is Git, SVN, or whatever? Then call it distscm, instead. :)

etc/mock/site-defaults.cfg
```diff
@@ -254,6 +254,8 @@
 # config_opts['scm_opts']['cvs_get'] = 'cvs -d /srv/cvs co SCM_BRN SCM_PKG'
 # config_opts['scm_opts']['git_get'] = 'git clone SCM_BRN git://localhost/SCM_PKG.git SCM_PKG'
 # config_opts['scm_opts']['svn_get'] = 'svn co file:///srv/svn/SCM_PKG/SCM_BRN SCM_PKG'
+# config_opts['scm_opts']['distgit_get'] = 'rpkg clone -a --branch SCM_BRN SCM_PKG SCM_PKG'
+# config_opts['scm_opts']['distgit_src_get'] = 'rpkg sources
```
Conan-Kudo: Missing terminating `'`.

xsuchy:
Merged. Thank you.

PR #44, opened 2017-02-17
rename package_state plugin's log to have .log suffix

Frodox:
To be more consistent. So, we can do now smth like
```
mv tmp_dir/*.log log_place/build_id/
```

from #38


xsuchy:
Merged. Thank you.

PR #46, opened 2017-02-21
config: add best=1 also into rawhide configs

praiskup:
Per discussion on fedora devel, and bug-report against Copr.
Complements 015f5771.

Version: 1.3.3-1

### DIFF
--- a/etc/mock/fedora-rawhide-aarch64.cfg
+++ b/etc/mock/fedora-rawhide-aarch64.cfg
@@ -22,6 +22,7 @@ syslog_ident=mock
 syslog_device=
 install_weak_deps=0
 metadata_expire=0
+best=1
 
 # repos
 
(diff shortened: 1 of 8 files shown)


xsuchy:
+1, but....

I plan to do release in one week. And I am really afraid that this change may break Koji builds. So I will merge it *after* the release. Give it 2 months in git, so it can be tested by people who use nightlies. And if everything will be fine, it will be part of 1.3.5.

ignatenkobrain:
koji never has more than one version of package in buildroot

xsuchy:
OK. Then lets break it :)

xsuchy:
Merged.

ausil:
I am not a fan of this change. however it will have zero impact on koji as koji generates its own mock configs on the fly. people may experience different behaviour between koji and mock. If they do we have to be prepared to revert this change

Issue #48, opened 2017-02-24
mock init does not set correct permissions after unpacking root_cache


adamel:
When having multiple users using mock the ownership of the /builddir tree is not set to the current user after unpacking the root_cache when doing a simple mock init.

(This worked fine in mock 1.1.41 and earlier, so we've been using that up until now when I finally decided we do not want to keep using an ancient version forever.)

xsuchy:
Merged.

adamel:
Thanks. I'd love to see this go into the 1.3 branch (or more specifically; released for EPEL 7).

xsuchy:
It will be in next release (in ~2 months). And it will be released for EL7 too. But not for EL6.

PR #49, opened 2017-02-24
Ensure mockbuild UID and homedir permissions are correct after mock init from root_cache

adamel:
Fixes issue #48.

### DIFF
--- a/py/mockbuild/buildroot.py
+++ b/py/mockbuild/buildroot.py
@@ -162,6 +162,7 @@ class Buildroot(object):
                 self.pkg_manager.update()
                 self.state.finish(update_state)
         else:
+            self._fixup_build_user()
             # Change owner of homdir tree if the root of it not owned
             # by the current user.
             home = self.make_chroot_path(self.homedir)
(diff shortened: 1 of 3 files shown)


xsuchy:
The idea is good. The implementation is sub-optimal. You should use
```
  # maybe elevate privs here using self.buildroot.uid_manager and then
  self.buildroot.uid_manager.changeOwner(self.homedir, recursive=True)
```

adamel:
I can do that, but it will change the existing code in chown_home_dir to follow symlinks (uid_manager.changeOwner() uses os.chown() instead of of.lchown()) and symlinks might point to something outside the mock root.

Is there a reason for uid_manager._tolerant_chown() not using os.lchown() ?

xsuchy:
I am not aware of any such reason.

adamel:
Pushed updated branch. Also optimized it so we do not do a full recursive traverse of the entire homedir if it is already owned by the correct user.

py/mockbuild/buildroot.py
```diff
@@ -161,6 +161,13 @@ def _init(self, prebuild, do_log):
                 self.state.start(update_state)
                 self.pkg_manager.update()
                 self.state.finish(update_state)
+        else:
+            self._fixup_build_user()
```
xsuchy: Instead of this call self.uid_manager.dropPrivsTemp() and at the enf of else block self.uid_manager.restorePrivs()

adamel: What will that achieve? The purpose of _fixup_build_user() is (only) to modify the /etc/passwd file in the mock root, and as far as I can see dropPrivsTemp() does nothing like that?

xsuchy:
Ah, now I see what you are trying to achieve. Merged. Thank you.

Issue #45, opened 2017-02-20
ChrootScan plugin creates directory owned by root


Frodox:
Tested on RHEL7.3

So, is smth like
```bash
$ mock --resultdir zt /build/result/src/master/pkg-2.13-124.el7.src.rpm
$ ls -lh zt/
total 1.7M
drwxrwxr-x. 3 root    root      17 Feb 20 19:31 chroot_scan/
-rw-rw-r--. 1 builder builder  22K Feb 20 19:31 build.log
-rw-rw-r--. 1 builder builder 1.1M Feb 20 19:31 root.log
-rw-rw-r--. 1 builder builder  878 Feb 20 19:31 state.log
-rw-rw-r--. 1 builder mock     29K Feb 20 19:31 pkg-2.13-124.el7.noarch.rpm
-rw-rw-r--. 1 builder mock    323K Feb 20 19:31 pkg-2.13-124.el7.src.rpm
-rw-rw-r--. 1 builder mock    299K Feb 20 19:31 pkg-gui-2.13-124.el7.noarch.rpm

$ rm -rf zt/chroot_scan/
/bin/rm: cannot remove ‘zt/chroot_scan/’: Permission denied
```

And current user can't access/edit `result_dir/chroot_scan` now. What do you think about that?
How about change perms to user/mock one ?

Frodox:
Even more.. I can't see content =)
```
$ ls -lh zt/chroot_scan/var/lib/mock/os-x86_64/root/builddir/                                                                       
ls: cannot open directory zt/chroot_scan/var/lib/mock/os-x86_64/root/builddir/: Permission denied
```

Frodox:
perms are correct now (unprivUID, unprivGID), I can access file inside `chroot_scan/var/lib/mock/os-x86_64/root/builddir/ `
but I can't delete it because of
```
$ ls -l /tmp/tmp.mock-results-z-tools.dfXoao/chroot_scan/var/lib/mock/os-x86_64-local-build-one/
total 0
dr-xr-xr-x. 5 builder builder 44 фев 21 15:34 root
```

same 555 as in /var/lib/mock/rootname/root

need one more fix

xsuchy:
See commit 12e8682 which may be relevant.

PR #50, opened 2017-02-28
fix permissions in chroot_scan's result dir, so user can delete it

Frodox:
Some packages (filesystem) set 555 perms on /,
also, some dirs with needed data may be 555 only.
In such cases unprivUser can't delete/move resutls.

From #45


xsuchy:
Merged. Thank you.

PR #51, opened 2017-03-01
Bootstrap package manager using outer chroot

mich181189:
This commit changes Mock so that it first builds
an outer chroot, then uses that chroot to build the main
chroot. This means the chroot is built with the version of
dnf or yum it contains, allowing the package being built to
use newer RPM/dnf features. This means the only packages
that need to be installable by the old rpm/yum/dnf are those
required in this minimal environment.

This is intended as an implementation of [RHBZ#1118719] - at least for the SIG-Rust case

### DIFF
--- a/py/mockbuild/plugin.py
+++ b/py/mockbuild/plugin.py
@@ -15,8 +15,8 @@ class Plugins(object):
         self.config = config
         self._hooks = {}
         self.state = state
-        
-        self.initted = False
+
+        self.already_initialized = False
 
         self.plugins = config['plugins']
         self.plugin_conf = config['plugin_conf']
@@ -24,9 +24,9 @@ class Plugins(object):
 
     @traceLog()
     def init_plugins(self, buildroot):
-        if self.initted:
+        if self.already_initialized:
             return
-        self.initted = True
+        self.already_initialized = True
         for key in list(self.plugin_conf.keys()):
             if key.endswith('_opts'):
                 self.plugin_conf[key]['basedir'] = buildroot.basedir
--- a/py/mockbuild/plugin.py
+++ b/py/mockbuild/plugin.py
@@ -15,6 +15,8 @@ class Plugins(object):
         self.config = config
         self._hooks = {}
         self.state = state
+        
+        self.initted = False
 
         self.plugins = config['plugins']
         self.plugin_conf = config['plugin_conf']
@@ -22,6 +24,9 @@ class Plugins(object):
 
     @traceLog()
     def init_plugins(self, buildroot):
+        if self.initted:
+            return
+        self.initted = True
         for key in list(self.plugin_conf.keys()):
             if key.endswith('_opts'):
                 self.plugin_conf[key]['basedir'] = buildroot.basedir
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -748,6 +748,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['rpmbuild_networking'] = False
     config_opts['nspawn_args'] = []
     config_opts['use_container_host_hostname'] = True
+    config_opts['use_bootstrap_container'] = True
 
     config_opts['internal_dev_setup'] = True
     config_opts['internal_setarch'] = True
@@ -913,11 +914,11 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
 
     # configurable commands executables
     config_opts['yum_command'] = '/usr/bin/yum'
-    if os.path.isfile('/usr/bin/yum-deprecated'):
-        config_opts['yum_command'] = '/usr/bin/yum-deprecated'
+    config_opts['yum_install_command'] = 'install yum yum-utils shadow-utils distribution-gpg-keys'
 
     config_opts['yum_builddep_command'] = '/usr/bin/yum-builddep'
     config_opts['dnf_command'] = '/usr/bin/dnf'
+    config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core distribution-gpg-keys'
     config_opts['rpm_command'] = '/bin/rpm'
     config_opts['rpmbuild_command'] = '/usr/bin/rpmbuild'
 
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -983,6 +983,8 @@ def set_config_opts_per_cmdline(config_opts, options, args):
         config_opts['unique-ext'] = options.uniqueext
     if options.rpmbuild_timeout is not None:
         config_opts['rpmbuild_timeout'] = options.rpmbuild_timeout
+    if options.bootstrapchroot is not None:
+        config_opts['use_bootstrap_container'] = options.bootstrapchroot
 
     for i in options.disabled_plugins:
         if i not in config_opts['plugins']:
--- a/py/mockbuild/util.py
+++ b/py/mockbuild/util.py
@@ -914,10 +914,12 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
 
     # configurable commands executables
     config_opts['yum_command'] = '/usr/bin/yum'
+    config_opts['system_yum_command'] = '/usr/bin/yum'
     config_opts['yum_install_command'] = 'install yum yum-utils shadow-utils distribution-gpg-keys'
 
     config_opts['yum_builddep_command'] = '/usr/bin/yum-builddep'
     config_opts['dnf_command'] = '/usr/bin/dnf'
+    config_opts['system_dnf_command'] = '/usr/bin/dnf'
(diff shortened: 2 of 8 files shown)


Conan-Kudo:
@mich181189 I don't see an ability to trigger this behavior via a command-line switch. Could you add the ability to do so?

mich181189:
I think I probably could do. I didn't initially because the goal was to make it the default on at least some targets (I'm not sure there's much benefit to it being default on older distros, but for newer ones it actually has to be default if I understand correctly to be really useful)

py/mockbuild/util.py
```diff
@@ -913,11 +914,11 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
 
     # configurable commands executables
     config_opts['yum_command'] = '/usr/bin/yum'
-    if os.path.isfile('/usr/bin/yum-deprecated'):
-        config_opts['yum_command'] = '/usr/bin/yum-deprecated'
```
Conan-Kudo: It seems like a very bad idea to remove this...

Conan-Kudo:
@mich181189 It's actually the opposite. It's more necessary the older the system is. :)

py/mockbuild/util.py
mich181189: It isn't removed, but moved to the package_manager module - that makes it easier to deal with the fact the outer chroot might have a different view of this

mich181189:
I was thinking more in terms of target rather than host but... yeah I see your point

mich181189:
added command line options to override config

ausil:
koji buildroots have no network support in them so this will not work. It would likely need changes in koji to be usable, and in order to deliver updates with new rpm features the entire toolchain needs to updated to support the rpms features. or to at the lease ignore and not fall over on the existence of them.  I would like to see much more clearly defined the end goal and the benefits of supporting them.

ausil:
I also do not see any handling of the cases where the rpmdb between the hosts rpm and the chroots rpm are incompatible, or any handling or what would happen in cases where the rpms going into the minimal environment are incompatible with the hosts rpm, or when the hosts package manager can not deal with the repodata for the chroot. I guess I really want to know what the problem you are trying to solve is. I am just trying to point out some of the issues that I see.

xsuchy:
@ausil This PR is trying to solve https://bugzilla.redhat.com/show_bug.cgi?id=1118719 See the duplicate BZs and GH issue for more use cases with the same root of cause.

Conan-Kudo:
@ausil Koji already does a flavor of this for doing things like creating live media, so doing this for everything would (ironically) make things more consistent.

py/mockbuild/plugin.py
```diff
@@ -15,13 +15,18 @@ def __init__(self, config, state):
...
 
         self.plugins = config['plugins']
         self.plugin_conf = config['plugin_conf']
         self.plugin_dir = config['plugin_dir']
 
     @traceLog()
     def init_plugins(self, buildroot):
+        if self.initted:
+            return
+        self.initted = True
```
xsuchy: initted is not a word. Initialized is the correct word. And I would rather prefer something more descriptive. E.g. self.already_initialized

xsuchy:
Whoa, we just yesterday discussed with @ignatenkobrain that he will start working on this and ... you just did it. Thank you. But this is big change, so it will take me some time to review this.
Can you add entries to man page for that command line option?

ignatenkobrain:
@xsuchy I've kindly asked @mich181189 to do this on `#fedora-rust`, so it was coordinated work :P

xsuchy:
Please run pylint3-pylint after your changes:
```
+************* Module mockbuild.plugin
+C: 18, 0: Trailing whitespace (trailing-whitespace)
+************* Module mockbuild.package_manager
+E: 39,23: No value for argument 'outer_buildroot' in constructor call (no-value-for-parameter)
+E: 43,15: No value for argument 'outer_buildroot' in constructor call (no-value-for-parameter)
+************* Module mockbuild.backend
+C:114, 0: Line too long (132/120) (line-too-long)
+C:119, 0: Line too long (136/120) (line-too-long)
+C:125, 0: Line too long (135/120) (line-too-long)
+C:126, 0: Line too long (135/120) (line-too-long)
+C:148, 0: Line too long (129/120) (line-too-long)
+C: 16, 0: external import "from mockbuild.mounts import BindMountPoint" comes before "from . import util" (wrong-import-order)
 ************* Module mock
+C:336, 0: Trailing whitespace (trailing-whitespace)
+C:338, 0: Wrong continued indentation (remove 2 spaces).
+                        help="build in two stages, using chroot rpm for creating the build chroot")
+                      | ^ (bad-continuation)
+C:340, 0: Wrong continued indentation (remove 2 spaces).
+                        help="build in a single stage, using system rpm for creating the build chroot")
+                      | ^ (bad-continuation)
+C:686, 0: Line too long (142/120) (line-too-long)
```

When I run `mock -r epel-7-x86_64 /tmp/tito/mock-1.4.0-0.git.23.6e32374.fc25.src.rpm` it will crash because there is no /usr/bin/yum-deprecated in that minimal root.

xsuchy:
As @ausil said Koji and Copr have no network on builders. But it can be easily enabled in outer chroot.
You should set:
```
config_opts['rpmbuild_networking'] = True
config_opts['use_host_resolv'] = True
```
for outter chroot.

xsuchy:
> I also do not see any handling of the cases where the rpmdb between the hosts rpm and the chroots rpm are incompatible

This is already happening. And right now it is not handled. Sometimes mock crash, sometimes somehow proceed. In outer chroot we does not care if rpmdb is broken there, because we just install there rpm/dnf/yum. And rpm from outer chroot is used to install rpm and rpmdb in inner chroot. And since outer and inner chroot have same rpm binary, it will be compatible by definition.

xsuchy:
@mich181189 I have no more comment right now, please update PR according the comments above.

py/mockbuild/plugin.py
mich181189: Hehe this was left over from an early and rather dirty attempt! Will fix.

mich181189:
I think that covers everything. 

`mock -r epel-7-x86_64 --rebuild /tmp/tito/mock-1.4.0-0.git.24.3d3d80a.fc25.src.rpm`
works for me. The yum-deprecated problem was something I hit when I first tried epel-7 and fixed before submitting this PR.

mich181189:
Comment from IRC:
> <ignatenkobrain> mich181189: one stylish nitpick
> <ignatenkobrain> def foo(bar = None)
> <ignatenkobrain> don't put spaces inside
> <ignatenkobrain> def foo(bar=None)

xsuchy:
epel7 still does not work for me. I will try to investigate.

One more thing. Language. We have 'bootstrap' in man page/options. We have 'minimal' in name of directory name. We have inner/outer in code. It would be nice to unify this before merging. I like the 'bootstrap' string. So I would propose to s/minimal/bootstrap/ and s/outer/bootstrap/. We can leave 'inner' as is as it is used on only one place in the code.

xsuchy:
> epel7 still does not work for me. I will try to investigate.

Ah that was because I have in config:
```
config_opts['yum_command'] = '/usr/bin/yum-deprecated'
```
which was worked so far and I was originally recommended when you wanted to build EL from latest Fedora. But it will not work now. Waiving out. This can be documented in release notes.
When I removed this line from my config then it is working without problem.

mich181189:
Another comment from IRC (Son_Goku):

> the current option should be used for the inner chroot
> system_{yum,dnf}_binary config option should be used for outer chroot
> because, for example, if someone wants to build the chroot with tdnf, microdnf, zif, or something else, it'd make it much easier

xsuchy:
Merged. Thank you.

mattdm:
Drive-by comment — could this be sped up by having the inner buildroot initially created by hardlinking all of the files from the outer one? (`shutil.copytree` with `copy_function=os.link`)?

mattdm:
And another drive-by comment (sorry)... in Commands class, rather than a bunch of `if self.bootstrap_buildroot is not None:` lines, how about just one statement like that, calling recursively with `buildroot` set to `bootstrap_buildroot` and `bootstrap_buildroot` set to `None`? That seems like it'd be easier to maintain if the code changes

voxik:
And how is this compatible with LVM backend? I can imagine there should be snapshot with the outer buildroot preserved most of the time, since RPM/DNF is not changing that rapidly ...

xsuchy:
@voxik The caching is completely transparent and caching mechanism does not know or need to know about all this. We just create chroot with suffix -bootstrap. It is created the same way as any other chroot (e.g. with --uniqueext ) and whatever caching mechanism is enable it is then used. So root-cache will create the tar file with this name. lvm-plugin will create snapshot with this name.

This also answer the @mattdm question. Only first run is slow, as in second run is used the cache. While the hardlinking is possible, it would only slow down the things. And we will have to explicitely skip some files (eg. that rpmdb file).
The second comment make sense and I welcome any patches which will improve this.

PR #58, opened 2017-03-10
Fix "init_install_output" error

mdionne:
This error:

  AttributeError: 'Buildroot' object has no attribute 'init_install_output'

can occur because init_install_output is initialized in the pkg_manager object but not in the buildroot object.


mdionne:
This was introduced by commit e58f0f5c62fe35740cc3fc5fa8e7 which added an initialization in PackageManager but used it from Buildroot.

The other possible fix is to initialize it in __init__ for Buildroot.

xsuchy:
Rebased and merged. Thank you.

xsuchy:
Commit 85ab3ce

PR #69, opened 2017-05-06
Fix calls of yum-builddep and repoquery, and use 'dnf repoquery' for dnf

Conan-Kudo:
When yum-utils commands are suffixed with `-deprecated`, they should be used instead. In addition, `dnf repoquery` should be used with dnf.


xsuchy:
Nice. Thank you.
I removed the requires of yum-utils in spec in commit 51adc1c

Conan-Kudo:
@xsuchy You'll probably want to make `dnf-plugins-core` a Requires rather than a Recommends now, since it's impossible to do a package build without it for Fedora.

xsuchy:
True. Done in commit 4868afd

PR #75, opened 2017-05-12
Use bash --login with systemd-nspawn as well (rhbz #1450516)

opoplawski:
This uses bash --login -c with nspawn as well to keep the same behavior as before and to allow manipulating the environment of the build with /etc/profile.d/ scripts.


xsuchy:
Looks good to me. Merged as 826c9d6. Thank you.

PR #91, opened 2017-07-10
backend.py: quote check_opt

jlebon:
Now that we always use `bash --login` (after 826c9d6), we have to always
quote the macro. Fixes the regression noted in
https://bugzilla.redhat.com/show_bug.cgi?id=1450516#c12.

---

Untested; though seems right by inspection.


xsuchy:
Thank you.

PR #92, opened 2017-07-11
Make LVM sleep time configurable

mizdebsk:
Let user configure mock to wait more patiently for LVM initialization.
`lvs' call can be quite expensive, especially when there are hundreds
of volumes and multiple parallel mocks waiting for common LVM
initialization to complete.

CC @msimacek


xsuchy:
Merged. Thank you.

PR #93, opened 2017-07-12
Add detection of OL (Oracle Linux) distribution

pixdrift:
On Oracle Linux EL7 platform, systemd-nspawn execution was was failing due to unsupported '-a' parameter being passed. This change adds detection for the 'ol' platform to implement the '-a' workaround for systemd-nspawn on Oracle Linux systems.


xsuchy:
Thank you.

PR #107, opened 2017-08-14
Always create /dev/loop nodes

bcl:
/dev/loop* nodes are only created when they are first used so it should
create them in the chroot unconditionally.

### DIFF
--- a/py/mockbuild/buildroot.py
+++ b/py/mockbuild/buildroot.py
@@ -478,7 +478,8 @@ class Buildroot(object):
             self.root_log.debug("kernel version == %s", kver)
             for i in devFiles:
                 # create node, but only if it exist on host too
-                if os.path.exists("/" + i[2]):
+                # except for loop devices, which only show up on the host after they are first used
+                if os.path.exists("/" + i[2]) or "loop" in i[2]:
                     os.mknod(self.make_chroot_path(i[2]), i[0], i[1])
                     # set context. (only necessary if host running selinux enabled.)
                     # fails gracefully if chcon not installed.


xsuchy:
Hmm I am not sure if solving this in mock is the right solution. I will try to ask systemd guys about this behaviour.

xsuchy:
Hmm this is problem. I talked to one systemd devel and one storage devel. One problem is that devices are not in CGROUPS so if you open device in container you are opening the host device. And even if you create only /dev/loop-control you  can retrieve via this device a lot of information about host. This is probably not issue for local builds, but it is problem for build systems like Koji.
I am willing to merge this, but only if this configuration driven and by default disabled.

Something like:
```
## make devices /dev/loop* available inside of container
## beware that this is security risk for shared systems
# config_opts['container_dev_loop'] = False
```

bcl:
A few things to be aware of:
* This only works with --old-chroot, I haven't figured out how to make it work with the systemd-nspawn /dev/ which seems to overwrite these nodes.
* This is simply restoring functionality of the old mock that we depend on for lorax and livemedia-creator
* Nothing prevents user code running as root inside the mock from making their own nodes with mknod, so this is no less secure than not creating them, or what we've had in the past.

I really don't want to see it hidden behind an option, the whole point in adding them was so that things would work without needed a bunch of other tweaking. That also makes it different from previous mock versions, which were working.

xsuchy:
True. We can create it always. Because nspawn will mount its /dev over the /dev we prepared. I am going to merge this PR. But we need to solve it somehow for --new-chroot. And I am afraid that the solution will be 'do not expect loop device to be available' :(
Can you explain me please for what you are using the loop device? So I can better understand your use case.

xsuchy:
Merged as b6cbd99

ausil:
Running livemedia-creator or other tool that make os disk images or livecds need to use loop devices to use a file as the disk to back the image

xsuchy:
Notes from Michal Sekletar:
you can have loop device inside of container if you
`systemd-nspawn --bind /dev/loop-control ...` or ` systemd-nspawn --register no --capability CAP_MKNOD -D CHROOTDIR`

Using `--register no` means that we will not create new CGROUP and then it does not make sense to use the container. IMHO.

We tried with Michal creating loop devices inside of container and that simply does not work because the loop driver in kernel is not prepared for that. Se we are doomed to old chroot().

> Running livemedia-creator or other tool that make os disk images or livecds need to use loop devices to use a file as the disk to back the image

But why lorax mount it during **build** time?

bcl:
https://rhinstaller.github.io/lorax/lorax.html#how-it-works describes how lorax works. livemedia-creator is similar, but it can create more than just a boot.iso

Basically loop devices are a key requirement for creating any kind of disk image.

I *think* I've tried 'systemd-nspawn --bind /dev/loop-control' and that didn't work, the loop devices didn't show up inside the container. But I can take another look.

Either way, I think this is a feature that mock needs to provide. lorax/livemedia-creator shouldn't need to be setting up /dev/ nodes.

benzea:
Oh, should mock become smarter about the loop device handling, then it might be nice to also be able to clean them up again. I Ctrl+C'ed a build at some point and IIRC I still had a mounted loop device lying around and mock stumbled over that later when trying to clean the tree. Not really important, but would be neat.

xsuchy:
> I think I've tried 'systemd-nspawn --bind /dev/loop-control' and that didn't work, the loop devices didn't show up inside the container

Yes. With loop-control binded inside container you are able to create loop device, but it will appear in the host and is not visible in container.

sergiomb2:
btw also when systemd-tmpfiles-setup.service doesn't start for some reason [1], we need run `modprobe loop`
to create /dev/loop-control ... 

[1]
Detected unsafe path transition / (owned by sergio) → /var (owned by root) 200 times

PR #114, opened 2017-08-28
Fix keeping the LVM root mounted

msimacek:


py/mockbuild/backend.py
```diff
@@ -152,9 +152,9 @@ def init(self, **kwargs):
                 # add the extra bind mount to the outer chroot
                 inner_mount = self.bootstrap_buildroot.make_chroot_path(self.buildroot.make_chroot_path())
                 util.mkdirIfAbsent(self.buildroot.make_chroot_path())
-                self.bootstrap_buildroot.mounts.managed_mounts.append(
-                    BindMountPoint(self.buildroot.make_chroot_path(), inner_mount))
                 self.bootstrap_buildroot.initialize(**kwargs)
+                self.buildroot.mounts.managed_mounts.append(
+                    BindMountPoint(self.buildroot.make_chroot_path(), inner_mount))
```
xsuchy: Can you please elaborate why you moved this down?

msimacek: I didn't only move this this line. I changed `self.bootstrap_buildroot` to `self.buildroot` - I moved the mountpoint to be managed by the non-bootstrap buildroot, which should effectively be the same, but  it makes it possible for LVM plugin to change the mount more easily. Then I moved the line only so that operations on the same object are grouped together.

xsuchy:
Merged. Thank you.

PR #117, opened 2017-09-12
introduce -N for --no-cleanup-after

jsynacek:


praiskup:
FWIW, I don't think it is worth wasting single-literal switch for this.  There's also --no-clean option, which is imo used more often (but still, I think long options are a better).

xsuchy:
@praiskup `--no-clean` already has -n

xsuchy:
Can you alter etc/bash_completion.d/mock as well?

jsynacek:
@praiskup --no-clean alread has -n and, in my opinion, both --no-clean and --no-cleanup-after are used equally, at least in my case. Why should one have a single string option and the other shouldn't? Also, mock uses hardly any single string options, so adding one that makes sense is not wasting. A the maintenance cost is close to nothing, really...

@xsuchy Bash completion only completes long options, doesn't it? I mean I couldn't find any way how to complete -n, for example. And the source doesn't mention it as well.

xsuchy:
@jsynacek right. Completation of one character does not make sense :)

praiskup:
>  Why should one have a single string option and the other shouldn't?

Good point, I didn't know that.  I dislike short options, that's it.

yarda:
Writing just -N is just shorter than writing --no-c Tab -uTab and there is currently inconsistency that --no-clean has short option, but --no-cleanup-after doesn't have one. I am using both options nearly equally. Just stating that "I dislike short options" isn't technical argument.

xsuchy:
merged as 616f00d
Thank you.

PR #118, opened 2017-09-12
new-chroot: set up new network namespace and add default route in it

### LINKED ISSUE #113: Add default route to `lo` in container when network is disabled
`ruby.spec` has a test for ipv4 multicast on mock environment.
I want to enable the multicast for `enp0s25` (ipv4) network interface in mock.

But when I tested with `ip` command to check it on mock, the process during check section, `enp0s25` multicast was still disabled.
Then when I did `mock --shell`, the multicast was enabled.
Do you have any idea to enable the multicast for the process of check section?

```
$ git diff ruby.spec

+BuildRequires: /usr/sbin/ip

-make check TESTS="-v $DISABLE_TESTS"
+sleep 30
+ip maddress
+make test-all TESTS="rinda/test_rinda.rb"
```

```
$ fedpkg srpm
$ mock -r fedora-rawhide-x86_64 *.rpm
```

build.log (https://gist.github.com/junaruga/c6e25d26b09ab89115ad64fa44274763)
```
+ sleep 30
+ ip maddress <= only lo's multicast is enabled.
1:  lo 
    inet  224.0.0.1
    inet6 ff02::1
    inet6 ff01::1
+ make test-all TESTS=rinda/test_rinda.rb
=> Errror
```

```
$ mock -r fedora-rawhide-x86_64 --shell

<mock-chroot> sh-4.4# ip maddress
1:	lo
	inet  224.0.0.1
	inet6 ff02::1
	inet6 ff01::1
2:	enp0s25
	link  01:00:5e:00:00:01
	inet  224.0.0.1
	inet6 ff02::1
	inet6 ff01::1
...
```

msekletar:
Some software excepts (implicitly or otherwise) that system always has
default route properly configured. For example, you can't bind() UDP
socket to all IP addresses and then join multicast group, w/o having
default route. This is what ruby test-suite does and fails due to
missing default route when building ruby with --new-chroot while network
access is disallowed.

rpmbuild_networking option allows you to configure whether mock should
set up build environment in way that it is possible to access
network. This option affects only mock commands spawned with
--new-chroot. Previously we added --private-network to nspawn command
when rpmbuild_networking was set to False. This commit introduces a
change in this regard. We never add --private-network to nspawn
arguments, instead we setup network namespace ourselves and we also add
default route pointing to loopback interface (only interface in the new
namespace). This should fix build ruby's build failure and provide more
"pleasant" build environment even when network access is disallowed.

Note that this commit introduces new dependency which is pyroute2. We
need pyroute2 in order to setup environment in new network namespace.

Fixes #113

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -56,11 +56,11 @@ class Commands(object):
         # do we allow interactive root shells?
         self.no_root_shells = config['no_root_shells']
 
+        self.private_network= not config['rpmbuild_networking']
+
     def _get_nspawn_args(self):
         nspawn_args = []
         if util.USE_NSPAWN:
-            if not self.config['rpmbuild_networking']:
-                nspawn_args.append('--private-network')
             nspawn_args.extend(self.config['nspawn_args'])
         return nspawn_args
 
@@ -313,6 +313,7 @@ class Commands(object):
             ret = util.doshell(chrootPath=self.buildroot.make_chroot_path(),
                                environ=self.buildroot.env, uid=uid, gid=gid,
                                nspawn_args=self._get_nspawn_args(),
+                               unshare_net=self.private_network,
                                cmd=cmd)
         finally:
             log.debug("shell: unmounting all filesystems")
@@ -340,10 +341,12 @@ class Commands(object):
                 self.buildroot.doChroot(args, shell=shell, printOutput=True,
                                         uid=self.buildroot.chrootuid, gid=self.buildroot.chrootgid,
                                         user=self.buildroot.chrootuser, cwd=options.cwd,
-                                        nspawn_args=self._get_nspawn_args())
+                                        nspawn_args=self._get_nspawn_args(),
+                                        unshare_net=self.private_network)
             else:
                 self.buildroot.doChroot(args, shell=shell, cwd=options.cwd,
                                         nspawn_args=self._get_nspawn_args(),
+                                        unshare_net=self.private_network,
                                         printOutput=True)
         finally:
             self.state.finish(chrootstate)
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -38,6 +38,7 @@ import distro
 from . import exception
 from .trace_decorator import getLog, traceLog
 from .uid import getresuid, setresuid
+from pyroute2 import IPRoute
 
 encoding = locale.getpreferredencoding()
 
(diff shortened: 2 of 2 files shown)


msekletar:
Few caveats,
 * I am not a python programmer, so please check this PR carefully
 * I've tested this only manually using ```--chroot``` and ```--shell``` and verifying that I have only loopback interface and default route pointing to it (when ```rpmbuild_networking=False```)
 * If setup of network namespace fails due to an exception in ```condUnshareNet``` it looks like the exception handler catches the exception, logs about it and continues. Well, this is actually not the case. If exception is raised in child it is also raised in parent process (see https://docs.python.org/3.6/library/subprocess.html#exceptions). Hence we should also catch exception in parent, but there are two problems,
    * Type information is lost, so I don't know what exception I am dealing with, because in parent we get generic SubprocesException
    * I need to know that exception is from child process. According to docs this should be possible to figure out because child process exceptions should have ```child_traceback``` attribute set. However, at least with ```python-3.6.2``` subprocess exception never has this attribute, even when it was really raised by child (tested by introducing artificial ```raise``` statement to ```condUnshareNet```)

msekletar:
> If the setup of network namespace fails due to an exception in condUnshareNet it looks like the exception handler catches the exception, logs about it and continues. Well, this is actually not the case. If an exception is raised in the child it is also raised in parent process (see https://docs.python.org/3.6/library/subprocess.html#exceptions). Hence we should also catch exception in parent, but there are two problems,

Above is entirely true because only unhandled exceptions are re-raised in a parent. I figured out that exception was raised in parent because actually exception handler from the child raised a new exception. I force pushed my branch where this is now fixed.

> Type information is lost, so I don't know what exception I am dealing with because in parent we get generic SubprocesException

This is still true, but we are no longer getting the exception in the parent as it is handled by child's exception handler.

> I need to know that exception is from the child process. According to docs, this should be possible to figure out because child process exceptions should have child_traceback attribute set. However, at least with python-3.6.2 subprocess exception never has this attribute, even when it was really raised by child (tested by introducing artificial raise statement to condUnshareNet)

This is most likely a bug that I will report to upstream Python bug tracker.

xsuchy:
I have done few minor changes.
Merged as a1a3dda

voxik:
I can't see any documentation of "rpmbuild_networking" config option in this PR :/ What is the default behavior actually?

xsuchy:
@voxik The behavior has been present there for some time. And it does not change. Previously we just passed option to systemd-nspawn and left nspawn to handle it. Now we do not pass the option to nspawn anymore and we handle the network namespaces ourself.

PR #119, opened 2017-09-15
Added option --debug-config

matejkudera:


xsuchy:
Rebased and merged as 7454609. I added man page entry in commit 792d1c5

PR #123, opened 2017-09-27
use primary key for F-27+ on s390x

sharkcz:
Starting with F-27 the s390x architecture is built in the primary koji instance
and thus using the primary signing key.


xsuchy:
Rebased and merged. Thank you. I will push it to Fedora today.

PR #125, opened 2017-10-05
ccache: use different bind mount directory

msekletar:
When using nspawn /tmp gets over-mounted with tmpfs. Let's use different
directory for our ccache bind mount.


xsuchy:
Merged. Thank you.

PR #126, opened 2017-10-07
Mageia support in the mock/mock-core-configs specs

Conan-Kudo:
This PR includes fixes to the mock and mock-core-configs spec files to support targeting Mageia.


xsuchy:
Merged. Thank you.

PR #127, opened 2017-10-13
Delete rootdir as well when calling clean

### LINKED ISSUE #90: Mock directory clean after
mock   --rootdir=$MOCK_LIBEN    --resultdir  $PACKGS --cleanup-after
cleanup-after option not cleaning the directory : --cleanup-after
I defined different path in my local $MOCK_LIBEN 
i can be able to delete with script "results" directory not able to delete "root" directory 
so i'm looking for solution cleanup --rootdir=$MOCK_LIBEN

athienem:
In case one overrides the rootdir option and the rootdir is located
outside of basedir, it was not cleaned up when calling --clean.

Fix this case by checking if the rootdir is outside basedir. If that
is the case, run an extra rmtree() on it.

Fixes #90


xsuchy:
Merged. Thank you.

Issue #130, opened 2017-10-26
Enable chrootgroup as a config file option


mdhorn:
Recent change (2fc870258473ab36ea65e8996c9b87d466b523e1) to fix Issue #110 changed the hardcoded value of `chrootgroup` from `mockbuild` to `mock`.

While this fixed the Issue seen in #110, it causes other OS RPM builds under Fedora to fail when the OS uses a different GID for the group `mock`.

Suggest changing the hardcoded chrootgroup value to configuration file option so it can be overridden for OS which use a different GID for the mock.

xsuchy:
??? It just changed the name *inside* of chroot. It still uses the same GID, which still maps to the same GID on the host.
Can you give me use case where this caused the problem?

mdhorn:
@xsuchy The PR I submitted is really just a workaround for my issue. But it is also a good idea in general as it makes the value configurable; which is the main reason it was justifiable to me as a workaround.

The problem I encountered was due to the Host OS and the Target chroot OS where the mock build is occurring have different GIDs. I don't have data on why the OS I'm building RPMs for didn't use 135, but that is the case.

The workaround, which is what was previously being done until the recent change in the hardcoded value, is to use a non-conflicting group name -- i.e. not `mock`. When the `groupadd` is done in the chrooted environment a new group `mockbuild` (or whatever the chrootgroup is set to) is created with GID 135,  and then the build is executed with that group. The existing `mock` group with it's different GID also exist, but isn't used for the build.

The first fix attempted was to set the `chrootgid` to the same value as the target OS, but unfortunately that didn't work because various steps in the mock build scripts expect that the GID exists (and is the same) within the Host OS and the Target, chroot OS. The real or better fix might be to rework the mock code to not use the value of chrootgid under the Host OS and introduce a new variable like mockgid or hostosgid for the Host OS.

Let me know if that clarifies the issue or raises more questions.

xsuchy:
Using different GID than 135 can be a problem, but that can be resolved using `chrootgid` option. I still do not see how even different group name can cause any problem inside of chroot.

But if the patch above fixes your issue then fine. I am closing this one as the patch have been merged. Thank you.

PR #131, opened 2017-10-26
Enable chrootgroup as a config file option

mdhorn:
Fix for #130

Change the previously hardcoded value for chrootgroup to be
a configuration file option which defaults to the previous
hardcoded value.

Signed-off-by: Mark D Horn <mark.d.horn@intel.com>


dorileo:
+1

xsuchy:
Merged.
I added two additional commits:
* 5fe7961 user and group is actually not used here since some logic moved to buildroot.py
* 9fb3ca7 add config_opts['chrootgroup'] to site-defaults.cfg

xsuchy:
This is good to have and because it is super easy I merged it. But my question in the original Issue still stands.

PR #139, opened 2017-12-02
Strip trailing / from mountpath in ismounted()

funkyHat:
Fixes an issue where a failed build can result in mountpoints left around causing subsequent builds to fail.

e.g. `/var/lib/mock/epel-7-x86_64/root/var/cache/yum`(`/`).
`ismounted()` currently returns `False` because the trailing slash does not appear in `/proc/mounts`

Removing the trailing slash from `config_opts['plugin_conf']['yum_cache_opts']['target_dir'] = "/var/cache/%(package_manager)s/"` also works around the issue.


xsuchy:
Thank you.

PR #141, opened 2017-12-07
Fix for RHBZ 1467314

ellert:
This is a fix for the issue in https://bugzilla.redhat.com/show_bug.cgi?id=1467314, the changes in this PR are the same as those proposed in comment 4 of that bugzilla report. See the bugzilla report for details.

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -78,6 +78,7 @@ Suggests: yum
 Requires: dnf-plugins-core
 Recommends: btrfs-progs
 Recommends: dnf-utils
+Recommends: dnf-yum
 %endif
 %if 0%{?rhel} == 7
 Requires: btrfs-progs
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -77,7 +77,7 @@ Requires: dnf
 Suggests: yum
 Requires: dnf-plugins-core
 Recommends: btrfs-progs
-Suggests: dnf-utils
+Recommends: dnf-utils
 %endif
 %if 0%{?rhel} == 7
 Requires: btrfs-progs
--- a/mock/py/mockbuild/plugins/selinux.py
+++ b/mock/py/mockbuild/plugins/selinux.py
@@ -114,6 +114,6 @@ class SELinux(object):
             supported = hasattr(cli.YumBaseCli, "_parseSetOpts")
             sys.path.pop(0)
             return supported
-        except SyntaxError:
+        except (SyntaxError, ModuleNotFoundError):
             # We're on python 3, assuming yum is new enough
             return True


mock/py/mockbuild/plugins/selinux.py
```diff
@@ -114,6 +114,6 @@ def _selinuxYumIsSetoptSupported(self):
             supported = hasattr(cli.YumBaseCli, "_parseSetOpts")
             sys.path.pop(0)
             return supported
-        except SyntaxError:
+        except (SyntaxError, ModuleNotFoundError):
```
xsuchy: Why this exception?... Hm looking at this function... This is 7 years old code. And it should be true on all supported platforms (RHEL6). So we do not need to discuss this and you can as well remove whole _selinuxYumIsSetoptSupported() and it will be ok.

mock/mock.spec
```diff
@@ -78,6 +78,7 @@ Suggests: yum
 Requires: dnf-plugins-core
 Recommends: btrfs-progs
 Recommends: dnf-utils
+Recommends: dnf-yum
```
xsuchy: Why?

ellert:
In order to use mock to build for EPEL you need /usr/bin/yum and /usr/bin/yum-builddep.

$ grep /usr/bin/yum /usr/lib/python3.6/site-packages/mockbuild/util.py
    config_opts['system_yum_command'] = '/usr/bin/yum'
    config_opts['yum_builddep_command'] = '/usr/bin/yum-builddep'

These programs are provided by dnf-yum and dnf-utils:

$ rpm -q --whatprovides /usr/bin/yum
dnf-yum-2.7.5-2.fc27.noarch
$ rpm -q --whatprovides /usr/bin/yum-builddep
dnf-utils-2.1.5-1.fc27.noarch

So in order for mock to work out of the box it needs to

Recommends: dnf-yum
Recommends: dnf-utils

It currently does:

Suggests: yum
Suggests: dnf-utils

The "Suggests: yum" is not helpful at all since it provides /usr/bin/yum-deprecated which is not used by mock. This suggests was probably meant to suggest a package that provides /usr/bin/yum, but to achieve this it suggests the wrong package.

So at least the "Suggests: yum" should be changed to "Suggests: dnf-yum".

Since suggestions are not installed by default, the functionality is missing after a default installation. Building packages for EPEL is such an important feature of mock that I think it deserves to work out of the box, so I think the two suggests should be upgraded to recommends.

xsuchy:
> The "Suggests: yum" is not helpful at all since it provides /usr/bin/yum-deprecated which is not used by mock. This suggests was probably meant to suggest a package that provides /usr/bin/yum, but to achieve this it suggests the wrong package.

No. Package dnf-yum provides /usr/bin/yum, but that is not that YUM. It is just wrapper around dnf, which prints warning about YUM being deprecated and calls DNF.
So real YUM is in package yum. And mock in fact calls yum-deprecated. See
https://github.com/rpm-software-management/mock/blob/devel/mock/py/mockbuild/package_manager.py#L210

ellert:
What you write used to be true some releases ago, but is no longer true. I don't have the package "yum" installed and have no issues using mock to build packages for EPEL. I do however have the "dnf-yum" package installed. If I uninstall dnf-yum mock breaks and EPEL builds are no longer possible.

As can be seen in the RHBZ issue referenced earlier in this ticket the error is as follows:

$ mock -r rhel-7-x86_64 --init 
INFO: mock.py version 1.4.2 starting (python version = 3.6.1)...
ERROR: Command /usr/bin/yum is not available. Either install package containing this command
or run mock with --yum or --dnf to overwrite config value. However this may
lead to different dependency solving!

xsuchy:
You do not have an issue. But you are in fact using DNF for installing EPEL chroot. Which can give you slightly different results than installing the same chroot using YUM.  Check your /usr/bin/yum what it actually runs (last two lines). That is reason why mock rather use /usr/bin/yum-deprecated if it exists.

Mock should warn user that he is using DNF instead of YUM. Similary like we do that in opposite direction (we have config_opts['dnf_warning'] = True for that), but the detection would be tricky.

PR #142, opened 2017-12-14
use recursive unmount for tmpfs

d3matt:
In some instances, proc/sys/fs/binfmt_misc would end up not getting
cleaned up correctly. Adding -R to the unmount lines fix it for me.


xsuchy:
Hmm, it is probably a good thing. But I wonder about the first -R ... We should probably not pass it for first call of umount(). If the umount will fail, we will print warning (non fatal) and set force=True and only for second call pass -R.
Sounds right?

d3matt:
That sounds reasonable to me... that way someone can eventually clean up the extra mount points more correctly...

xsuchy:
Thank you.

Issue #14, opened 2016-11-10
The do function should always check timeout even when running yum


xning:
Now when running _yum_ mock does not check timeout because the timeout parameter's value is zero. In this case if _yum_ encounters a dead lock, and this situation will hit  _mock_  In a infinite loop.

    do('/usr/bin/yum --installroot /var/lib/mock/RHEL-6-build-repo_latest/root/  groupinstall build', True, None, None, 0, True, 1, None, None, None)

What if we have the _timeout_ parameter has a default non-zero value?

xning:
To make sure that the _timeout_ logic always could be available and will be run, I think we also need a _watchdog_ which should be a standalone process or a handler of the SIGALARM signal, hence even mock process is hit in endless loop or encounters a dead lock, the _timeout_ logic still could work.

xsuchy:
I think it is hard to set some timeout value as some lines can be slow and this command can last quite long.
I will be happy to review patch which adds this option to site-defaults.cfg and which will be used in code and propagated to do().

xsuchy:
I beleive this is resolved now with `config_opts['opstimeout'] = 0` See https://github.com/rpm-software-management/mock/wiki/Release-Notes-1.4.9

PR #146, opened 2017-12-28
New config option 'opstimeout' has been added.

plageat:
This is regarding issue #14

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -536,12 +536,14 @@ def resize_pty(pty):
 #
 @traceLog()
 # pylint: disable=unused-argument
-def _do(command, shell=False, chrootPath=None, cwd=None, timeout=0, raiseExc=True,
+def do(command, shell=False, chrootPath=None, cwd=None, timeout=0, raiseExc=True,
        returnOutput=0, uid=None, gid=None, user=None, personality=None,
        printOutput=False, env=None, pty=False, nspawn_args=None, unshare_net=False,
        *_, **kargs):
 
     logger = kargs.get("logger", getLog())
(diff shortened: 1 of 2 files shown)


mock/py/mockbuild/util.py
```diff
@@ -620,6 +622,10 @@ def do(command, shell=False, chrootPath=None, cwd=None, timeout=0, raiseExc=True
 
     return output
 
+def do(command, *args, **kargs):
+    timeout = kargs.get('timeout', _OPS_TIMEOUT)
+    f = functools.partial(_do, timeout=timeout)
+    return f(command, **kargs)
```
xsuchy: Why so complicated. Cannot be just 
```
def do(... timeout=None, ..):
if timeout is None:
   timeout = _OPS_TIMEOUT
```

xsuchy:
Beside those inline comments. I like this patch. Once you address those inline comments I will merge it.

mock/py/mockbuild/util.py
plageat: Agree.I will re-work this

mock/py/mockbuild/util.py
```diff
@@ -620,6 +621,10 @@ def do(command, shell=False, chrootPath=None, cwd=None, timeout=0, raiseExc=True
 
     return output
 
+def do(command, *_, **kargs):
+    if 'timeout' not in kargs:
+        kargs['timeout'] = _OPS_TIMEOUT
```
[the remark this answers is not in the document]
plageat: Agree, there is no reason. Will be re-worked

xsuchy:
Thank you.

PR #152, opened 2018-01-08
Don't setup mounts in the bootstrap buildroot

ifireball:
Its supposed to be a minimal buildroot just for running the package
manager, so don't try to mount things into it.

### DIFF
--- a/mock/py/mockbuild/plugins/mount.py
+++ b/mock/py/mockbuild/plugins/mount.py
@@ -42,6 +42,9 @@ class Mount(object):
         self.config = buildroot.config
         self.state = buildroot.state
         self.opts = conf
+        # Skip mounting user-specified mounts if we're in the boostrap chroot
+        if buildroot.is_bootstrap:
+            return
         plugins.add_hook("postinit", self._mountCreateDirs)
         for device, dest_dir, vfstype, mount_opts in self.opts['dirs']:
             buildroot.mounts.add_user_mount(
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -685,15 +685,15 @@ def main():
     log.info("mock.py version %s starting (python version = %s)...",
              __VERSION__, py_version)
     state = State()
-    bootstrap_buildroot_state = State(bootstrap=True)
     plugins = Plugins(config_opts, state)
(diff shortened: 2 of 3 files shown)


xsuchy:
Not actually true. We want to have Mounts.essential_mounts to have mounted (procfs and sysfs). But I agree that we likely wanto mount *just* esential mounts and not anything else.

xsuchy:
I think that bind_mount can be safely set to False.

xsuchy:
I think the best way to disable use mounts will be to add in mock/py/mockbuild/plugins/mount.py if condition and skip the for-loop in __init__() if buildroot.is_bootstrap.

ifireball:
ping

mock/py/mockbuild/plugins/mount.py
```diff
@@ -42,6 +42,9 @@ def __init__(self, plugins, conf, buildroot):
         self.config = buildroot.config
         self.state = buildroot.state
         self.opts = conf
+        # Skip mounting user-specified mounts if we're in the boostrap chroot
+        if buildroot.is_bootstrap:
+            return
```
xsuchy: This  needs to go one line lower. We need to enable this hook (to mount essential mounts), but we do not want to add custom mounts (in that for loop 2 lines below).

ifireball: Given that self.opts['dirs'] seems to only contain user defined (non-essential) mounts, and that the '_mountCreateDirs' function only deals with mounts listed there, it doesn't seem to have anything to do with essential mounts.

Also IMO its better not to run '_mountCreateDirs' if we're not going to mount so unnecessary mount points will not be created. 

IMO it would be kind of strange if this plugin that seems to only deal with user-defined mounts had something to do with the essential mounts...

Also in my testing the essential mount were being configure in the bootstrap chroot just fine...

xsuchy: Right, I was confused.

xsuchy:
Thank you.

ifireball:
when can we expect an updated mock package to be released?

xsuchy:
I usually try to stick to 2 month cadence. Last release was just before Christmas. So you can expect release late in February.  
In the meantime you can use package from nightly build https://copr.fedorainfracloud.org/coprs/g/mock/mock/

PR #155, opened 2018-01-22
 fix #65 

### LINKED ISSUE #65: mockchain does not understand 'include' in mock configs
Hello,
it appears that mockchain is having issues parsing 'include' statements in mock config files. RPM Fusion mock config contains the following line:
`include('/etc/mock/fedora-rawhide-x86_64.cfg')`
using `-r fedora-rawhide-x86_64-rpmfusion_free` parameter works perfectly fine when using with standalone mock, but with mockchain the following happens:
`Traceback (most recent call last):`
`  File "/usr/bin/mockchain", line 416, in <module>`
`    sys.exit(main(sys.argv))`
`  File "/usr/bin/mockchain", line 298, in main`
`    res, msg = add_local_repo(config_opts['config_file'], my_mock_config, local_baseurl, 'local_build_repo')`
`  File "/usr/bin/mockchain", line 142, in add_local_repo`
`    exec(code)`
`  File "/etc/mock/fedora-rawhide-x86_64-rpmfusion_free.cfg", line 1, in <module>`
`    include('/etc/mock/fedora-rawhide-x86_64.cfg')`
`NameError: name 'include' is not defined`
This is with mock-1.3.4-1.fc25.noarch

Best regards,
Julian

sergiomb2:
and simplify a lot include

### DIFF
--- a/mock/py/mockchain.py
+++ b/mock/py/mockchain.py
@@ -138,7 +138,9 @@ def add_local_repo(infile, destfile, baseurl, repoid=None):
 
     try:
         with open(infile) as f:
-            code = compile(f.read(), infile, 'exec')
+            content = f.read()
+            content = re.sub(r'include\((.*)\)', r'exec(open(\g<1>).read(), {}, locals())', content)
+            code = compile(content, infile, 'exec')
         # pylint: disable=exec-used
         exec(code)
         if not repoid:
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1173,25 +1173,6 @@ def check_config(config_opts):
 
 
 @traceLog()
-def include(config_file, config_opts, is_statement=False):
-    if os.path.exists(config_file):
-        if is_statement and config_file in config_opts['config_paths']:
-            getLog().warning("Multiple inclusion of %s, skipping" % config_file)
-            return
-
-        config_opts['config_paths'].append(config_file)
-
-        with open(config_file) as f:
-            content = f.read()
-            content = re.sub(r'include\((.*)\)', r'include(\g<1>, config_opts, True)', content)
-            code = compile(content, config_file, 'exec')
-        # pylint: disable=exec-used
-        exec(code)
-    else:
-        raise exception.ConfigError("Could not find included config file: %s" % config_file)
-
-
-@traceLog()
 def update_config_from_file(config_opts, config_file, uid_manager):
     config_file = os.path.realpath(config_file)
     r_pipe, w_pipe = os.pipe()
@@ -1200,7 +1181,11 @@ def update_config_from_file(config_opts, config_file, uid_manager):
             os.close(r_pipe)
             if uid_manager and not all(getresuid()):
                 uid_manager.dropPrivsForever()
-            include(config_file, config_opts)
+            with open(config_file) as f:
+                content = f.read()
+                content = re.sub(r'include\((.*)\)', r'exec(open(\g<1>).read(), {}, locals())', content)
+                code = compile(content, config_file, 'exec')
+            exec(code)
             with os.fdopen(w_pipe, 'wb') as writer:
                 pickle.dump(config_opts, writer)
         except:


xsuchy:
What happens when the config file does not exist? Can you raise exception.ConfigError ?

praiskup:
The infinite recursion protection is there intentionally IMO.  I would keep that, too.

mock/py/mockchain.py
```diff
@@ -138,7 +138,9 @@ def add_local_repo(infile, destfile, baseurl, repoid=None):
 
     try:
         with open(infile) as f:
-            code = compile(f.read(), infile, 'exec')
+            content = f.read()
+            content = re.sub(r'include\((.*)\)', r'exec(open(\g<1>).read(), {}, locals())', content)
+            code = compile(content, infile, 'exec')
```
praiskup: Why you remove the function and rather duplicate the code on two places?

sergiomb2:
> What happens when the config file does not exist? Can you raise exception.ConfigError ?

FileNotFoundError: [Errno 2] No such file or directory: '/etc/mock/fedora-rawhide-x86_64sss.cfg'

mock/py/mockchain.py
sergiomb2: the code are in 2 places , just must more simplified

praiskup:
What happens now if file includes itself?

mock/py/mockchain.py
praiskup: I see, looking again -- the duplication comes from different reason (mockchain support) ... so wouldn't be `code = parse_config(filename)` function cleaner?

plageat:
This is an error with no pros:
- ofc, you can't avoid include loops, see RHBZ1272381
- code dups looks really ugly
- I really doubt whether opening a file + exec in a code is more readable then name 'include'

xsuchy:
I just asked one intern to merge mockchain code into mock itself. So I think this PR will not be merged at the end. But I will leave it open for some time till the new work will be done.

sergiomb2:
Hi, thanks, I'm still studying the subject .

> so wouldn't be code = parse_config(filename) function cleaner?

To me make sense have one function that reads and parse configs and be used by other packages , like  fedora-review package , for example. 

In other hand I still thinking in a better way of use or read includes , but I hadn't much time ... 

Best regards,

xsuchy:
> To me make sense have one function that reads and parse configs and be used by other packages , like fedora-review package , for example.

That would be cool.

sergiomb2:
I have one idea , which is not have include in cfg files , and have templates files, these one with include funcitons, these tpl  files generate the cfgs , we just need do a command that regenerate cfg files which we run when we modify tpl . And it solves a few problems

praiskup:
On Thursday, March 15, 2018 5:06:53 AM CET Sérgio Basto wrote:
> I have one idea , which is not have include in cfg files and have
> templates files, these one with include funcitons, these tpl files
> generate the cfgs

That would mean that tools like copr -- now using trivial `copr
mock-config praiskup/ping epel-7-x86_64` configs -- would stop working...
and they would have to always generate fully expanded configuration files.

It is though absolutely convenient to use the "defaults" provided by mock,
and just adjust a few options (minimal buildroot package set, additional
package repositories), without re-generating the config over-again.

> , we just need do a command that regenerate cfg
> files which we run when we modify tpl . And it solves a few problems 

I see _no_ problems in the actual implementation to be solved;  except
that the actual code should be wrapped into function and reused in
mockchain.  But note (a) there's attempt to merge mockchain into mock (PR
#163) and also the macro-through-jinja idea is in PR #161.

sergiomb2:
> That would mean that tools like copr -- now using trivial `copr mock-config praiskup/ping epel-7-x86_64` configs -- would stop working... and they would have to always generate fully expanded configuration files.

Can you please give me more information ? why it would break ? 

About defaults , I also though about that , I want remove config_opts['yum_common_opts'] and config_opts['dnf_common_opts'] , and use one template with defaults ,  the goal is cfg files have the complete and absolute configuration of mock-config generated , so we avoid confusions . 

I have to go , thanks

praiskup:
On Thursday, March 15, 2018 5:44:36 PM CET Sérgio Basto wrote:
> > That would mean that tools like copr -- now using trivial `copr
> > mock-config praiskup/ping epel-7-x86_64` configs -- would stop
> > working... and they would have to always generate fully expanded
> > configuration files.
> 
> Can you please give me more information ? why it would break ? 

We would have to fix copr to instantiate some template, probably, instead
of relying on include().

> About defaults , I also though about that , I want remove
> config_opts['yum_common_opts'] and config_opts['dnf_common_opts'] , and
> use one template with defaults , the goal is cfg files have the complete
> and absolute configuration of mock-config generated , so we avoid
> confusions .

I'm not sure what confusion you mean here.  Can you elaborate?

sergiomb2:
> I'm not sure what confusion you mean here. Can you elaborate? 

OK, maybe call it an ambiguous configuration ,  when I set in /etc/mock/site-defaults.cfg  [1] metadata_expire=28800,  in the end what is the metadata_expired value ? 28800 or 0 from /etc/mock/fedora-26-x86_64.cfg ? 

[1]
config_opts['dnf_common_opts'] = ['--disableplugin=local', '--setopt=deltarpm=false', '--setopt=fastestmirror=1', '--setopt=max_parallel_downloads=10','--setopt=metadata_expire=28800' ]

xsuchy:
@sergiomb2  The order of evaluation is documented here: https://github.com/rpm-software-management/mock/wiki#generate-custom-config-file The former override the first one.

xsuchy:
Closing this in favour of https://github.com/rpm-software-management/mock/pull/163 (which still needs some work)

PR #156, opened 2018-01-23
Detect if essential mounts are already mounted

msimacek:
Previously, mock assumed that essential mounts are never mounted when
mock starts up. That's not true, as multiple non-destructive mock
processes are allowed (`--shell`, `--install`, etc.) to run
concurrently. So when you use `mock --shell` and do a `mock --install`
in parallel, it breaks your shell, because it unmounts its proc.
This improves the situation by first asking whether the mounts aren't
there already.


xsuchy:
Thank you.

PR #159, opened 2018-02-01
mock-core-configs: add CentOS SCL repositories to EPEL 6 & 7 (x86_64)

tmzullinger:
EPEL now allows packages to use 'devtoolset*' packages from the SCLo and
SCLo-rh repositories as BuildRequires.  Restrict the package list from
the SCLo repositories to 'devtoolset*' with the includepkgs option.

These repositories are currently only available for x86_64.  aarch64 and
ppc64le are in testing now in CentOS¹.

¹ https://bugs.centos.org/view.php?id=14078

### DIFF
--- a/mock-core-configs/etc/mock/epel-6-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-6-x86_64.cfg
@@ -49,6 +49,22 @@ gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-6
 gpgcheck=1
 skip_if_unavailable=False
 
+[sclo]
+name=sclo
+baseurl=http://mirror.centos.org/centos/6/sclo/x86_64/sclo/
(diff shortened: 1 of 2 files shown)


tmzullinger:
This is per discussion on epel-devel: https://lists.fedoraproject.org/archives/list/epel-devel@lists.fedoraproject.org/thread/6EJIQ72JXWAAOQPFLRFYE5A7CUKC756D/

I used the repository URL's published in the centos-release-scl and centos-release-scl-rh packages.  The CentOS bug referenced in the commit message mentions that the aarch64 and ppc64le repositories will be moved to [buildlogs](https://buildlogs.centos.org/centos/).  It is unclear to those of us on the epel-devel list which are intended to be the canonical URLs.  Using buildlogs with https is a bit more secure/private for users.  But using what is shipped in the centos-release-scl* packages seems like a better plan.

xsuchy:
Thank you.

Issue #138, opened 2017-11-29
provide --setopt=opt=val option


praiskup:
It would be nice to be able to set any option by, e.g. --setopt=option=value, for example
--setupt=rpmbuild_networking=False, etc.

xsuchy:
I am pretty sure we already have issue or BZ for that. The "problem" is how to differ "False" and False and how to implement lists. This is for a long time on my TODO list. And I am afraid it will stay there few more years. If anyone wants to implement I will happily review it.

xsuchy:
Done in commit:
* 80273e1 (HEAD -> devel) add new option --config-opts [GH#138]

You can:

run:
  mock --config-opts yum_command=/usr/bin/yum-deprecated --enable-network
which will set:
  config_opts['system_yum_command'] = '/usr/bin/yum'

or for list:
  mock --config-opts extra_chroot_dirs=/mnt/b --config-opts extra_chroot_dirs=/mnt/a
which will set
  config_opts['extra_chroot_dirs'] = ['/mnt/b', '/mnt/a']

or list with single item:
  mock --config-opts extra_chroot_dirs=/mnt/b --config-opts extra_chroot_dirs=
which will set
  config_opts['extra_chroot_dirs'] = ['/mnt/b']

It can detect boolean:
  mock --config-opts nosync=False  --debug-config |grep nosync
  config_opts['nosync'] = False

Specialized option has priority. Therefore:
  mock --config-opts rpmbuild_networking=False --enable-network --debug-config |grep rpmbuild_networking
config_opts['rpmbuild_networking'] = True

It is unable to set complicated variables. Like config_opts['plugin_conf']['package_state_opts'] or anything which has dictionary as value

praiskup:
FTR https://bugzilla.redhat.com/show_bug.cgi?id=1544810

PR #161, opened 2018-02-12
Unified configuration system

plageat:
So, this is an attempt to re-design mock's configuration system (BZ#1208344, issue #6) with templates args based on jinja2.
I hope this at least will stimulate some discussion of the topic as well as this patch does not introduce a lot of new bugs :)

Short description of new algorithm:
1. gather all config paths
2. merge all configs in one
3. expand jinja2 templates
4. provide final config or configuration error
5. do config post setup

Notes:
- config_opt definitions and expansions are located in one place (hope so)
- almost all string formatting transformed to jinja2 show template, for example:
    "%(cache_topdir)s/%(root)s/ccache/u%(chrootuid)s/"
         => 
    "{{ config.cache_topdir }}/{{ config.root }}/ccache/u{{ config.chrootuid }}/"
- 'include()' became '{% include '/absolute/path/to/config.cfg' %}'
- multiple inclusion detection is not detected anymore, if there is recursive include, max depth error will be shown
- template string substitution is missing in cmd options at the moment (that one: --resultdir=./my/"%(dist)s"/"%(target_arch)s")
- ofc, not all of jinja2 features can be used (like {% import ... % as ... %}, etc.)


praiskup:
-1. Incompatible change, why you don't create some common "parent" configuration
file which other config files will depend on by `include()`?

Btw. include is now used e.g. in copr, and I don't really see any reason to drop that.

Btw. this is something I use on daily basis.

```python
include('/etc/mock/fedora-rawhide-x86_64.cfg')
some_option=True
```

plageat:
> create some common "parent" configuration file which other config files will depend on by include()
I don't really understand, could you provide an example content of such a parent config and include your include() logic?

 `include('/etc/mock/fedora-rawhide-x86_64.cfg')`
 `some_option=True`

is now:

 `{% include '/etc/mock/fedora-rawhide-x86_64.cfg' %}`
 `some_option=True`

include feature works almost the same as the old one, there are few differences:
- new syntax (to use include from jinja2)
- recursion does not skipped silently, instead recursion error is thrown (may be reworked somehow or left)

praiskup:
You can have one config for mageia.cfg, then mageia-N.cfg can inherit from mageia.cfg,
and architecture specific mageia-N-ARG.cfg can inherit from that. It is python, so you can do things like
`config_opts['yum.conf'] += .. something ..`
That way we could save some kilobytes in RPM installation, and make a lot less cut-and-paste
work.

Btw. I can see you have `config_opts['useradd'] = '/usr/sbin/useradd -m -u %(uid)s -g %(gid)s -d %(home)s -n %(user)s'` into mageia configs.  This line can be added to mageia.cfg and mageia-N and mageia-N-ARCH don't have to care about that.


I wouldn't have a problem with templating (even though it is IMO pretty useless) if that
wasn't incompatible change.

praiskup:
I wouldn't have problem with templating here, even though it adds jinja dependancy, but -1 for incompatible change without clear motivation.

praiskup:
Also, see `copr mock-config ngompa/dnf-el7 epel-7-x86_64`, similar mock profiles
uses copr internally nowadays.

plageat:
> You can have one config for mageia.cfg, then mageia-N.cfg can inherit from mageia.cfg,
and architecture specific mageia-N-ARG.cfg can inherit from that. It is python, so you can do things like
config_opts['yum.conf'] += .. something ..
That way we could save some kilobytes in RPM installation, and make a lot less cut-and-paste
work.

As I said before, include logic worked and still works as it was. You can do it with new include without any problem, nothing is broken.

>Btw. I can see you have config_opts['useradd'] = '/usr/sbin/useradd -m -u %(uid)s -g %(gid)s -d %(home)s -n %(user)s' into mageia configs. 

This pull request is not about config files(mock-core-configs) itself and better config files structure, but rather config files handling.

>This line can be added to mageia.cfg and mageia-N and mageia-N-ARCH don't have to care about that.

Yes, it can be and should be. And again: it still works as it was The only change in such nested configs is 'include()' turned to '{% include %}' if you already use 'include()' statement

The new syntax is used to avoid 'double-include' syntax - one from jinja2 ({% include %}) and second include() as python function (the old include). As far as as 2 of them can do almost the same functions, the old python include is deleted.

praiskup:
> This pull request is not about config files(mock-core-configs) itself and better config files structure, but rather config files handling.

Then please show the concrete case which motivates you to do the rewrite.  So far it looks
like the actual config file handling should work fine for you.

> The new syntax is used to avoid 'double-include' syntax - one from jinja2 ({% include %}) and second include() as python function (the old include). As far as as 2 of them can do almost the same functions, the old python include is deleted.

But there's no reason to drop the old syntax without asking when someone is using that.
At least it should be "obsolted" for some time in my opinion.

plageat:
> Then please show the concrete case which motivates you to do the rewrite. So far it looks
like the actual config file handling should work fine for you.

- https://bugzilla.redhat.com/show_bug.cgi?id=1208344 , see comments 3 and 4
- issue #6 comments 2,5 and other

> But there's no reason to drop the old syntax without asking when someone is using that.
At least it should be "obsolted" for some time in my opinion.

I provided my reason about 'double-include'. It's not a problem to leave the old one if more of people want to have 2 ways of include

praiskup:
> https://bugzilla.redhat.com/show_bug.cgi?id=1208344 , see comments 3 and 4

I don't understand that bug request well.  That request requires to expand rpm
macros in variables without using RPMs (NAME/VERSION/...)?  Shouldn't one
use `subprocess.call(['rpm', 'eval', <something>])` if any?  All in all, looks like
user error.

> issue #6 comments 2,5 and other

Also seems to just say that we need something, but not why. :-(

plageat:
> I don't understand that bug request well
You do not need. Just read the pointed comments

>Also seems to just say that we need something, but not why. :-(
The first few comments explains, that a little usability improvement can be done to allow parametrized config options in cmd

>mock now can set pretty much any config option on commandline (#138)

well, did not know, but anyway - is not complete, for example, because (taken from #138)

> It is unable to set complicated variables. Like config_opts['plugin_conf']['package_state_opts']'

With this patch that is easy to implement in future

praiskup:
> > I don't understand that bug request well
>
> You do not need. Just read the pointed comments

I read them, and these commends don't justify this change.  At least comment 4 seems to be totally off-topic and orthogonal (OP claims that mock should check that some variables (macros in fact) should not be accepted as part of config options).  Original issue is not caused by mock.

This is PR is rework of something you wrote (thanks!) and which works fine, without clear description what real problem the PR solves (except that it adds jinja2, which has pros/cons).  But I really don't need to understand if you don't feel like that, only if you wanted my understanding.

> > It is unable to set complicated variables. Like config_opts['plugin_conf']['package_state_opts']'
>
> With this patch that is easy to implement in future

Same as without it, the mock's config is turing complete now.  So FWIW I'm -1 here at least as long as it:
- brings incompatible change
- there's missing motivation for this change (real use-case which is not now working fine)

plageat:
>there's missing motivation

The changes will allow us:
- a base for implementing parametrized config options in cmd (solve #6) - this is very clear motivation 
- as a result - usability improvement - this is very clear motivation (if you do not agree with 2 above, please write to connected #6 not here)
- gathering all config option handling in one function instead of scattering across all the files (need to be checked by someone experienced) = code refactoring - this is very clear motivation
- jinja2 templating instead of 'random' string formatting - this is very clear motivation
- improve #138 constraints like for nested dict in config_opts - this is very clear motivation

> brings incompatible change

As I said before, if old include() syntax is a problem, then it can be reverted any time if more people vote
with plan how to deal with 'double include()' described above

xsuchy:
* This PR is hard to review. It is huge. It would be nice to have the code change in one commit and then replacing all those macros with jinja variables in another commit.
* All previous macros (including %include) should work. We can mark them as obsolete. But we have to give users some grace period for migrating their configs.
* I would rather see config_opt as an instance of UserDict or custom class inherited from dict class where you modify __getitem__() and expand the jinja temples only when they are being read. This way it will work transparently even if some plugin later modifies config_opts.

plageat:
> All previous macros (including %include) should work

2 vs 1, so will be reverted

> an instance of UserDict or custom class

I thought of it, is it ok if we have:

`config_opts = {}`
`config_opts['plugin_conf'] = { ... `
`    'mount_opts' : { 'dirs' : [ ... ]`
`    }`
`}`

turned to:

`config_opts['plugin_conf'] = UserDict()`
`config_opts['plugin_conf']['mount_opts'] =  UserDict()`
`config_opts['plugin_conf']['mount_opts']['dirs'] = UserDict()`
`# ... initialization`

Also, question for this values which are lists:

`config_opts['plugin_conf']['root_cache_opts']['exclude_dirs'] = [ val1, val2, val3 ]`

Should we use UserList + getitem() as well?

plageat:
Anyway, because mock have nested configs and includes (site-defaults.cfg, .mock/user.cfg etc.) it means, that we need render jinja2 templates for all of them on each config attribute access (maybe thousands of renders). As the result - significant performance decrease. Actually, I do not really thing that  getitem() + UserDict() have future

xsuchy:
> config_opts['plugin_conf']['root_cache_opts']['exclude_dirs'] = [ val1, val2, val3 ]

Hmm, right that would have to be:

```
config_opts['plugin_conf']['root_cache_opts']['exclude_dirs'] = UserList([ val1, val2, val3 ])
```

which would be weird and I guess people will often forget to use the UserList or UserDict. I am pulling back this idea. You are right.

plageat:
Finally, if there is the old include() in config, it will be transformed to jinja2 include statement

mock/py/mockbuild/util.py
```diff
@@ -32,7 +32,7 @@
 from textwrap import dedent
 import time
 import uuid
-
+import jinja2
```
xsuchy: There have to be `Requires: python3-jinja` in spec file. And %if condition for EL7 with python2.

mock/py/mockbuild/util.py
```diff
@@ -1175,35 +1175,71 @@ def check_config(config_opts):
...
-            content = f.read()
-            content = re.sub(r'include\((.*)\)', r'include(\g<1>, config_opts, True)', content)
-            code = compile(content, config_file, 'exec')
-        # pylint: disable=exec-used
+def _check_unrendered_config_opts(config_opts, content):
+    env = jinja2.Environment(undefined=jinja2.StrictUndefined, loader=jinja2.FileSystemLoader('/'))
+    template = env.from_string(content)
+
+    try:
+        template.render(config=config_opts)
```
xsuchy: I am slightly in favour of `config_opts=config_opts`. This way the variable in a template will be the same as in config.

mock/py/mockbuild/util.py
```diff
@@ -1175,35 +1175,71 @@ def check_config(config_opts):
...
+    except jinja2.exceptions.UndefinedError as ex:
+        m_ar = ex.message.split('attribute')
+        if len(m_ar) == 2:
+            new_msg = "Missing config option %s" % m_ar[-1].strip()
```
xsuchy: I would rather see `.format()` here.

mock/py/mockbuild/util.py
```diff
@@ -1238,12 +1274,16 @@ def setup_operations_timeout(config_opts):
     _OPS_TIMEOUT = config_opts.get('opstimeout', 0)
 
 @traceLog()
-def do_update_config(log, config_opts, cfg, uidManager, name, skipError=True):
+def do_update_config(log, config_opts, cfg_paths, uidManager):
```
xsuchy: why the change of cfg to plural? And similarly in update_config_from_template(). The config is still only one file, isn't it?

xsuchy: You ommited name and skipError, but you still use it in load_config()

mock/py/mockbuild/util.py
```diff
@@ -1175,35 +1175,71 @@ def check_config(config_opts):
...
+    template = env.from_string(content)
+    part_old = None
+    while True:
+        part = template.render(config=config_opts)
+        code = compile(part, '<string>', 'exec')
         exec(code)
-    else:
-        raise exception.ConfigError("Could not find included config file: %s" % config_file)
+        if part_old == part:
+            break
```
xsuchy: Here should be same safety stop. I.e. break the loop after e.g., 1000 loops anyway. Ideally, this would be controlled by some config option.

xsuchy:
@plageat  I still did not get to the point where I actually executed your changes. So far I just read the code. I think this will not be easy merge and it will take some time, but I like this approach and I am really glad for this PR.

mock/py/mockbuild/util.py
plageat: Because here cfg argument is now a list of paths rather then one path:

`do_update_config(log, config_opts, config_opts['config_paths'], uidManager)`

mock/py/mockbuild/util.py
```diff
@@ -1277,17 +1317,18 @@ def load_config(config_path, name, uidManager, version, pkg_python_dir):
     config_opts['config_file'] = chroot_cfg_path
 
     cfg = os.path.join(config_path, 'site-defaults.cfg')
-    do_update_config(log, config_opts, cfg, uidManager, name)
-
-    do_update_config(log, config_opts, chroot_cfg_path, uidManager, name, skipError=False)
+    do_update_config_paths(log, config_opts, cfg, name)
```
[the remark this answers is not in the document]
plageat: Yes, because we do only 'list.append()' operation in function do_update_config_paths().

So the logic is to grab all config path, then grab all content of all paths as one, then do template rendering and do exec() of result to evaluate config

mock/py/mockbuild/util.py
plageat: This 'while True' loop is for the nested level references:

`config_opts['var1'] = 'value1'`
`config_opts['var2'] = {{ config_opts.var1  }}`
`config_opts['var3'] = {{ config_opts.var2  }}`

The loop will try to expand everything until nothing else is not expanded more

For code above 3 render loops:
- one for {{ config_opts.var1  }}
- second for {{ config_opts.var2  }}`
- third to get same rendering result and to stop

So, if you think there will be 1000 nested level references , consider to add such an option. But I really doubt of it

xsuchy: What will happen if:
```
config_opts['var1'] = '{{ config_opts.var2 }}'
config_opts['var2'] = '{{ config_opts.var1 }}'
```
?

plageat: This currently evaluates to:

`config_opts['var1'] = 'None'`
`config_opts['var2'] = 'None'`

praiskup:
On Tuesday, February 13, 2018 10:16:47 AM CET Mikhail Campos-Guadamuz wrote:
> > there's missing motivation
> 
> The changes will allow us:
> - a base for implementing parametrized config options in cmd (solve #6)
>   - this is very clear motivation 

Not clear to me, though.  You can have variables and functions already in
mock config, if needed.  So this only brings different way of doing the
same.  When I speak about motivation, I rather mean real use-cases though.

> - as a result - usability improvement - this is very clear motivation
> (if you do not agree with 2 above, please write to connected #6 not
> here)

I did, because I haven't seen valid use-case so far.

> - gathering all config option handling in one function instead of
>   scattering across all the files (need to be checked by someone
>   experienced) = code refactoring - this is very clear motivation

We had one function to handle the config options.  It seems to be 1:1
replacement.

> - jinja2 templating instead of 'random' string formatting - this is very
>   clear motivation

This is the only benefit I see, agreed.  But I wouldn't reimplement
something which works just to have jinja.

> - improve #138 constraints like for nested dict in config_opts - this is
> very clear motivation

Hmms, #138 is totally orthogonal.  With the actual code we could invent
ultimate fix for #138 same as with the proposal.

Ok, don't take me wrong -- I'm OK with jinja2 (I even proposed that in #6)
but still I haven't seen real problem solved by this PR which would
justify the need for new run dependency and relatively big change.  It
would be much, much easy to me to understand if you said "this is not
possible now" and it will be possible later (ideally in concrete example
configuration).  As a counter example, you invented templating system,
but you invent C&P change N times (`config_opts['useradd'] = ..).  Shouldn't
changes like that be ruled out by the templates so you don't have to set the
useradd command on N places?

mock/py/mockbuild/util.py
plageat: well, I can rewrite this to show something like:

mockbuild.exception.ConfigError: Missing config option 'var2'

praiskup: Uch, we did this (re-expanding templates) in distgen tool, and I'm still not convinced that calculating fixed point with jinja templates was a good step forward.  If that was the
right way of thinking, jinja would implement it itself.

plageat: >  jinja would implement it itself

jinja2 would never implement this because this is completely irrelevant to this mock's PR, which target here is to setup correct config with feature improvements ( jinja2 templating ) rather then simple rendering templates what jinja2 does. If you read and understood the PR (and mock's ) code, you would never write such a comment

xsuchy:
> Not clear to me, though. You can have variables and functions already in mock config, if needed. 

The problem is that you can have e.g.:

```
config_opts['plugin_conf']['ccache_opts']['dir'] = "%(cache_topdir)s/%(root)s/ccache/u%(chrootuid)s/"
```

But if you want to use %target_dir it will not work. And vice versa, you cannot use %cache_topdir in nearly any other variable. The replacement happens only for some variables and for some keys.

mock/py/mockbuild/util.py
praiskup: You underestimate what I say, jinja is not macro language, but templating language.  That's what I'm saying.

praiskup: Or in other words, you implement macro language with jinja, and that's clear misconception.

plageat: > and that's clear misconception

why?

praiskup:
Ok, I understand the movement now, @xsuchy thanks for the example!

The purpose of this PR is to implement make the configuration behave like macro language, so it makes us able to reference macros which are not yet defined and are going to be defined "later" in the file.  This deserves some form of "looping" around the configuration, but really jinja2?

praiskup:
Btw., yaml back-references might be worth looking at.

mock/py/mockbuild/util.py
praiskup: Because jinja language is not designed to be a macro language (tokens, AST interpolation, etc.).  It is designed for 1-hit templating.... Did you consider hiring macropy?

plageat: > Because jinja language is not designed to be a macro language (tokens, AST interpolation, etc.). It is designed for 1-hit templating

That's you vision and I did not argue with this.

But what we create for mock here using any tools (including jinja2) can consider as a little macro language, can't it? If something in the code world can't do something, then a programmer takes this as a part, extends with other code parts and gets what he want. Easy

praiskup: Just for reference, we touched some pros/cons with this approach here https://github.com/devexp-db/distgen/pull/41

plageat:
So, what is the desirable action for recursion then?

`config_opts['var1'] = '{{ config_opts.var2 }}'`
`config_opts['var2'] = '{{ config_opts.var1 }}'`

Issue #43, opened 2017-02-16
RFE: add support for .spec in --installdeps


Frodox:
Since both `yum-builddep` and `dnf builddep` supports installing deps for 
```
Install whatever is needed to build the given .src.rpm, .nosrc.rpm or .spec file.
```

I do not see any reasons why mock can't support smth like
```
$ mock -r epel7 --installdeps kernel.spec
INFO: mock.py version 1.3.3 starting (python version = 2.7.5)...
Start: init plugins
INFO: selinux enabled
Finish: init plugins
Start: run
ERROR: Cannot find/open srpm: kernel.spec.  Error: error reading package header
```

```
mock  [options] --installdeps {SRPM|RPM}
```

xsuchy:
Feel free to send patch :)

Frodox:
Ok, I'll try! :) Doesn't look very complex for start

plageat:
I think it is quite complex. There is only one problem to do that: making macro expansion for spec file using macro from buildroot, because of all this conditional BuildReqs in specfile (like %?arch, %?rhel, %?fedora etc.). Solutions:
1) Do it by python rpm:
   1.1 use rpm.spec(). Returns buildrequires, using macro from build system == incorrect result. Also, we can't just use external macro files because there is no such python rpm API (not found by me)
   1.2 so, find all macro files in buildroot, parse/process, substitute system macro by thise from buildroot (can be done with rpm.addMacro()/rpm.delMacro() one by one) = million hours of work and new errors
   1.3 or maybe define/redefine/undefine most common macro manually using info from mock config and undefine = I think the same as above
   1.4 hacking: import rpm in build system, chroot to buildroot, use rpm.reloadConfig() which will set all macro to buildroot's values, go back to build systemm, read specfile with rpm.spec and extract correct buildrequires = tested, does not work
2) Do it with yum-builddep in chroot:
   2.1 copy spec to buildroot, install yumbuildep, install dependencies, (also maybe undo packager transaction of installing yumbuildep or something)

   Any opinions?

xsuchy:
https://github.com/bkircher/python-rpm-spec it will need some improvements, but better than invent something new. And yes the evaluation needs to be done in chroot.

plageat:
> better than invent something new
If so, why don't use the old good yum-builddep?

python-rpm-spec - nice try, but no. I really doubt whether it is useful at all cause it works only as simple spec string parser without any rpm macro expansion (and implementing your own is complex enough)

[...]

PR #162, opened 2018-02-18
add support for .spec in --installdeps

ocasek:
add support for .spec in --installdeps #43

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -813,7 +813,7 @@ def run_command(options, args, config_opts, commands, buildroot, state):
         commands.init()
         rpms = []
         for file in args:
-            if file.split(".")[-1] == "spec":
+            if os.path.splitext(file)[1] == ".spec":
                 commands.installSpecDeps(file)
             else:
                 rpms.append(file)
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -810,10 +810,16 @@ def run_command(options, args, config_opts, commands, buildroot, state):
         if len(args) == 0:
             log.critical("You must specify an SRPM file with --installdeps")
             sys.exit(50)
-
-        util.checkSrpmHeaders(args, plainRpmOk=1)
         commands.init()
-        commands.installSrpmDeps(*args)
+        rpms = []
+        for file in args:
+            if file.split(".")[-1] == "spec":
+                commands.installSpecDeps(file)
+            else:
+                rpms.append(file)
+        if rpms:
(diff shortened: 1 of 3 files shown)


plageat:
How it parse condition buildreqs ?

`%if %{use_python3}`
`BuildRequires: python3-pylint`
`%endif`

plageat:
Actually, any solid argument for using external pyrpm package, at least one?

It seems that all the functionality have been implemented in standard rpm python binding module + much more features like macro expansion.
For example, for simple incorrect buildreqs is:

`import rpm`
`rpm.spec('target.spec').sourceHeader.dsFromHeader()`

ocasek:
Okay I'll implement intern functionality. I have used hints in issue. Thanks for info.

mock/py/mock.py
```diff
@@ -810,10 +810,16 @@ def run_command(options, args, config_opts, commands, buildroot, state):
         if len(args) == 0:
             log.critical("You must specify an SRPM file with --installdeps")
             sys.exit(50)
-
-        util.checkSrpmHeaders(args, plainRpmOk=1)
         commands.init()
-        commands.installSrpmDeps(*args)
+        rpms = []
+        for file in args:
+            if file.split(".")[-1] == "spec":
```
xsuchy: I would rather use https://docs.python.org/2/library/os.path.html#os.path.splitext
```
>>> os.path.splitext("/foo/bar")
('/foo/bar', '')
>>> os.path.splitext("/foo/bar.gc")
('/foo/bar', '.gc')
```

xsuchy:
Additionally, please modify man page for `--installdeps` so it state SPEC files as well.

mock/py/mockbuild/backend.py
```diff
@@ -203,6 +204,18 @@ def installSrpmDeps(self, *srpms):
         finally:
             self.uid_manager.restorePrivs()
 
+    @traceLog()
+    def installSpecDeps(self, spec_file):
+        try:
+            spec=rpm.spec(spec_file).sourceHeader.dsFromHeader()
+            self.uid_manager.becomeUser(0, 0)
+            for i in range(len(spec)): 
+                requirement_name = spec[i].split(" ")[1]
```
xsuchy: I did not test this yet, but this will likely not work for something like `BuildRequires: mock-core-configs > 26`. Can you try it?

ocasek: It will ignore everything, after name of the package.

plageat:
Nice changes. Now consider opening a new bug about reqs inside rpm macros...

PR #163, opened 2018-02-22
merge mockchain to mock

mnecas:
This PR is created to merge mockain.py to mock.py

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -557,7 +557,7 @@ def unshare_namespace(config_opts):
 
 
 @traceLog()
-def main(rpms=None, resultdir=None):
+def main():
     "Main executable entry point."
 
     # initial sanity check for correct invocation method
@@ -582,11 +582,6 @@ def main(rpms=None, resultdir=None):
     (options, args) = command_parse()
     if options.printrootpath or options.list_snapshots:
         options.verbose = 0
-
-    if rpms:
-        args = rpms
-    if resultdir:
-        options.resultdir = resultdir
 
     # config path -- can be overridden on cmdline
     config_path = MOCKCONFDIR
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 usage:
        mock [options] {--init|--clean|--scrub=[all,chroot,cache,root-cache,c-cache,yum-cache,dnf-cache,lvm,overlayfs]}
        mock [options] [--rebuild] /path/to/srpm(s)
+       mock [options] [--chain] /path/to/srpm(s)
        mock [options] --buildsrpm {--spec /path/to/spec --sources /path/to/src|
        --scm-enable [--scm-option key=value]}
        mock [options] {--shell|--chroot} <cmd>
@@ -117,6 +118,9 @@ def command_parse():
     parser.add_option("--rebuild", action="store_const", const="rebuild",
                       dest="mode", default='__default__',
                       help="rebuild the specified SRPM(s)")
+    parser.add_option("--chain", action="store_const", const="chain",
+                      dest="mode",
+                      help="build multiple RPMs in chain loop")
     parser.add_option("--buildsrpm", action="store_const", const="buildsrpm",
                       dest="mode",
                       help="Build a SRPM from spec (--spec ...) and sources"
@@ -199,9 +203,6 @@ def command_parse():
                       dest="mode", help="Mount the buildroot if it's "
                                         "mounted from separate device (LVM/overlayfs)")
     # chain
-    parser.add_option("--chain", action="store_true", dest="chain",
-                      default=False,
-                      help="if you want put multiple RPMs")
     parser.add_option('--localrepo', default=None,
                       help="local path for the local repo, defaults to making its own")
     parser.add_option('-c', '--continue', default=False, action='store_true',
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -220,9 +220,6 @@ def command_parse():
                       help="log to the file named by this option, defaults to not logging")
     parser.add_option('--tmp_prefix', default=None, dest='tmp_prefix',
                       help="tmp dir prefix - will default to username-pid if not specified")
-    parser.add_option('-m', '--mock-option', default=[], action='append',
-                      dest='mock_option',
-                      help="option to pass directly to mock")
     # options
     parser.add_option("-r", "--root", action="store", type="string", dest="chroot",
                       help="chroot config file name or path. Taken as a path if it ends "
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -46,16 +46,22 @@ import logging
 import logging.config
 # pylint: disable=deprecated-module
 from optparse import OptionParser
+
+from pprint import pformat
 import os
 import os.path
 import pwd
-import shlex
 import shutil
 import sys
 import time
+import cgi
+import tempfile
+import requests
 
-from pprint import pformat
+# pylint: disable=import-error
+from six.moves.urllib_parse import urlsplit
 from six.moves import configparser
+from mockbuild import util
 
 
 # all of the variables below are substituted by the build system
@@ -76,7 +82,7 @@ log = logging.getLogger()
 
 # our imports
 # pylint: disable=wrong-import-position
-from mockbuild import util
+
 import mockbuild.backend
 from mockbuild.backend import Commands
 from mockbuild.buildroot import Buildroot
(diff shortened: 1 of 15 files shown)


mock/py/mock.py
```diff
@@ -619,9 +647,231 @@ def unshare_namespace(config_opts):
             else:
                 sys.exit(e.resultcode)
 
+@traceLog()
+def chain():
+    mockconfig_path = '/etc/mock'
```
xsuchy: you can use MOCKCONFDIR variable here.

mock/py/mock.py
```diff
@@ -619,9 +647,231 @@ def unshare_namespace(config_opts):
...
+
+    def createrepo(path):
+        if os.path.exists(path + '/repodata/repomd.xml'):
+            comm = ['/usr/bin/createrepo_c', '--update', path]
+        else:
+            comm = ['/usr/bin/createrepo_c', path]
```
xsuchy: This would be nice to merge with config_opts["createrepo_command"] and post_build()

mock/py/mock.py
```diff
@@ -619,9 +647,231 @@ def unshare_namespace(config_opts):
...
+            with open(destfile, 'w') as br_dest:
+                for k, v in list(config_opts.items()):
+                    br_dest.write("config_opts[%r] = %r\n" % (k, v))
+            return True, ''
+        except (IOError, OSError):
+            return False, "Could not write mock config to %s" % destfile
+
+        return True, ''
+
+    def log(lf, msg):
```
xsuchy: Hmm maybe replace this with logOutput() from util.py?

mock/py/mock.py
```diff
@@ -619,9 +647,231 @@ def unshare_namespace(config_opts):
...
+                            for chunk in r.iter_content(4096):
+                                fd.write(chunk)
+                except Exception as e:
+                    log(opts.logfile, 'Error Downloading %s: %s' % (url, str(e)))
+                    failed.append(url)
+                    continue
+                else:
+                    downloaded_pkgs[pkg] = url
+            log(opts.logfile, "Start build: %s" % pkg)
+            main([pkg])
```
xsuchy: This can be done better. :)

mock/py/mock.py
```diff
@@ -619,9 +647,231 @@ def unshare_namespace(config_opts):
...
+            log(opts.logfile, "Packages successfully built in this order:")
+        for pkg in built_pkgs:
+            log(opts.logfile, pkg)
+    return return_code
+
+
 
 @traceLog()
-def main():
+def main(rpm=None):
```
xsuchy: I would prefer rpms as accept list.

mock/py/mock.py
```diff
@@ -644,7 +894,8 @@ def main():
         uidManager.dropPrivsTemp()
 
     (options, args) = command_parse()
-
+    if rpm is not None:
+        args=rpm
```
xsuchy: Originall in args can be only one rpm. You need to run `run_command(options, args, config_opts, commands, buildroot, state)` in a loop.

mock/tests/testenvironment
```diff
@@ -14,7 +14,7 @@ outdir=${TOPDIR}/mock-unit-test
 cfgdir=${TOPDIR}/../mock-core-configs/etc/mock
 #MOCKCMD="sudo ./py/mock.py $VERBOSE --resultdir=$outdir --uniqueext=$uniqueext --configdir=$cfgdir -r $testConfig --plugin-option=root_cache:age_check=False $MOCK_EXTRA_ARGS"
 MOCKCMD="mock $VERBOSE --resultdir=$outdir --uniqueext=$uniqueext --configdir=$cfgdir -r $testConfig $MOCK_EXTRA_ARGS"
-MOCKCHAIN="mockchain -r $testConfig"
+MOCKCHAIN="mock -r $testConfig"
```
Conan-Kudo: Wouldn't you need `--chain` here?

ocasek: I don't need it, because when you put more than one rpm it will automatically call chain functionality and every test about mockchain uses more than one rpm.

xsuchy:
Please run:
```
pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/*
```
There are a few warining and error, which were not previously presented.

xsuchy:
Hmmm, this:
```
mock --chain tests/*src.rpm
```
does not work for me. :(

ocasek:
Yeah I have also this problem, when I compile mock.
No such option: --chain. Don't now right now where is the problem.

xsuchy:
That is not a problem. Build and install mock with: `tito build --test --rpm -i` and then try the command above. `--chain` is recognized, but the processing stops after A.src.rpm with some failure on B.src.rpm.

ocasek:
Now it should be working, but I don't like my solution some idea?

xsuchy:
I like it. It is pretty good. Only thing I dislike is catching too broad type Exception. It can be anything. And in case it is IOError, it will actually does not help to try to rebuild. It would be nice if you can catch only e.g., mockbuild.exception.BuildError or some limited list of exceptions.

xsuchy:
`mock --chain test-C-1.1-0.src.rpm test-B-1.1-0.src.rpm test-A-1.1-0.src.rpm` This should succeed (at the end), but your code goes only through first loop and then end. And in the result dir is not even binary test-C.rpm which succeed in first loop.

xsuchy:
I am getting:

```
Start: chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: skipping root_cache aging check
Start: unpacking root cache
Finish: unpacking root cache
INFO: enabled dnf cache
Start: cleaning dnf metadata
Finish: cleaning dnf metadata
INFO: enabled HW Info plugin
Mock Version: 1.4.9
INFO: Mock Version: 1.4.9
Start: dnf install
Config error: Parsing file failed: Source contains parsing errors: 'file:///var/lib/mock/fedora-27-x86_64-msuchy-8951/root/etc/dnf/dnf.conf'
        [line 66]:                 name=file:///var/tmp/mock-chain-msuchy-8951-es3viqii/results/default

        [line 67]:                 baseurl=file:///var/tmp/mock-chain-msuchy-8951-es3viqii/results/default

        [line 68]:                 enabled=1

        [line 69]:                 skip_if_unavailable=1

        [line 70]:                 metadata_expire=0

        [line 71]:                 cost=1

        [line 72]:                 best=1

```

Please try `tito build --test --rpm -i` and then `mock tests/test-C-1.1-0.src.rpm tests/test-B-1.1-0.src.rpm tests/test-A-1.1-0.src.rpm`

ocasek:
I have added to calling just my confg "-r /etc/mock/fedora-27-x86_64.cfg" and its working right don't know where is problem but I'll look on it.

xsuchy:
rebased

xsuchy:
Rebased. And squashed to one commit. I fixed several issues, but it still does not work as expected. Therefore not ready for merge.

sergiomb2:
Join to this issue since https://github.com/rpm-software-management/mock/pull/155 was closing in favour of this

sergiomb2:
Hi, I'm a little late for the party and also haven't much time but finally some days ago I could picture what I'd like to do. The logic is the logic of search and replace , search for include and replace it the text of file, just after we evaluate the code ... 

3 commits here [1] the main one is [2] 

[1]
https://github.com/sergiomb2/mock/commits/simplify_include
[2]
https://github.com/sergiomb2/mock/commit/5653ab65f1f00ec5452e1564f9e7503973d27a88

xsuchy:
@sergiomb2 I really do not want to continue with special keywords and special handling. We can now use Jinja2 in configs. So you can use Jinja2 include http://jinja.pocoo.org/docs/2.10/templates/#include

xsuchy:
The current state is stat `11-mockchain-success.tst` is passing. 

But partial builds returns the incorrect exit code and if packages are in the wrong order, then chain will not try the next loop. I will continue on this one.

xsuchy:
OK. It should work now. I will let it sit here for some time for anyone willing to test/comment and then I will merge it.

Changes:
* when only some build were successfully built, mockchain previously returned exit code 2. This code is used in mock for something else, so mock return in this situation exit code 4.
* Some wording has been changed, e.g. "pkgs" changed to "packages".
* When some packages has been skipped, mock now print the summary which ones have been skipped.
* The files "success" and "fail" in result_dir had previously content "done" and "undone". Now, they are empty.

sergiomb2:
Hi ,

> @sergiomb2 I really do not want to continue with special keywords and special handling. We can now use Jinja2 in configs. So you can use Jinja2 include http://jinja.pocoo.org/docs/2.10/templates/#include

I need more information about this please, where or when we use the new (I guess) Jinja2 in config ? 
we need fix fedora-review to work with include [1] .

[1]  https://pagure.io/FedoraReview/issue/359

xsuchy:
@sergiomb2 https://github.com/rpm-software-management/mock/wiki/Release-Notes-1.4.14#mock-new-features

PR #167, opened 2018-03-06
overlayfs - new snapshot plugin

zzambers:
This new plugin implements mock's snapshot functionality using overlayfs. From user perspective it works similar to LVM plugin, but unlike LVM plugin, it only needs directory (not a volume group) for it's  data (snapshots). Plugin has no aditional dependencies, it only requires kernel with overlayfs support, but this is the case for both current Fedora and RHEL-7.

For more more information, see detailed description in plugin file:
mock/py/mockbuild/plugins/overlayfs.py

### DIFF
--- a/mock/py/mockbuild/plugins/overlayfs.py
+++ b/mock/py/mockbuild/plugins/overlayfs.py
@@ -2,10 +2,10 @@
 # This file is part of overlayfs plugin for mock
 # Copyright (C) 2018  Zdeněk Žamberský ( https://github.com/zzambers )
 #
-# This program is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License
-# as published by the Free Software Foundation; either version 2
-# of the License, or (at your option) any later version.
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
 #
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
@@ -13,8 +13,7 @@
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
 
@@ -29,8 +28,8 @@
 # config_opts['plugin_conf']['root_cache_enable'] = False
 # config_opts['plugin_conf']['overlayfs_enable'] = True
 # config_opts['plugin_conf']['overlayfs_opts']['base_dir'] = /some/directory
-# config_opts['plugin_conf']['overlayfs_opts']['debug'] = False
 # config_opts['plugin_conf']['overlayfs_opts']['touch_rpmdb'] = False
+# config_opts['plugin_conf']['overlayfs_opts']['trace_hooks'] = False
 #
 # ( Plugin uses postinit snapshot, similary to LVM, root chache is pointless. )
 #
@@ -38,15 +37,18 @@
 #            asociated with snapshots (layers, refs etc., see lower for details)
 #            It is further namespaced by configname so the same directory can be
 #            used in multiple mock configs without problems.
-# debug - print debug info, default: False
 # touch_rpmdb - automatically "touch" rpmdb files after each mount to copy
 #               them to upper layer to overcome rpm/yum issue,
 #               when calling them directly in chroot. See:
 #                   https://bugzilla.redhat.com/show_bug.cgi?id=1213602
+# pylint: disable=line-too-long
 #                   https://docs.docker.com/storage/storagedriver/overlayfs-driver/#limitations-on-overlayfs-compatibility
+# pylint: enable=line-too-long
 #               ( Option is not required when installing using mock --install,
 #               issue is work-arounded automatically there)
 #               defult: False
+# trace_hooks - print info messages about plugin's hooks being called,
+#               default: False
 #
 # plugin's resources asociated with config can be released by:
 #     mock -r <config> scrub all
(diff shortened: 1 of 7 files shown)


mock/py/mockbuild/plugins/overlayfs.py
```diff
@@ -0,0 +1,815 @@
...
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```
xsuchy: Nowadays is preferred to use electronic address instead of the physical, which already changed few times. See "Copyright notice" at https://www.gnu.org/licenses/gpl-howto.html

mock/py/mockbuild/plugins/overlayfs.py
```diff
@@ -0,0 +1,815 @@
...
+# perspective it works similar to LVM plugin, but unlike LVM plugin there is no
+# need for LVM volume. It only needs (base) directory, where internal data are
+# stored (layers etc., see lower).
+#
+# Configuration:
+# config_opts['plugin_conf']['root_cache_enable'] = False
+# config_opts['plugin_conf']['overlayfs_enable'] = True
+# config_opts['plugin_conf']['overlayfs_opts']['base_dir'] = /some/directory
+# config_opts['plugin_conf']['overlayfs_opts']['debug'] = False
+# config_opts['plugin_conf']['overlayfs_opts']['touch_rpmdb'] = False
```
xsuchy: Can this example go to mock/etc/mock/site-defaults.cfg as well?

mock/py/mockbuild/plugins/overlayfs.py
```diff
@@ -0,0 +1,815 @@
...
+        counter = self.getLayerRefcount(layerId)
+        counter += 1
+        self.setLayerRefCount(layerId, counter)
+        return counter
+
+    def unrefLayer(self, layerId):
+        counter = self.getLayerRefcount(layerId)
+        if counter <= 0:
+            # should not happen
+            raise Exception("refcounter is already <= 0: " + layerId +  " !")
```
xsuchy: Can you use `.format()`?

xsuchy: The same on several other places bellow.

mock/py/mockbuild/plugins/overlayfs.py
```diff
@@ -0,0 +1,815 @@
...
+
+        workDir = self.getWorkDir()
+        if os.path.exists(workDir):
+            shutil.rmtree(workDir)
+        os.mkdir(workDir)
+
+        # make sure kernel has required module loaded
+        modprobeCmds = []
+        modprobeCmds.append("modprobe")
+        modprobeCmds.append("overlay")
```
xsuchy: Can be done in one step, which is more effecient `modprobeCmds = ["modprobe", "overlay"]`

mock/py/mockbuild/plugins/overlayfs.py
```diff
@@ -0,0 +1,815 @@
...
+        except OSError:
+            raise Exception("Failed to obtain mount lock !")
+
+    def mountUnlock(self):
+        mountLockFile = self.getMountLockFile()
+        os.rmdir(mountLockFile)
+
+    def debugPrint(self, message):
+        if self.debug:
+            sys.stderr.write("DEBUG: Overalyfs pluin: " + message + "\n")
```
xsuchy: Instead of direct print to stdderr, it is better to use set up logger (which can point somewhere else)
`self.buildroot.root_log.debug("foo")`

xsuchy:
I just briefly checked the code. I am ill right now. I will do a full review in a few days. Generally, I like it and I want to merge it once the code is good. Also please run pylint-3 on your code, you have there a lot of stylistic "errors" `pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/*`

zzambers:
Thank you for your quick response (I hope you are not ill after seeing my code :) ).

I have done changes you requested. I also fixed all pylint-3 complains, except for one, where line is too long due to long URL link (I disabled check for that line).

zzambers:
I also tried using debug messages, but found out that these one are not really well suitable, for what I want. What I need here is quick way to see when and which plugin's hooks are being called. I have found, that debug messages are not printed to console by default. It is also way too much other debug info in log and logs are removed, when scrub/clean is called. So I ended up using info messages and renamed option to trace_hooks.

zzambers:
And 2 more things:

1. I found out (realized) that currently mock --mount does not work with my plugin (root does not remain mounted), It is unmounted at the end as for other commands ( I'll look into that later, maybe in separate PR. I need to think that trough. )

2. I also have something like unit test, which calls methods of plugin (snapshot/layers/refs related operations). And validates integrity of internal data/file structures being created. It does not actually mount anything and does not require privileged user, but it only operates on single file (one with plugin's class).  But I haven't seen similar tests in mock, so I am not sure, how this one could be integrated into mock project. Test file can be found here:
https://github.com/zzambers/random-files/blob/master/mock-overlayfs/test.py

xsuchy:
You can put it in `tests/`. Create there 15-foo.tst - it is shell script and it can run anything. Event that test.py.

xsuchy:
The logging is driven by `-vvvv` and can be tunned by `/etc/mock/logging.ini`.

zzambers:
I have added test file together with shell wrapper. ( I have done similar fixes in test file as for plugin file before, so hopefully there should not be same problems... )

xsuchy:
Merged. It would be nice, if you can extend the Plugins documentation on wiki: https://github.com/rpm-software-management/mock/wiki#plugins

zzambers:
Thank you.
Sure, I'll write something to wiki, when I find some time.

PR #177, opened 2018-04-05
Change sign plugint to sign only builded rpm and not every in results

ocasek:
https://bugzilla.redhat.com/show_bug.cgi?id=1217495

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -498,7 +498,7 @@ class Commands(object):
         bd_out = self.make_chroot_path(self.buildroot.builddir)
         results = glob.glob(bd_out + '/RPMS/*.rpm')
         results += glob.glob(bd_out + '/SRPMS/*.rpm')
-        self.buildroot.final_rpm_list = results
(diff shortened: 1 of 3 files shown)


xsuchy:
The list is e.g., ['/var/lib/mock/fedora-26-x86_64/root/builddir/build/RPMS/mock-1.4.9-1.git.26.a7dcdfa.fc26.noarch.rpm', '/var/lib/mock/fedora-26-x86_64/root/builddir/build/RPMS/mock-scm-1.4.9-1.git.26.a7dcdfa.fc26.noarch.rpm', ...] i.e. full path.
But before the rebuild_package() and the postbuild hook is called copy_build_results(), which copy it to result dir. I.e. after this change you are signing packages which are then not used at all and only deleted.

I recommend to store only the base name of files https://docs.python.org/3/library/os.path.html#os.path.basename and in sign.py you have to prefix it with result dir: '%s/*.rpm' % self.buildroot.resultdir

xsuchy:
Merged. Thank  you.

Frodox:
Here *must* be signing by chunks, because some packages (like texlive) have 4000+ rpms in result, so it would be too long arguments line for shell

Issue #169, opened 2018-03-10
Add support for bsdtar


tpgxyz:
### Short description of the problem
Currently mock supports tar from GNU, while bsdtar from libarchive is not supported

### Steps to reproduce issue

1. Set up bsdtar as default tar
2. Observe mock failure
3.

Do not forget to mention full commandline with the mock command you executed.

### Any additional notes

Consider adding output of `mock --debug-config` this can help developers to reproduce the issue.

xsuchy:
What would be the benefit? I am afraid that this is soooo low prio, that it will not happen unless someone (you) send a patch.

praiskup:
GNU tar is supposed to be the default on GNU/Linux distro.  Btw.:
```bash
$ rpm -q mock
mock-1.4.9-1.fc27.noarch
```

praiskup:
Is this issue related to mock scm plugin?

tpgxyz:
@xsuchy  @praiskup All calls to tar inside mock code are not compatible with bsdtar https://github.com/libarchive/libarchive

Conan-Kudo:
To give a more detailed explanation here:

OpenMandriva is in the process of switching to rpm.org rpm and DNF. As a part of this, their build system build engine is changing to mock. However, they replaced GNU tar with libarchive's tar back in OpenMandriva Lx 3, and libarchive's `tar` works slightly differently.

A proper patch for this would probably involve making it configurable to use either `tar` (expecting GNU tar) or `bsdtar` for libarchive's `tar`. What would be required for that, I don't yet know, but that's what I think would be required.

praiskup:
On Tuesday, April 24, 2018 11:21:42 AM CEST Neal Gompa (ニール・ゴンパ) wrote:
> > > What would be required for that, I don't yet ...
> >
> > Where this is discussed?
> 
> I'm not sure what you mean?

I thought that it is being discussed somewhere, and you'll know sooner or later.
Having some reference links would be nice.

Conan-Kudo:
No, this was discussed on IRC in `#OpenMandriva-Cooker` some time ago. The switch to `bsdtar` occurred two years ago, and GNU tar is no longer in the base system package set at all.

[...]

PR #180, opened 2018-04-24
call 'gtar' binary, instead of 'tar'

praiskup:
This helps OpenMandriva where /bin/tar is equivalent for
/bin/bsdtar .  That setup is somewhat unexpected on GNU/Linux
distro, but anyway, it doesn't hurt Fedora/RHEL/CentOS where we
symlink gtar to tar.

As original issue suggests, we could implement everything using
the POSIX tar options (+ helper scripting); but that would be just
too much work which is not worth spending ATM, at least as long as
it is safe to expect that 'gtar' is in PATH and that it really is
GNU tar.

Resolves: #169


praiskup:
Closing in favor of #181.

PR #181, opened 2018-04-24
add support for bsdtar

xsuchy:
Resolves: #169 

@praiskup @Conan-Kudo  what about this one?

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -113,6 +113,10 @@ config_opts['bootstrap_module_install'] = []
 # config_opts['createrepo_on_rpms'] = False
 # config_opts['createrepo_command'] = '/usr/bin/createrepo_c -d -q -x *.src.rpm'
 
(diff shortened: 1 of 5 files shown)


praiskup:
If you both (mock maintainer and OP) prefer this approach, why not.  `s/__cmd/__tar_cmd/` (nit).  I'd prefer one config option less, and slightly more consistent behavior across installations (plus I tend to prefer GNU tar usually).

praiskup:
Oh, maybe combine it with #180; I mean `s/__cmd = "tar"/__cmd = "gtar"/`, and mock scm should be fixed too, there are other bsdtar incompatible options.

Conan-Kudo:
I like this approach, and I agree with @praiskup that it should be combined with the other one so that we consistently call `gtar` vs `bsdtar`.

xsuchy:
Updated.

Conan-Kudo:
I agree with @praiskup. Looks good, but we should account for scm plugin. If you don't want to support `bsdtar` there, then just make sure it always calls `gtar`.

PR #184, opened 2018-05-10
allow bind-mounts in bootstrap for local repos

wenningerk:
Don't know if there is a more elegant way to allow local repos ('file://') with container-builds.
Other way to address this would be a new plugin that automatically takes care of everything addressed via 'file://' but people can probably be quite inventive opening up a lot of corner-cases. So this probably covers the generic case for the first while giving people all possibilities.

### DIFF
--- a/mock/py/mockbuild/plugins/bind_mount.py
+++ b/mock/py/mockbuild/plugins/bind_mount.py
@@ -32,7 +32,8 @@ class BindMount(object):
         self.state = buildroot.state
         self.bind_opts = conf
         # Skip mounting user-specified mounts if we're in the boostrap chroot
-        if buildroot.is_bootstrap:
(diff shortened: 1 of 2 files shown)


xsuchy:
I do not think this is a good approach. I would love to see a solution where mock will parse the baseurl and if there is `file://` then it will bind_mount that directory (even to bootstrap).

wenningerk:
Well, as I had written in the first paragraph I would like to see that approach without any manual actions needed as well. But I still expect cases where this can't be picked up automatically so that this suggestion would probably still be needed. And till the automatic thing is here it could serve as some kind of possibility to cope with the situation.
Though I do see that pulling this or something similar would reduce the pain to a level where it might not be high enough to do the 'nice thing' anymore ;-)

PR #185, opened 2018-05-16
WIP: Added support for MicroDnf

ArrayMy:
You have to add to config:

Determining path to MicroDnf:
config_opts['microdnf_command'] = '/usr/bin/microdnf'
Command on install with MicroDnf:
config_opts['microdnf_install_command'] = 'microdnf install'

MicroDnf cannot use source config at /etc/microdnf/. At this moment, MicroDnf uses source config of Dnf, but its structure is already prepared for moment, when it will be available. Also, MicroDnf can't use installroot and builddep params. You need to use Dnf for these options.

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # vim: noai:ts=4:sw=4:expandtab
 
+import copy
 import glob
 import os.path
 import shutil
@@ -42,6 +43,8 @@ in Mock config.""")
         # This will likely mean some error later.
         # Either user is smart or let him shot in his foot.
         return Dnf(config_opts, buildroot, plugins, bootstrap_buildroot)
+    elif pm == 'microdnf':
+        return MicroDnf(config_opts, buildroot, plugins, bootstrap_buildroot)
     else:
         # TODO specific exception type
         raise Exception('Unrecognized package manager')
@@ -53,6 +56,9 @@ class _PackageManager(object):
     install_command = None
     builddep_command = None
     resolvedep_command = None
+    # When support_installroot is False then command is run in target chroot
+    # you must ensure that `command` is available in the chroot
+    support_installroot = True
 
     @traceLog()
     def __init__(self, config, buildroot, plugins, bootstrap_buildroot):
@@ -79,7 +85,8 @@ class _PackageManager(object):
                 invocation = [self.command]
         else:
             invocation = [self.command]
-        invocation += ['--installroot', self.buildroot.make_chroot_path('')]
+        if self.support_installroot:
+            invocation += ['--installroot', self.buildroot.make_chroot_path('')]
         if cmd in ['upgrade', 'update', 'module']:
             invocation += ['-y']
         releasever = self.config['releasever']
@@ -92,6 +99,7 @@ class _PackageManager(object):
         invocation += common_opts
         invocation += args
         return invocation
+
 
     @traceLog()
     def execute(self, *args, **kwargs):
@@ -115,7 +123,9 @@ class _PackageManager(object):
             kwargs['pty'] = kwargs.get('pty', True)
         self.buildroot.nuke_rpm_db()
         try:
-            if self.bootstrap_buildroot is None:
+            if not self.support_installroot:
+                out = util.do(invocation, env=env, chrootPath=self.buildroot.make_chroot_path(), **kwargs)
+            elif self.bootstrap_buildroot is None:
                 out = util.do(invocation, env=env, **kwargs)
             else:
                 out = util.do(invocation, env=env, chrootPath=self.bootstrap_buildroot.make_chroot_path(), **kwargs)
@@ -188,9 +198,11 @@ def check_yum_config(config, log):
 
 class Yum(_PackageManager):
     name = 'yum'
+    support_installroot = True
 
     def __init__(self, config, buildroot, plugins, bootstrap_buildroot):
         super(Yum, self).__init__(config, buildroot, plugins, bootstrap_buildroot)
+        self.pm = config['package_manager']
         self.command = config['yum_command']
         self.install_command = config['yum_install_command']
         self.builddep_command = [config['yum_builddep_command']]
@@ -293,9 +305,11 @@ def _check_missing(output):
 
 class Dnf(_PackageManager):
     name = 'dnf'
+    support_installroot = True
 
     def __init__(self, config, buildroot, plugins, bootstrap_buildroot):
         super(Dnf, self).__init__(config, buildroot, plugins, bootstrap_buildroot)
+        self.pm = config['package_manager']
         self.command = config['dnf_command']
         self.install_command = config['dnf_install_command']
         self.builddep_command = [self.command, 'builddep']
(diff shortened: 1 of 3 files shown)


ArrayMy:
My apologies for doubled commits, It will be all fixed over night. Also, I'll fix all issues reported by PEP8.

Conan-Kudo:
We shouldn't need a mix of package managers for this to work. I think we should hold off on this until the needed functionality is in microdnf.

ArrayMy:
I agree

xsuchy:
@Conan-Kudo  The wait for `--installroot` can be quite long. We can do that in parallel. I see nothing wrong on the implementation when we initialize the root with dnf (especially when we document it) and only then we switch to use of microdnf to install dependencies.
The base installation is more or less the same no matter what PM we use. The reason why we fiddle with other PM is to catch potential differences in dep solving of BR and Requires of the package we are going to build.

mock/py/mockbuild/buildroot.py
```diff
@@ -413,6 +413,8 @@ def _setup_dirs(self):
                 'tmp/ccache',
                 'var/tmp',
                 'etc/dnf',
+                # After fixing the bug it will be useful
+                'etc/microdnf',
```
xsuchy: I would remove it for now, and only add it once it is merged into microdnf

mock/py/mockbuild/buildroot.py
```diff
@@ -516,6 +518,8 @@ def _setup_files(self):
         for item in [self.make_chroot_path('etc', 'fstab'),
                      self.make_chroot_path('etc', 'yum', 'yum.conf'),
                      self.make_chroot_path('etc', 'dnf', 'dnf.conf'),
+                     # After fixing the bug it will be useful
+                     self.make_chroot_path('etc', 'microdnf', 'microdnf.conf'),
```
xsuchy: dtto

mock/py/mockbuild/package_manager.py
```diff
@@ -19,6 +19,10 @@ def package_manager(config_opts, buildroot, plugins, bootstrap_buildroot=None):
     pm = config_opts.get('package_manager', 'yum')
     if pm == 'yum':
         return Yum(config_opts, buildroot, plugins, bootstrap_buildroot)
+    # Informations and call class for microdnf
+    if pm == 'microdnf':
+            print('When using MicroDnf, it is necessary to builddep and installroot using Dnf!')
```
xsuchy: You should use log() for this.

mock/py/mockbuild/package_manager.py
```diff
@@ -78,9 +82,13 @@ def build_invocation(self, *args):
             else:
                 invocation = [self.command]
                 common_opts = self.config[self.name + '_common_opts']
+        # Condition for using the dnf command at --installroot for microdnf
         else:
-            invocation = [self.command]
-            common_opts = self.config[self.name + '_common_opts']
+            if self.command == "/usr/bin/microdnf":
+                invocation = ['/usr/bin/dnf']
```
xsuchy: This is wrong.

mock/py/mockbuild/package_manager.py
```diff
@@ -338,3 +350,50 @@ def builddep(self, *pkgs, **kwargs):
...
+
+# class for microdnf
+class MicroDnf(_PackageManager):
+    name = 'microdnf'
+
+    def __init__(self, config, buildroot, plugins, bootstrap_buildroot):
+        super(MicroDnf, self).__init__(config, buildroot, plugins, bootstrap_buildroot)
+        self.command = config['microdnf_command']
+        self.install_command = config['microdnf_install_command']
+        self.builddep_command = [self.command, 'builddep']
```
xsuchy: I believe that this should be like:
`self.builddep_command =  config['dnf_builddep_command']`
which is better than the other option
self.builddep_command = config['dnf_command'], 'builddep']

mock/py/mockbuild/package_manager.py
```diff
@@ -338,3 +350,50 @@ def builddep(self, *pkgs, **kwargs):
...
+        if bootstrap_buildroot is None:
+            self._check_command()
+        self.resolvedep_command = [self.command, 'repoquery', '--resolve', '--requires']
```
xsuchy: dtto

Conan-Kudo:
Strictly speaking, we don't actually need most of this change. The intent is that the microdnf package can be invoked simply by changing `/usr/bin/dnf` to `/usr/bin/microdnf`.

That said, I have a PR open for installroot support, I just need help fixing the behavior of it.

mock/py/mockbuild/package_manager.py
```diff
@@ -338,3 +350,50 @@ def builddep(self, *pkgs, **kwargs):
                 if 'no such command: builddep' in line.lower():
                     raise BuildError("builddep command missing.\nPlease install package dnf-plugins-core.")
             raise
+
+
+# class for microdnf
+class MicroDnf(_PackageManager):
```
xsuchy: I believe it will be better if we inherit from DNF here.

mock/py/mockbuild/package_manager.py
```diff
@@ -338,3 +350,50 @@ def builddep(self, *pkgs, **kwargs):
                 if 'no such command: builddep' in line.lower():
                     raise BuildError("builddep command missing.\nPlease install package dnf-plugins-core.")
             raise
+
+
+# class for microdnf
```
xsuchy: This is definitely redundant. You should rather use docstring.

mock/py/mockbuild/package_manager.py
```diff
@@ -338,3 +350,50 @@ def builddep(self, *pkgs, **kwargs):
...
+    @traceLog()
+    def initialize_config(self):
+        if 'microdnf.conf' in self.config:
+            config_content = self.config['microdnf.conf']
+        else:
+            config_content = self.config['yum.conf']
+            config_content = self.replace_in_config(config_content)
+        check_yum_config(config_content, self.buildroot.root_log)
+        util.mkdirIfAbsent(self.buildroot.make_chroot_path('etc', 'microdnf'))
+        # Bug in microdnf , microdnf takes the repository resources from the same location as dnf
```
xsuchy: Can you reference the bug?

mock/py/mockbuild/package_manager.py
```diff
@@ -338,3 +350,50 @@ def builddep(self, *pkgs, **kwargs):
...
+        # After fix, set new general path
+        dnfconf_path = self.buildroot.make_chroot_path('etc', 'dnf', 'dnf.conf')
+        with open(dnfconf_path, 'w+') as dnfconf_file:
+            dnfconf_file.write(config_content)
```
xsuchy: Does this work with UTF8?

ArrayMy:
If possible, We'll talk about it tomorrow. Thanks for fix fixing my mistakes/bugs and I'll try to fix them all.
[This is the bug](https://bugzilla.redhat.com/show_bug.cgi?id=1578742)
[Code](https://github.com/ArrayMy/microdnf/commit/dfe97b94fa63c4cf91a6776d7e9de9406a217f30)

mock/py/mockbuild/package_manager.py
ArrayMy: I was looking at microdnf code today, And I realized that I have made a mistake.

xsuchy:
I updated this PR with commit which really works.

It would be nice to replace 'dnf builddep` with `rpm -qR src.rpm` and let microdnf install this list.

xsuchy:
Hmm. I will do that in separate commit and merge this for now.

PR #187, opened 2018-05-25
Allow --spec arg to be used with rebuild option

sfowl:
This is an unsolicited patch to allow for 'mock' to be used with the ```--spec``` option to ```rebuild``` (default) a package.

The use case is to support rebuilding a package with compilation options like ASAN (```-fsanitize=address```) to test reproducers for security vulnerabilities.

E.g.

```
# dnf download --source foo
# rpm2cpio foo.src.rpm | cpio -dimv
(Add/modify compilation flags in .spec file)
# mock --spec foo.spec foo.src.rpm --postinstall
```

Then spawn a shell in the chroot and test the newly compiled binaries with reproducer inputs.

I also considered adding an option like ```--buildbrpm```, but found that the current approach made it easier as a user to get to the 'ready-to-run-commands-stage'.

Feedback is welcomed! Suggestions on other methods to achieve the above is also welcomed.


mock/tests/test-A.spec
```diff
@@ -0,0 +1,35 @@
+Summary: test-A
+Name: test-A
+Version: 1.1
+Release: 0
+License: GPL
+Group: System Environment/Base
```
xsuchy: Remove the group. It is not used for ages.

mock/tests/test-A.spec
```diff
@@ -0,0 +1,35 @@
...
+%build
+
+%install
+[ "$RPM_BUILD_ROOT" != "/" ] && rm -rf $RPM_BUILD_ROOT
+mkdir -p $RPM_BUILD_ROOT/etc
+touch $RPM_BUILD_ROOT/etc/%{name}-installed
+echo "A is here" >> $RPM_BUILD_ROOT/etc/%{name}-installed
+
+%clean
+[ "$RPM_BUILD_ROOT" != "/" ] && rm -rf $RPM_BUILD_ROOT
```
xsuchy: %clean is not needed. It can be removed.

mock/tests/test-A.spec
```diff
@@ -0,0 +1,35 @@
...
+
+
+%prep
+
+%build
+
+%install
+[ "$RPM_BUILD_ROOT" != "/" ] && rm -rf $RPM_BUILD_ROOT
+mkdir -p $RPM_BUILD_ROOT/etc
+touch $RPM_BUILD_ROOT/etc/%{name}-installed
```
xsuchy: Why the touch? It will be created by the echo anyway.

mock/tests/test-A.spec
```diff
@@ -0,0 +1,35 @@
...
+echo "A is here" >> $RPM_BUILD_ROOT/etc/%{name}-installed
+
+%clean
+[ "$RPM_BUILD_ROOT" != "/" ] && rm -rf $RPM_BUILD_ROOT
+
+
+%files
+%defattr(-, root, root)
```
xsuchy: This is default. It can be remove. Make it simple.

mock/tests/test-A.spec
xsuchy: Can you use %{_sysconfdir} instead of /etc/?

mock/py/mockbuild/backend.py
```diff
@@ -431,6 +434,13 @@ def copy_srpm_into_chroot(self, srpm_path):
         shutil.copyfile(srpm_path, os.path.join(dest, srpmFilename))
         return os.path.join(self.buildroot.builddir, 'originals', srpmFilename)
 
+    @traceLog()
+    def copy_spec_into_chroot(self, spec_path):
+        specFilename = os.path.basename(spec_path)
+        dest = self.buildroot.make_chroot_path(self.buildroot.builddir, 'originals')
+        shutil.copyfile(spec_path, os.path.join(dest, specFilename))
```
xsuchy: I would say that shutil.copy2() would be a better choice. As the target file will not be owned by root:root but likely by the non-privileged user.

mock/tests/16-spec-and-multiple-srpms-fail.tst
```diff
@@ -0,0 +1,13 @@
...
+
+header "test --spec with multiple srpms failure"
+runcmd "$MOCK --spec ${TESTDIR}/test-A.spec ${TESTDIR}/test-A-1.1-0.src.rpm ${TESTDIR}/test-B-1.1-0.src.rpm"
+res=$?
+
+if [ $res -ne 50 ]; then
+   echo "mock returned success when it should have failed!"
+   exit 1
+fi
+exit 0
```
xsuchy: It will be nice to add one test which succeeds.

xsuchy:
Otherwise, I have no objection.

sfowl:
Thanks for the feedback. The latest commits should address all the review comments.

Conan-Kudo:
Could this be extended to support `--spec` + `--sources` to do a package build rather than having to do the `--buildsrpm` -> `--rebuild` manually?

sfowl:
de82836 implements using `--spec` and `--sources` to do a package build. However, I'm not familiar enough with either the code or the relevant use cases to fully appreciate the implications of this change. I particular have doubts over compatability with `scm` related uses and setting `config_opts['clean']` to `False`.

If significant changes need to be made in order for the requested extension using `--spec` and `--sources` to work, I would prefer for it to be addressed outside of this PR.

xsuchy:
Merged with minor changes in tests.

xsuchy:
Thank you.

PR #190, opened 2018-05-30
RFC: do not populate /etc/resolv.conf when networking is disabled (RHBZ#15…

tmzullinger:
…14028)

systemd-nspawn overrides the 'use_host_resolv' config and copies the
host /etc/resolv.conf to the chroot even when networking is disabled.

In most common cases, this causes any networking calls in the chroot to
timeout rather than fail quickly, resulting in significant delays.  For
example:

    <mock-chroot> sh-4.4# time curl http://example.com/
    curl: (6) Could not resolve host: example.com

    real	0m20.541s
    user	0m0.009s
    sys	0m0.008s

With an empty /etc/resolv.conf in the chroot, this is much quicker:

    <mock-chroot> sh-4.4# time curl http://example.com/
    curl: (6) Could not resolve host: example.com

    real	0m0.016s
    user	0m0.005s
    sys	0m0.008s

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -691,9 +691,10 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None, c
         cmd = [cmd]
     nspawn_argv = ['/usr/bin/systemd-nspawn', '-q', '-M', uuid.uuid4().hex, '-D', chrootPath]
     distro_label = distro.linux_distribution(full_distribution_name=False)[0]
-    if (distro_label != 'centos') and (distro_label != 'ol') and \
-       (distro_label != 'rhel') and (distro_label != 'deskos'):
-        # EL7 does not support it (yet). See BZ 1417387
+    distro_version = float(distro.version() or 0)
+    epel = ('centos', 'deskos', 'ol', 'rhel')
+    if distro_label not in epel or distro_version >= 7.5:
+        # EL < 7.5 does not support the nspawn -a option. See BZ 1417387
         nspawn_argv += ['-a']
     nspawn_argv.extend(nspawn_args)
     if cwd:
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -8,6 +8,7 @@
 from __future__ import print_function
 
 from ast import literal_eval
+import atexit
 import ctypes
 import errno
 import fcntl
@@ -27,6 +28,7 @@ import stat
 import struct
 import subprocess
 import sys
+import tempfile
 # pylint: disable=wrong-import-order
 import termios
 from textwrap import dedent
@@ -677,6 +679,16 @@ def is_in_dir(path, directory):
     return os.path.commonprefix([path, directory]) == directory
 
 
+def _nspawnTempResolvAtExit(path):
+    """Remove nspawn temporary resolv.conf from host."""
+    try:
+        os.remove(path)
+    except OSError as e:
+        if e.errno == errno.ENOENT:
+            pass
+        getLog().warning("unable to delete temporary resolv.conf (%s): %s", path, e)
+
+
 def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None, cwd=None):
     cmd_is_list = isinstance(cmd, list)
     if nspawn_args is None:
@@ -1301,12 +1313,19 @@ def load_config(config_path, name, uidManager, version, pkg_python_dir):
         '~' + pwd.getpwuid(os.getuid())[0]), '.config/mock.cfg')
     do_update_config(log, config_opts, cfg, uidManager, name)
 
-    # default /etc/hosts contents
-    if not config_opts['use_host_resolv'] and 'etc/hosts' not in config_opts['files']:
-        config_opts['files']['etc/hosts'] = dedent('''\
-            127.0.0.1 localhost localhost.localdomain
-            ::1       localhost localhost.localdomain localhost6 localhost6.localdomain6
-            ''')
+    if not config_opts['use_host_resolv']:
+        # default /etc/hosts contents
+        if 'etc/hosts' not in config_opts['files']:
+            config_opts['files']['etc/hosts'] = dedent('''\
+                127.0.0.1 localhost localhost.localdomain
+                ::1       localhost localhost.localdomain localhost6 localhost6.localdomain6
+                ''')
+        # bind mount an empty /etc/resolv.conf when using nspawn and networking is disabled
+        if config_opts['use_nspawn'] and not config_opts['rpmbuild_networking']:
+            resolv_fd, resolv_path = tempfile.mkstemp(prefix="mock-resolv.")
+            atexit.register(_nspawnTempResolvAtExit, resolv_path)
+            config_opts['nspawn_args'] += ['--bind={0}:/etc/resolv.conf'.format(resolv_path)]
+
     if config_opts['use_container_host_hostname'] and '%_buildhost' not in config_opts['macros']:
         config_opts['macros']['%_buildhost'] = socket.getfqdn()
     return config_opts


tmzullinger:
This is marck as an RFC, as I know it won't work well with EL7's systemd-nspawn as-is.  The `--bind` option in that version of systemd-nspawn lacks support for mounting files from within the chroot (the `+` used in the source path) -- at least according to the man page.

Before I go further with this, I thought I'd seek some input on whether it's a good direction.  It might not be worth the effort to create the empty file we bind mount in the chroot.  It could simply be on the host file system.  If so, I wonder what path might work best there.  Maybe `/var/tmp/mock/empty` would work.

Nothing great came to mind while I was working on this today and I implemented it using the nspawn bind with source inside the chroot before I noticed EL7 lacked that support.  (I did check --bind was available before starting, but not the `+$source_path` part).

mock/py/mockbuild/util.py
```diff
@@ -1300,12 +1300,18 @@ def load_config(config_path, name, uidManager, version, pkg_python_dir):
...
+        # default /etc/hosts contents
+        if 'etc/hosts' not in config_opts['files']:
+            config_opts['files']['etc/hosts'] = dedent('''\
+                127.0.0.1 localhost localhost.localdomain
+                ::1       localhost localhost.localdomain localhost6 localhost6.localdomain6
+                ''')
+        # bind mount an empty /etc/resolv.conf when using nspawn and networking is disabled
+        if config_opts['use_nspawn'] and not config_opts['rpmbuild_networking']:
+            config_opts['files']['var/tmp/empty'] = ''
+            config_opts['nspawn_args'] += ['--bind=+/var/tmp/empty:/etc/resolv.conf']
```
xsuchy: Can this be rather: `--bind=:/etc/resolv.conf'?
Quoting from man page:
```
or a colon-separated pair of paths — in which case the first
specified path is the source in the host, and the second path is the destination in the container,...
... 
The source path may be specified as empty string, in which case a temporary directory below the host's /var/tmp directory is used. It is
automatically removed when the container is shut down.
```

xsuchy:
Wow, this looks good. I have been trying to bind-mount it using bind-mount mock's plugin, but that does not work. I am going to test this soon.

mock/py/mockbuild/util.py
tmzullinger: I can try that.  Reading it, I thought it would bind mount a directory and that a dir as /etc/resolv.conf wouldn't work well.  But let's confirm that empirically. :)

xsuchy:
I will be fine with doing this step everywhere but on RHEL-7. I.e. test using python-distro what is the host system, and if it is RHEL-7 skip this part (so no-network will have to timeout as it does now) but everywhere else we will pass this argument.

mock/py/mockbuild/util.py
tmzullinger: Sadly, that does fail:

```
Start: shell
Cannot bind mount directory /var/tmp/nspawn-temp-8SCtjW/src on file /var/lib/mock/fedora-27-x86_64/root/etc/resolv.conf.
```

That's with `config_opts['nspawn_args'] += ['--bind=:/etc/resolv.conf']`.  I scrubbed the root-cache and chroot, but perhaps there's still an /etc/resolv.conf present which causes the error.  I do still wonder how many things would not be happy if nspawn would bind mount a dir at /etc/resolv.conf.

We can keep at that if you think it's going to be cleaner than mounting an empty file from the host (at least on EL7).  We could use `+/var/tmp/empty` on Fedora and EL8, but fall back to a file on the host for EL7, perhaps.  I'm not sure it's worth the extra code to do that, but it's an option.

tmzullinger:
That's not a bad option to start.  In `_prepare_nspawn_command()` there's already a list of EL7-compatible systems which are used to exclude the `-a` option.  We'd need to use that sort of list to exclude the --bind arg.  That list doesn't have any version info, so it's going to be outdated when RHEL-8 is released, it seems.  Fixing that to check not only the distribution name but version and then using that in both places would be a good preparatory patch I guess.

The other option is using tempfile.mkstemp to create a file and remove it, similar to `_selinuxCreateFauxFilesystems` and `_selinuxAtExit`.  Or just ship something like `/var/cache/mock/empty` and use it.  That doesn't seem too bad to me.

Since we know the path of the chroot dir on the host, I wonder if we can just pass the full path to the nspawn --bind option?  I'll give that a try.  I don't know offhand if the chrootPath is set where I have the `--bind` args now.  But it could even be done as a replacement once in `_prepare_nspawn_command()`.

tmzullinger:
Something like this (without fixing the lack of a version check used to add the `-a` option:

```
diff --git i/mock/py/mockbuild/util.py w/mock/py/mockbuild/util.py
index 316ed8b..034f15f 100644
--- i/mock/py/mockbuild/util.py
+++ w/mock/py/mockbuild/util.py
@@ -695,6 +695,9 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None, c
        (distro_label != 'rhel') and (distro_label != 'deskos'):
         # EL7 does not support it (yet). See BZ 1417387
         nspawn_argv += ['-a']
+    else:
+        # nspawn on EL7 lacks support for + in the source path arg to --bind
+        nspawn_args = [arg.replace('--bind=+', '--bind=%s' % chrootPath) for arg in args]
     nspawn_argv.extend(nspawn_args)
     if cwd:
         nspawn_argv.append('--chdir={0}'.format(cwd))
```

After lunch I'll test this properly on an EL-7 host.

tmzullinger:
It might also be worth considering whether `use_host_resolv` should still default to true when `use_nspawn` is set.  At least for the resolv.conf portion, the defaults there result in a less-than-ideal experience.  I'm not sure if having the hosts file in place without networking is something people use.

xsuchy:
Ahhh it is directory. I did not noticed that. I thought that is is file. Then your original solution would be fine.
I just wonder why use '+'. We can create some tempfile on host using `mktemp()` and then bindmount that to nspawn. It does not need to be inside of container.

tmzullinger:
Yeah, there are certainly many ways to tackle this.  :)

I initially tested with a file on the host (with just a quick and dirty `touch /etc/mock/empty`).  I moved to a file in the chroot mainly because it was quicker and cleaner to implement without leaving cruft on the host.

If the file is on the host, then we either need to ensure it's created as part of the mock install or ensure it exists it each time a chroot is created (and possibly remove it when the chroot is removed).

I don't much care whether it's on the host or in the chroot.  If the current code looks like it will be more work to maintain over time compared to what it might look like to create the file on the host, I'm happy to work on changing it.  I don't think the current code is too bad though.  And when EL7 is dropped or systemd-nspawn learns the `--bind +$source_path` option, the 2 lines needed to conditionally replace the `+` with the chroot path can be removed, leaving only the few lines in `utils.locad_config()` to implement this.

If you'd prefer to explore using a file on the host for the bind mount, I have a few questions about how we might best implement it.  These are mostly a matter of preferences.

Would we create the file on the host once at mock install time and use it repeatedly?  If so, what path might be best?  (Either `/var/cache/mock` or `/var/tmp/mock` seem like good candidates.  The former dir, of course, exists already but it's a bit of a stretch to call an empty file like this a 'cache' file.  The latter is more sensible for temp files, but we may need a tmpfiles.d entry to manage the dir.)

Or would we create the file on-demand when the chroot is created?  If so, are there any existing examples in the mock code of that setup/teardown pattern for a temp file on the host that I can use/copy to keep things consistent?  (I'm comfortable with tempfile.mkstemp and friends, I just don't want to do it differently than other examples in mock and make maintenance more difficult.)

Thanks.

tmzullinger:
Here's a version using an on-host temp file.  I didn't remove the `add version to EL check in _prepare_nspawn_command()` commit from my branch, but it's now now at all related to this issue.  I think it's a worthwhile cleanup, but if you prefer I can move it to another branch and file a separate PR for it. :)

xsuchy:
Thank you.

tmzullinger:
Thanks @xsuchy.  What do you think of a follow-up to this which sets `config_opts['use_host_resolv'] = False` by default?  I think that works better along with the 'rpmbuild_networking' & 'use_nspawn' defaults and provides a nicer experience to users without requiring any changes to the site config.

I did this in https://github.com/tmzullinger/mock/commit/fad91dd83 and will be happy to file a separate PR for that if you don't have any strong objections to changing the default.

xsuchy:
Yes. Please file a PR.

PR #193, opened 2018-06-05
Allow root

nhorman:
I have a project (https://github.com/nhorman/freight) which requires that some commands be run as the root user.  unfortunately, mock is not currently able to run commands as uid/gid 0 / user name root.  This set of patches enhances mock such that the uid/gid/username can be set in the config at build time.  Teste successfully by myself building freight container rpms

Signed-off-by: Neil Horman <nhorman@tuxdriver.com>

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -359,6 +359,8 @@ config_opts['bootstrap_module_install'] = []
 # name in this line of the mock PAM config:
 #   auth  sufficient pam_succeed_if.so user ingroup mock use_uid quiet
 # config_opts['chrootgid'] = grp.getgrnam("mock")[2]
+# name of user that is used when executing commands inside the chroot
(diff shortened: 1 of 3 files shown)


nhorman:
I can fix those up.  Would you like a new pr, or shall I just add a patch to this one?

nhorman:
Note those pep8 errors in util.py do not appear to be mine.  I'm happy to fix them if you like, but I'm not sure thats appropriate for this commit set

xsuchy:
Ignore the pep8speak for the errors that are not yours. That's a bot I recently added. 

The PR is just fine. It does not change anything for normal users. 

Can you add the default to `mock/etc/mock/site-defaults.cfg` with some comment? Everything commented out.

nhorman:
Can do!

nhorman:
handled, site-defaults.cfg now contains the documentation and default setting for chrootuser

xsuchy:
Thank you.

PR #195, opened 2018-06-07
Disable use_host_resolv by default

tmzullinger:
Mock has disabled networking by default since be4b13a ("Add
"rpmbuild_networking" key (False by default) for nspawn backend",
2015-07-27).  When using a systemd-nspawn chroot (the default), the host
resolv.conf is copied to the container by nspawn when 'use_host_resolv'
is True, which causes spurious network timeouts as noted in
RHBZ#1514028.

When the variable is unset, mock avoids populating it since 3f93978 ("do
not populate /etc/resolv.conf when networking is disabled
(RHBZ#1514028)", 2018-05-30).

Disabling 'use_host_resolv' by default provides a better out-of-the-box
experience.  The variable will be enabled when '--enable-network' option
is used, similar to the 'rpmbuild_networking' variable.

This is a follow-up to #190.

### DIFF
--- a/mock/docs/mock.1
+++ b/mock/docs/mock.1
@@ -239,6 +239,7 @@ It can be specified multiple times.
 .TP
 \fB\-\-enable\-network\fR
 Enable networking. If you want to have reproducible builds then your builds should run without a network.
+This option overrides config_opts['rpmbuild_network'] and config_opts['use_host_resolv'], setting both True.
 .TP
 \fB\-h\fR, \fB\-\-help\fR
 Show usage information and exit.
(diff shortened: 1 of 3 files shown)


tmzullinger:
I wonder whether we should document that `--enable-network` overrides this setting.  I started to add such documentation a few times and couldn't convince myself I had the right wording or location.  A note about this behavior could be placed in the man page and/or in the example site-defaults.cfg.  We don't note that `rpmbuild_networking` is similarly overridden, but it's arguably much more obvious that is the intention of the `--enable-network` option.

I'm not sure if overriding this will cause anyone trouble or surprise, but I've done this long enough to presume to know how everyone else uses the tools. ;)

xsuchy:
Thank you.

PR #196, opened 2018-06-08
fix @VERSION@ processing in man pages

ktdreyer:
Prior to this change, the rpm build process would replace all single `@` characters in the man pages with the version number.

The problem is that Perl treats `@VERSION` as an (undefined) array, so the regex matcher is simply `@`.

Escape the two `@` characters so we match the literal string `@VERSION@`
in the man page nroff sources.


xsuchy:
Thank you.

PR #198, opened 2018-06-13
add qemu_user_static plugin

wenningerk:
building against foreign architectures seemed not to work for me in a systemd container.
not sure if the approach is desirable but at least it showcases that building
foreign architectures with bootstrap-containers is working if one simply copies
the qemu-binary into to destination-buildroot. 
copyin & bind-mount don't work as the qemu-binary has to be available during
installation as a lot of scriptlets can't be executed properly without.
of course it would be possible to detect automatically which and if the qemu-binary
should be copied. A bindmount done before the packages are installed should of
course work as well I guess.
Excuse my ignorance if I've overseen a mechanism that is already implemented.

Example from my config:

config_opts['plugin_conf']['qemu_user_static_enable'] = True
config_opts['plugin_conf']['qemu_user_static_opts']['binary_path'] = '/usr/bin/qemu-aarch64-static'

### DIFF
--- a/mock/py/mockbuild/plugins/qemu_user_static.py
+++ b/mock/py/mockbuild/plugins/qemu_user_static.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# vim:expandtab:autoindent:tabstop=4:shiftwidth=4:filetype=python:textwidth=0:
+# License: GPL2 or later see COPYING
+# Written by Klaus Wenninger
+# Copyright (C) 2018 Klaus Wenninger <kwenning@redhat.com>
+
+# python library imports
+
+# our imports
+
+import os
+import os.path
+import shutil
+from mockbuild.trace_decorator import traceLog
+import mockbuild.util
+
+requires_api_version = "1.1"
+
+
+# plugin entry point
+@traceLog()
+def init(plugins, conf, buildroot):
+    QemuUserStatic(plugins, conf, buildroot)
+
+
+class QemuUserStatic(object):
+    """copies user-space-qemu into buildroot"""
+    # pylint: disable=too-few-public-methods
+    @traceLog()
+    def __init__(self, plugins, conf, buildroot):
+        self.buildroot = buildroot
+        self.config = buildroot.config
+        self.state = buildroot.state
+        self.qemu_user_static = conf['binary_path']
+        # Skip for boostrap chroot as this is anyway host-arch
+        if buildroot.is_bootstrap or (self.qemu_user_static == None):
+            return
+        plugins.add_hook("preinit", self._copyQemuUserStatic)
+
+    @traceLog()
+    def _copyQemuUserStatic(self):
+        destpath = self.buildroot.make_chroot_path(self.qemu_user_static)
+        directory = os.path.dirname(destpath)
+
+        try:
+            os.stat(directory)
+        except:
+            os.makedirs(directory)
+
+        try:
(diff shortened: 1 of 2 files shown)


xsuchy:
> building against foreign architectures seemed not to work for me in a systemd container.

Can you elaborate more? What error you get? With this config:

```
config_opts['use_bootstrap_container'] = True
config_opts['use_nspawn'] = True

```

the command: ` mock -r fedora-28-ppc64le --forcearch=ppc64le shell` works for me (as well the build of the package. Without the need to copy something into buildroot.

wenningerk:
Have to check if nspawn changes anything.
Currently on mobile-internet (not a good idea for online repos) as dsl is having an outage.
Will come back to you ...

wenningerk:
Without copying qemu I'm still getting the scriptlet-errors like:

```
  Running scriptlet: glibc-2.27-15.fc28.aarch64                                                                     16/177 
  Installing       : glibc-2.27-15.fc28.aarch64                                                                     16/177 
  Running scriptlet: glibc-2.27-15.fc28.aarch64                                                                     16/177 
```

let me check with ppc64le & verify if anything crucial in the config got stuck with the old mock-version ...

wenningerk:
configs checked/updated & tried with ppc64le - still failing scriptlets without qemu copied
maybe some difference in the systemd/nspawn-setup or different versions of systemd/nspawn, qemu, kernel ... 
btw. no selinux-issue either ...

xsuchy:
Can you share the exact error> And/or the steps to reproduce? Which scriptlets of which package fails?

Conan-Kudo:
This feature _will not_ work on RHEL 7. You need a sufficiently new version of `dnf` along with `qemu` packaging that provides `qemu-user-static`.

wenningerk:
Well, as written it does work in principle. Of course I had to build a qemu-user-static package (actually not that big a deal as all the requirements for system-qemu that can hardly be fulfilled on rhel-7 are not there for userspace-qemu). Regarding dnf the one in the bootstrap-chroot should do the trick.
Btw. did you check with Miroslav? We had an irc-session and he was able to reproduce the issue.
And to be honnest - even if it turns out that there is no nicer way to tackle the issue than to copy the binary - so what if it gives us the possibility of using a broader variety of build-hosts.

Conan-Kudo:
If you need a compatible version of DNF to use with mock, you can use [DNF from my COPR repo](https://copr.fedorainfracloud.org/coprs/ngompa/dnf2-el7/). Alternatively, you can use the YUM4 repo provided by CentOS (which replaces a couple of system components).

wenningerk:
Thanks regarding the DNF from your COPR. Looks much more appealing than the CentOS approach.
Regarding --forcearch it seems as if that isn't needed to setup the bootstrap - that is still x86_64. But maybe that is the problem already somehow...
I'll give the DNF from you COPR a try though as then it might work without the bootstrapping.
From how it looks like (qemu being taken just from within the container instead of from the host) I would expect that the main issue is somehow the setup of the container that might be somehow different with rhel 7.5.

xsuchy:
I was thinking a lot about this PR. At the end, I decided to reject it.

This code will not work out of the box for common users. Instead of this PR I would rather see a documentation page, which will state why qemu builds does not work on rhel7, where you can get more recent qemu packages for rhel7 and how to use `copyin` command to get qemu-static inside of chroot and successfully finish on EL7. You can put this document on mock.wiki if you want.

PR #204, opened 2018-06-27
Completion improvements

tmzullinger:
I hit the issue of `--nocheck` causing subsequent completion to be broken again yesterday and decided to take a look at fixing that.  In the process I found a small handful of options which were not being completed and some others which completed directories but should also provide completion for files.

While I was here, I tried to lightly modernize the completion using features from newer bash-completion releases. All supported Fedora/RHEL releases have versions of bash-completion with the necessary functions.

I tried to test and verify that all completions work as well or better after these changes, but more testing is always a good thing.


xsuchy:
LGTM. Thank you.

PR #208, opened 2018-07-06
fallback to C.UTF-8 locale

zdzichu:
Locale en_US.UTF-8 is not guaranteed to be available. It may be in optional, not installed by default langpack, or may need to be generated.
Instead, use C.UTF-8. It is available in glibc for 3 years already, even longer in Debian (courtesy of local patch).

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -766,7 +766,7 @@ def clean_env():
         'HOSTNAME': 'mock',
         'PATH': '/usr/bin:/bin:/usr/sbin:/sbin',
     }
-    env['LANG'] = os.environ.setdefault('LANG', 'en_US.UTF-8')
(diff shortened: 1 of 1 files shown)


zdzichu:
Hm, you seem to be right. I was under wrong impression. The upstream bug is https://sourceware.org/bugzilla/show_bug.cgi?id=17318
Nevertheless, all the major distros have this locale.

keszybz:
It doesn't really matter if glibc upstream merged that, it looks like they are going to do that sooner or later, and Fedora already has that locale, and it's not going away, so we don't need to wait for upstream.

LGTM.

xsuchy:
Hmm RHEL* shows for: `locale -a|grep '^C'` only C, but not C.UTF8. This is listed only on Fedora. But passing it works. It is probably silently fall back to C. Without warning. So LGTM.

PR #211, opened 2018-07-23
Add bash completion for --copyin with files (not just dirs)

sfowl:
Fixed behaviour to match documentation. Useful for copying files.

https://github.com/rpm-software-management/mock/blob/devel/mock/docs/mock.1#L92


sfowl:
Nevermind, I made this compared to 'master', but it's already fixed in 'devel'. Thanks!

PR #212, opened 2018-08-03
Keep the proxy by default

BrunoVernay:
Keep the proxy configuration by default.
It allows to use    config_opts['http_proxy'] = os.getenv("http_proxy")
and a configured proxy could just work

### DIFF
--- a/mock/etc/consolehelper/mock
+++ b/mock/etc/consolehelper/mock
@@ -2,4 +2,4 @@ USER=root
 PROGRAM=/usr/libexec/mock/mock
 SESSION=false
 FALLBACK=false
-KEEP_ENV_VARS=COLUMNS,SSH_AUTH_SOCK
+KEEP_ENV_VARS=COLUMNS,SSH_AUTH_SOCK,http_proxy,https_proxy,no_proxy


mock/etc/mock/site-defaults.cfg
```diff
@@ -378,6 +378,10 @@ config_opts['bootstrap_module_install'] = []
 #
 # Proxy settings (https_proxy, ftp_proxy, and no_proxy can also be set)
 # config_opts['http_proxy'] = 'http://localhost:3128'
+config_opts['http_proxy']  = os.getenv("http_proxy")
+config_opts['https_proxy'] = os.getenv("https_proxy")
+config_opts['no_proxy']    = os.getenv("no_proxy")
```
xsuchy: in site-defaults.cfg everything should be actually commented out. No real code there. Just inform the user about default. So these lines are fine, just put # in front of them. The real code should go to mock/py/mockbuild/util.py to `get_proxy_environment()`

mock/etc/consolehelper/mock
```diff
@@ -2,4 +2,4 @@ USER=root
 PROGRAM=/usr/libexec/mock/mock
 SESSION=false
 FALLBACK=false
-KEEP_ENV_VARS=COLUMNS,SSH_AUTH_SOCK
+KEEP_ENV_VARS=COLUMNS,SSH_AUTH_SOCK,http_proxy,https_proxy,no_proxy
```
xsuchy: maybe ftp_proxy too?

xsuchy:
Otherwise, it looks good.

mock/etc/mock/site-defaults.cfg
BrunoVernay: In `get_proxy_environment`, not in `setup_default_config_opts` ???

BrunoVernay:
Should be OK now.

PR #214, opened 2018-08-06
expose print_main_output in configs

squarebracket:
When building the kernel with mock on gitlab CI, I found that there was no way of having it print the output. Digging into the code, I found that `print_main_output` was the main thing controlling whether or not I saw the logs.

This PR simply exposes that configuration value in the actual configs. The default behaviour is preserved; it will detect whether stderr is a tty, and set it True or False based on that. However, it can now be overridden by the configuration.

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1043,7 +1043,8 @@ def set_config_opts_per_cmdline(config_opts, options, args):
     config_opts.update(cli_opt_new)
 
     config_opts['verbose'] = options.verbose
-    config_opts['print_main_output'] = config_opts['verbose'] > 0 and sys.stderr.isatty()
+    if 'print_main_output' not in config_opts or config_opts['print_main_output'] is None:
+        config_opts['print_main_output'] = config_opts['verbose'] > 0 and sys.stderr.isatty()
 
     # do some other options and stuff
     if options.arch:
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -35,6 +35,11 @@
 # config_opts['root_log_fmt_name']  = "detailed"
 # config_opts['state_log_fmt_name'] = "state"
 #
+# By default, mock only prints the build log to stderr if it is a tty; you can
+# force it on here (for CI builds where there is no tty, for example) by
+# setting this to True, or force it off by setting it to False.
+# config_opts['print_main_output'] = None
+#
 # mock will normally set up a minimal chroot /dev.
 # If you want to use a pre-configured /dev, disable this and use the bind-mount
 # plugin to mount your special /dev
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -35,9 +35,10 @@
 # config_opts['root_log_fmt_name']  = "detailed"
 # config_opts['state_log_fmt_name'] = "state"
 #
-# By default, mock only prints the build log to stderr if it is a tty; you can
+# By default, mock only prints the build log to stderr if it is a tty. You can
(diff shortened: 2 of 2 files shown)


xsuchy:
This was done intentionally. When you have tty (you are human) you want to have the output on tty. When you are building without tty you are a script (likely CI) and you should grab a log file.

praiskup:
After rhbz#1166609, and after neverending fight with the isatty() hack in mock (in copr-builder/copr-rpmbuild/covscan) I very much understand why this PR has been submitted.  IMO the magic isatty() should be even opt-in only.

praiskup:
> When you are building without tty you are a script (likely CI) and you should grab a log file.

Mostly the problem is even with scripts you might want to see the combined log;  e.g. when you 
try to send live logs somewhere...  That's koji, copr, covscan (kobo), ...

squarebracket:
Even as a human, I will sometimes kick off a build manually to see what happens when loading certain patches programmatically. In those cases, I am generally interested in watching the output, and it's nice that I don't have to spin up a dev environment every time I do that.

I suppose I can see if gitlab CI can handle multiple `artifact` entries. As it stands I'm collecting both the log files and RPMs, but I let those artifacts expire (i.e. get deleted) after a certain amount of time so that they don't take up tons of space. Keeping those logs around indefinitely would be nice in case I need to review a build log much later.

In any case, I guess my argument here is that doing the tty-check is a reasonable _default_, but I find it strange that it's _enforced_ when it could simply be exposed like other configuration values.

xsuchy:
> In any case, I guess my argument here is that doing the tty-check is a reasonable default, but I find it strange that it's enforced when it could simply be exposed like other configuration values.

Ok. I bend in front of this argument :)

mock/py/mockbuild/util.py
```diff
@@ -1043,7 +1043,8 @@ def set_config_opts_per_cmdline(config_opts, options, args):
     config_opts.update(cli_opt_new)
 
     config_opts['verbose'] = options.verbose
-    config_opts['print_main_output'] = config_opts['verbose'] > 0 and sys.stderr.isatty()
+    if 'print_main_output' not in config_opts or config_opts['print_main_output'] is None:
```
xsuchy: This will not work if someone set `print_main_output` to False as suggested in `site-defaults.cfg`. Can you rather make it `... or not config_opts['print_main_output']`? And the default in `site-defaults.cfg` set `False`? Because the line below set it to a boolean value.

squarebracket: What specifically would not work if `print_main_output` were set to False? It should be that `True` will make it always print, `False` will make it never print, and `None` (or not set) will make it do the tty check.

If you feel that my explanation was insufficient in the `site-defaults.cfg`, I can explicitly say that a value of `None` or leaving it unset will use the tty check.

xsuchy: Ahh, you meant it this way... Yes explanation in `site-defaults.cfg` that "None will use tty check, False means disabled, True means enabled" would be helpful.

squarebracket: Added language in site-defaults.cfg that explicitly says `None` or leaving it defined will use the default (tty-check) behavior.

xsuchy:
Awesome. Thank you.

squarebracket:
Thank you for the merge!

Is there any sort of time frame for when I could expect an official release that includes this patch? I can use the nightly copr for now, but I would like to switch back to the official EPEL builds eventually :)

xsuchy:
https://github.com/rpm-software-management/mock/wiki/FAQ#when-i-can-expect-next-release
(in short next week)

squarebracket:
Woops! Didn't see it in the README but didn't think of checking the wiki. Thanks again.

PR #215, opened 2018-08-09
Don't try to use a spec we've already cleaned up

owtaylor:
Since commit 3a9221c5d13, a rebuild with scm in the config options
would a) check out from SCM b) generate an SRPM from spec and sources
c) clean up the SCM checkout including the spec file
d) try to copy the spec into the chroot


xsuchy:

PR #221, opened 2018-09-13
[mockbuild] create custom error message for dnf-utils not being installed

pjunak:
Created a new exception for absent weak dependent package.
Added custom message for error caused by dnf-utils not being installed.

bugzilla url:
https://bugzilla.redhat.com/show_bug.cgi?id=1626879

I am an intern working under Mirek Suchý.

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -5,6 +5,7 @@ import copy
 import glob
 import os.path
 import shutil
+import sys
 from textwrap import dedent
 
 import distro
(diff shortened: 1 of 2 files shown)


praiskup:
I've been kicked by this yesterday :-) +1 for solving this problem.

mock/py/mockbuild/package_manager.py
```diff
@@ -156,7 +157,22 @@ def update(self, *args, **kwargs):
...
+        except (FileNotFoundError) as error:
+            er = str(error)
+            if "builddep" in er:
+                print(error)
+                print("""
+Error:      dnf-utils is not installed. Dnf-utils is needed to complete this action.
+            To install dnf-utils use one of the commands below:
+            $ dnf install dnf-utils
+            or
+            $ yum install dnf-utils""")
```
praiskup: Shouldn't here be `yum install yum-utils`?

pjunak: I have to admit I don't know for sure. Both ```yum install dnf-utils``` and ```yum install yum-utils``` seems to work. But I guess it should be the later.
I'll create a fix.

praiskup: It IMO depends how mock behaves on rhel-7;  if `/usr/bin/yum-builddep` is used or not.  I don't think it is OK to rely on `dnf-utils` in rhel-7.

praiskup: That said, I'm not 100% sure whether the `or yum install ...` is even needed :-)

PR #222, opened 2018-09-25
Fix for sclo-rh

sergiomb2:
repo sclo includepkgs=devtoolset* and repo sclo-rh should includepkgs=rh*


sergiomb2:
ok this is not totally correct , closing

PR #226, opened 2018-11-07
Add source repos to fedora-28-x86_64.cfg

sfowl:
This is just a small example to test if it would be acceptable to include source repos in mock configs.

This would allow for commands like, e.g.
```
$ mock -r fedora-28-x86_64 --dnf-cmd download -- --source vim
```
Note that the source repos have `enabled=0` so would not be checked by default.

If this is acceptable, I would be happy to make another PR that added source repos for all currently shipped fedora configs.


xsuchy:
I struggle to find the use case here.  Why would you download sources to chroot?

sfowl:
> I struggle to find the use case here. Why would you download sources to chroot?

To be clear, a package downloaded with `mock --dnf-cmd download` appears in the current working directory (i.e. not the chroot). I often need to download and repoquery srpms for inspection/testing.

Currently, I do this by manually updating the mock configs. `mock` has proven a very useful tool for switching between source repositories on a whim.

xsuchy:
OK than. Feel free to add it to all config and I will merge it.

sfowl:
Thanks, I've added source repos to all the fedora configs. 

It wasn't clear to me why some repos use `*-primary` gpgkeys and some used `*-secondary` gpgkeys, I tried to copy what was in the existing configs but this was done by hand so it's possible I made some mistakes.

xsuchy:
Primary GPG key is used for primary arches and secondary for secondary arches. See https://fedoraproject.org/wiki/Architectures

xsuchy:
Thank you.

PR #229, opened 2018-11-15
Require dnf-utils

frantisekz:
Fedora 29 does not ship with by yum default, so require dnf-utils to allow non-bootstrapped epel-7 builds on Fedora without needing to manually install yum/dnf-utils.

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -74,7 +74,7 @@ Requires: dnf
 Suggests: yum
 Requires: dnf-plugins-core
 Recommends: btrfs-progs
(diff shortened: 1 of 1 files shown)


xsuchy:
I am against this. This should not be Requires. This is only needed for building epel7 on Fedora. Which is a minority for most users.
I am open to discussion if this should be flipped to Recommends, but definitely not to Requires.

frantisekz:
Yeah, Recommends suits this use case far more than Requires. It Just didn't come to my mind. Thanks!

I've amended the original commit.

PR #238, opened 2019-02-21
Fix --enable-network documentation in man page

directhex:
The config file option is rpmbuild_networking not rpmbuild_network


xsuchy:
Nice catch. Thank you.

PR #239, opened 2019-02-26
Forcing armv7hl arch when it is the target arch enables transparent n…

Pix13:
Forcing armv7hl arch when it is the target arch enables transparent native build on aarch64 hosts with EL0 CONFIG_COMPAT support

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1124,6 +1124,11 @@ def set_config_opts_per_cmdline(config_opts, options, args):
     if options.forcearch:
         config_opts['forcearch'] = options.forcearch
 
+    # if armv7hl is forced when being the target it enables it to build from aarch64 hosts
+    if config_opts['target_arch'] == 'armv7hl':
+        config_opts['forcearch'] = 'armv7hl'
+        options.arch = 'armv7hl'
+
     if not options.clean:
         config_opts['clean'] = options.clean
 


xsuchy:
Well, with this patch it will allow you to build armv7hl on Intel platforms as well. But it has some drawbacks. This option (see --forcearch in the man page) has to be supported by the package manager (DNF yes, YUM no) and will use/require QEMU static emulation.
Additionally, there are known reports of builds which under QEMU static emulation false. Not many of them, but there are some.
For these reasons, I would like to keep it as a separate option. You can always pass `--forcearch=arm7hl` to mock, but you should know what you are doing and be aware of potential drawbacks.

Pix13:
Hmm keeping this MR closed, what would you say about a patch that checks first if we are running aarch64 ( easy) and that tests if the aarch64 host supports aarch32 ( tricky  - aka the armv7hl compatible armv8l 32 bit instruction set), would it be acceptable then to have an "automatic" forcearch ?

xsuchy:
I would prefer the following feature:
```
if target architecture is incompatible with current_archiecture AND /usr/bin/qemu-$TARGET_ARCH-static is present then
    print HUGE WARNING the we are going to build using software emulation
    sleep 2
    set force_arch
```
Additionaly, I must say this is first time I hear about aarch32, so I will have to study something about it. But if it is like i368 vs. x86_64 then 

`config_opts['legal_host_arches'] = ('armv7l', 'aarch64')`

in 

fedora-*-armhfp.cfg configs should be enough. But I lived under the impression that 64bits ARM is not able to run 32bits binaries.

kwizart:
FYI, we eventually plan to use this feature in koji (for RPM Fusion).
So using the same interface for the koji builder mock invocation than any others build targets would be nice.
(that's the main reason of the patch).

Aarch64 vs armhfp is a little different than x86_64 vs i386. But basically:
Most current arm64 HW implement a state machine allowing to run armv7hl binaries (no emulation).

In the Fedora case, there is no multilibs support for aarch64/armhfp as designed, because arm32 support will not always be available in HW. But it should be possible to run an armhfp container from an aarch64 host if the HW supports it (actually only current fedora rawhide kernel 5.0 has COMPAT enabled).

In the next months, there it is expected to see more arm64 HW missing some arm32 support. In this case it will not be possible to run an armhfp kernel as guest, but only armhfp userspace binaries (from aarch64 host or guest kernel).

Pix13:
> Additionaly, I must say this is first time I hear about aarch32, so I will have to study something about it. But if it is like i368 vs. x86_64 then
> 
> `config_opts['legal_host_arches'] = ('armv7l', 'aarch64')`
> 
> in
> 
> fedora-*-armhfp.cfg configs should be enough. But I lived under the impression that 64bits ARM is not able to run 32bits binaries.

It's a bit different as x86_4/i386 on several aspect ( on intel, you use -m32 to build 32 bit, on arm you have two separate gcc toolchains), and as kwizart mentionned, fedora is not multilib (32/64) so it's a different rootfs installed for armv7 and aarch64. Basically, when an aarch64 cpu supports aarch32, the instruction set is armv8l which is designed to be nearly identical as armv7l, enabling "native" builds without the need of Qemu, hence what i'm trying to do. Detecting if an aarch64 host is able to execute aarch32 is not trivial ( see https://community.arm.com/android-community/b/android/posts/runtime-detection-of-cpu-features-on-an-armv8-a-cpu ) but could be done

PR #240, opened 2019-03-01
Add debug logging for systemd-nspawn and related args

jcpunk:
I found an odd behavior when running nspawn.  Being able to see the exact nspawn arguments was helpful in diagnosing my issue (missing capability).  I figured this would be helpful to others too.


xsuchy:
Thank you.

PR #243, opened 2019-04-01
Run the `module_enable` and `module_install` after the `chroot_setup_cmd`.

hanzz:
Modules cannot be enabled before the chroot is setup.

Signed-off-by: Jan Kaluza <jkaluza@redhat.com>

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -244,12 +244,6 @@ class Buildroot(object):
     def _init_pkg_management(self):
         update_state = '{0} install'.format(self.pkg_manager.name)
         self.state.start(update_state)
-        if 'module_enable' in self.config and self.config['module_enable']:
-            cmd = ['module', 'enable'] + self.config['module_enable']
-            self.pkg_manager.init_install_output += self.pkg_manager.execute(*cmd, returnOutput=1)
-        if 'module_install' in self.config and self.config['module_install']:
-            cmd = ['module', 'install'] + self.config['module_install']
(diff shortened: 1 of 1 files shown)


xsuchy:
Hmm really? What if I want to install something from the module in chroot_setup_cmd?

What about this order:

1) dnf module enable
2) chroot_setup_cmd
3) dnf module install

?

BTW do you have some real world use case, where this caused an issue?

hanzz:
@xsuchy, `dnf module enable` needs `PLATFORM_ID` in `/etc/os-release`. This is added by `fedora-release` RPM, which is installed as part of `chroot_setup_cmd`. So, without installing `fedora-release` before `module_enable`, it is not possible to install any module right now.

So far there is no need to install any basic package usually available in `chroot_set_cmd` from module, because all of them are available as part of non-modular repo and there is no plan to change that. On the otherside, it should be possible to install current modules in Mock, which is not working.

I would vote for fixing the ordering the way I proposed, because it is backward-compatible change. If we want to make the order more general in the future, it probably need bigger change.

xsuchy:
@hanzz This is not an issue, because 
https://github.com/rpm-software-management/mock/blob/devel/mock-core-configs/etc/mock/fedora-30-x86_64.cfg#L27

It is more about the question - will ever be some package from `@buildsys` group in a module?

hanzz:
OK, I didn't know about module_platform_id=platform:f30. I will try setting that myself with unpatched Mock and see if it works.

hanzz:
This seems to work, thanks for help :). I'm closing this PR.

PR #253, opened 2019-04-28
Fix openSUSE Tumbleweed DistTag definition

Conan-Kudo:
The DistTag macro definition for openSUSE Tumbleweed was slightly
incorrect, and did not correctly retrieve the `VERSION_ID` from the
`os-release(5)` file due to how shell expansion works.

This should be fixed now.


xsuchy:
Thank you.

PR #257, opened 2019-05-05
Allow AArch64 systems to build 32-bit ARM packages

Conan-Kudo:
There are a number of AArch64 systems that natively support
building and running code for 32-bit ARM ISAs.

In particular, some of the new ARM cloud servers are capable of it,
like the new ARM cloud servers from Packet.com.

RPM already has enhancements for supporting this as well.

So, let's go ahead and allow it.


Conan-Kudo:
cc: @pterjan

xsuchy:
ack, in worst case, it will fail few cycles later

PR #261, opened 2019-05-16
Custom chroot config

xsuchy:
This is WIP.

For some reason, the bash completion for ~/.config/mock does not work.
@praiskup can you please try to review it?

### DIFF
--- a/mock/etc/bash_completion.d/mock
+++ b/mock/etc/bash_completion.d/mock
@@ -2,7 +2,12 @@
 
 _mock_root()
 {
-    COMPREPLY+=( $( compgen -W "$( command ls ${1:-/etc/mock} 2>/dev/null | \
+    if [ -z "$1" ]; then
+        CFGDIRS=(/etc/mock, "$HOME/.config/mock/")
+    else
+        CFGDIRS=($1)
+    fi
+    COMPREPLY+=( $( compgen -W "$( command ls \"${CFGDIRS[@]}\" 2>/dev/null | \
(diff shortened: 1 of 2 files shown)


mock/etc/bash_completion.d/mock
```diff
@@ -2,7 +2,12 @@
 
 _mock_root()
 {
-    COMPREPLY+=( $( compgen -W "$( command ls ${1:-/etc/mock} 2>/dev/null | \
+    if [ -z "$1" ]; then
+        CFGDIRS=(/etc/mock, "$HOME/.config/mock/")
```
praiskup: the comma is part of the first item in array, needs to be removed

mock/etc/bash_completion.d/mock
```diff
@@ -2,7 +2,12 @@
...
+    else
+        CFGDIRS=($1)
+    fi
+    COMPREPLY+=( $( compgen -W "$( command ls \"${CFGDIRS[@]}\" 2>/dev/null | \
```
praiskup: I'm not sure `ls` is good option here.  If you list one directory, only the files are on the output;  but if you have two directories, you have the format:
```
<DIR>:
file1 file2 file3

<DIR2>:
....
```

praiskup: maybe `$( for dir in "${CFGDIRS[@]}"; do ls "$dir"; done )`;  it would remove the `command` builtin, but I'm not sure it is needed ..

xsuchy:
Merged as a776c3ce382a1930bf6731bf8d268e61256be765

Issue #255, opened 2019-05-02
RFE: provide a user specific configuration directory


hroncok:
Hi, apparently, mock already reads the config file from `~/.config/mock.cfg` - that is great, I don't need to mess up the configuration in `/etc/mock`.

This is a request for enhancement to allow storing chroot configuration files in `$HOME` as well.

I now have `~/.config/mock/fedora-rawhide-x86_64-python3.8.cfg`, however I always need to use full path when running `mock -r ...` to use this file. Mock only looks for the files in `/etc/mock/<CONFIG>.cfg`, I'd suggest to support `~/.config/mock/<CONFIG>.cfg` as well (probably with higher priority).

Basically I want to change:

       -r CONFIG, --root=CONFIG
              Uses  specified  chroot  configuration  as  defined  in /etc/mock/<CONFIG>.cfg. Optionally if CONFIG ends in '.cfg', it is interpreted as full path to config file. If none specified, uses the chroot config linked to by /etc/mock/default.cfg.

To:

       -r CONFIG, --root=CONFIG
              Uses  specified  chroot  configuration  as  defined  in ~/.config/mock/<CONFIG>.cfg or /etc/mock/<CONFIG>.cfg. Optionally if CONFIG ends in '.cfg', it is interpreted as full path to config file. If none specified, uses the chroot config linked to by /etc/mock/default.cfg.

hroncok:
JFYI: That commit message says `~/.mock/FOO.cfg` yet the code has `~/.config/mock/FOO.cfg`

xsuchy:
yes, commit message is wrong, I will make sure that in release notes it will be correct.

PR #262, opened 2019-05-17
Update the man page about ~/.config/mock/FOO.cfg

hroncok:
See https://github.com/rpm-software-management/mock/issues/255


xsuchy:
Thank you

PR #265, opened 2019-05-27
grant anyone access to bind-mounted /etc/resolv.conf

praiskup:
The mkstemp() call creates file with 0600 permissions, and it is
not a natural setup for the /etc/resolv.conf file.

The 0600 permissions are especially problematic with
'systemd-nspawn --private-users=pick' (recommended settings by
systemd docs, and used e.g. by copr in some cases).  The option
doesn't imply that the file is re-mapped according to the picked
user namespace.

Fixes: #265

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1424,6 +1424,9 @@ def setup_host_resolv(config_opts):
     resolv_path = (tempfile.mkstemp(prefix="mock-resolv."))[1]
     atexit.register(_nspawnTempResolvAtExit, resolv_path)
 
+    # make sure that anyone in container can read resolv.conf file
+    os.chmod(resolv_path, 0o644)
+
     if config_opts['use_host_resolv']:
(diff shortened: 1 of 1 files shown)


praiskup:
@msekletar can you please confirm that this is not a bug in systemd-nspawn; i.e that bind-mounted files are intentionally not chowned according to ` --private-users=pick` decisions (even though the root (recursively) is)?

I mean, I doubt mock can somehow predict in advance the mapping which systemd-nspawn will use to set the ownership accordingly (to mimic real 'root:root' ownership of /etc/resolv.conf).

msekletar:
> can you please confirm that this is not a bug in systemd-nspawn; i.e that bind-mounted files are intentionally not chowned according to ` --private-users=pick` decisions (even though the root (recursively) is)?

I didn't check the code but I think the behaviour you describe actually makes sense. Changing ownership of on the bind-mounted file would change the permissions on the host and you probably don't want that.

PR #268, opened 2019-05-29
bootstrap: enforce elfutils-libelf installation in epel-*

praiskup:
There are several devtoolset-*-elfutils-libelf packages which
provide 'libelf.so.1()(64bit)' soname (bug), and since rpm-libs
depend on that soname those could be installed instead.

This actually happens if the bootstrap --installroot is done with
dnf (dnf-yum) package, where /usr/bin/yum calculates the
transaction a little bit differently than /usr/bin/yum-deprecated.

It's (probably?) not OK to just '--disablerepo sclo*' for
bootstrap --installroot, since those repositories are now by
default allowed even in Koji [1] and some package in that
transaction might (at least in theory) depend on some DTS package.

https://lists.fedorahosted.org/archives/list/\
epel-devel@lists.fedoraproject.org/thread/6EJIQ72JXWAAOQPFLRFYE5A7CUKC756D/

### DIFF
--- a/mock-core-configs/etc/mock/epel-6-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-6-x86_64.cfg
@@ -6,6 +6,7 @@ config_opts['dist'] = 'el6'  # only useful for --resultdir variable subst
 # beware RHEL uses 6Server or 6Client
 config_opts['releasever'] = '6'
 config_opts['use_nspawn'] = False
+config_opts['yum_install_command'] += ' --disablerepo sclo*'
 
 config_opts['yum.conf'] = """
 [main]
--- a/mock-core-configs/etc/mock/epel-7-aarch64.cfg
+++ b/mock-core-configs/etc/mock/epel-7-aarch64.cfg
@@ -4,6 +4,7 @@ config_opts['legal_host_arches'] = ('aarch64',)
 config_opts['chroot_setup_cmd'] = 'install @buildsys-build'
 config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '7'
+config_opts['yum_install_command'] += ' --disablerepo sclo*'
 
 config_opts['yum.conf'] = """
 [main]
--- a/mock-core-configs/etc/mock/epel-7-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/epel-7-ppc64le.cfg
@@ -4,6 +4,7 @@ config_opts['legal_host_arches'] = ('ppc64le',)
 config_opts['chroot_setup_cmd'] = 'install @buildsys-build'
 config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '7'
+config_opts['yum_install_command'] += ' --disablerepo sclo*'
 
 config_opts['yum.conf'] = """
 [main]
--- a/mock-core-configs/etc/mock/epel-7-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-7-x86_64.cfg
@@ -4,6 +4,7 @@ config_opts['legal_host_arches'] = ('x86_64',)
 config_opts['chroot_setup_cmd'] = 'install @buildsys-build'
 config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '7'
+config_opts['yum_install_command'] += ' --disablerepo sclo*'
 
 config_opts['yum.conf'] = """
 [main]


praiskup:
@hhorak, fyi

praiskup:
Ok, thinking again -- the installroot probably shouldn't really depend on
sclo* repos.  That is aimed to help only for build requires, per the cited
thread:

>  CAVEAT: EPEL packages may only use devtoolset as a build requirement
>  and not as a Requires. This is because it needs access to extra
>  repositories and channels a user may not wish or be allowed to add to
>  their system.

--installroot only calculates Requires.  I'll change the patch.

mock/py/mockbuild/util.py
```diff
@@ -1100,7 +1100,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     # configurable commands executables
     config_opts['yum_command'] = '/usr/bin/yum'
     config_opts['system_yum_command'] = '/usr/bin/yum'
-    config_opts['yum_install_command'] = 'install yum yum-utils shadow-utils distribution-gpg-keys'
+    config_opts['yum_install_command'] = 'install --disablerepo sclo* yum yum-utils shadow-utils distribution-gpg-keys'
```
Conan-Kudo: Could you please put this in the `epel-*` chroot configs instead? You're essentially saying `sclo*` repos are disabled for _all_ targets, and there's more than CentOS and Fedora using mock.

Conan-Kudo: Meaning, you don't know what you're disabling by doing this.

praiskup: I may.  Do you have any reasons to keep the sclo* repos enabled anywhere else?

Conan-Kudo: Do you know that `sclo*` is going to match only for the Software Collections repos? People can make their own mock configs where those IDs can be anything...

praiskup: Ok, it's rather really theoretical worry which brings slightly more maintenance work (we'll have to think about this when epel-8 very likely, downstream rhel, etc...).   But I don't mind in particular, and I'll move it to epel configs.

Conan-Kudo: Thank you.

praiskup:
Honza reported this against CentOS tracker: https://bugs.centos.org/view.php?id=16125

sergiomb2:
hum "devtoolset-X-elfutils-libs provide library that is the same as in the system" and it was fixed [1].
I'd like also remove includepkgs=devtoolset* in [2] and [3] , as I think as koji don't have this kind of disablerepo and include= ... I have to check this better 

[2]
https://github.com/rpm-software-management/mock/blob/devel/mock-core-configs/etc/mock/epel-7-x86_64.cfg#L64
[3]
https://github.com/rpm-software-management/mock/blob/devel/mock-core-configs/etc/mock/epel-7-x86_64.cfg#L72




[1] 
https://bugzilla.redhat.com/show_bug.cgi?id=1412305
https://access.redhat.com/errata/RHBA-2017:1149

praiskup:
> hum "devtoolset-X-elfutils-libs provide library that is the same as in the system" and it was fixed

The erratum you cited is from 2017, the problem I faced is from 2019 - and it is unlikely to be fixed ATM.  Unless you claim CentOS release engineers dropped the broken packages from the repositories.

> as I think as koji don't have this kind of disablerepo and include=

I don't think Koji uses `mock-core-cofigs` package.

praiskup:
This problem is _not_ fixed at the time of writing this comment.

Using `podman --rm -ti centos:6 bash`:
```
[root@45e1ef5253c1 /]# yum repolist
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
...
repo id             repo name                                         status
base                CentOS-6 - Base                                    6713
centos-sclo-rh      CentOS-6 - SCLo rh                                 3220
centos-sclo-sclo    CentOS-6 - SCLo sclo                                432
epel                Extra Packages for Enterprise Linux 6 - x86_64    12570
extras              CentOS-6 - Extras                                    46
updates             CentOS-6 - Updates                                  609
repolist: 23590
[root@45e1ef5253c1 /]# repoquery --whatprovides 'libelf.so.1()(64bit)'
devtoolset-6-elfutils-libelf-0:0.167-2.el6.x86_64
elfutils-libelf-0:0.164-2.el6.x86_64
```

Using `podman --rm -ti centos:7 bash`:
```
[root@cceaa03de671 /]# repoquery --whatprovides 'libelf.so.1()(64bit)'
devtoolset-4-elfutils-libelf-0:0.166-1.el7.x86_64
elfutils-libelf-0:0.176-2.el7.x86_64
devtoolset-3-elfutils-libelf-0:0.161-1.el7.x86_64
```

Please install `yum-utils` `epel-release`, `centos-release-scl-rh` and
`centos-release-scl` packages first.

PR #269, opened 2019-05-29
bootstrap: independent network configuration

praiskup:
First, make sure that 'util.do' uses the pre-configured
Buildroot's "nspawn_args" config (when using nspawn is enabled).

Second, postpone setup_host_resolv() call after bootstrap chroot
initialization step, so the bootstrap chroot doesn't inherit
resolv.conf from the main chroot.  This is needed because the main
chroot might or might not have network disabled, while bootstrap
chroot has network always enabled.  IOW, even though both chroots
are using nspawn, the former resolv.conf might be intentionally
empty while the latter one should always be a copy of hosts's
/etc/resolv.conf.

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -594,8 +594,6 @@ def main():
     # cmdline options override config options
     util.set_config_opts_per_cmdline(config_opts, options, args)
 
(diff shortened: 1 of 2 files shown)


xsuchy:
Merci bien.

PR #271, opened 2019-06-06
Add 'fastestmirror=1' to Mageia mock configs

Conan-Kudo:
Mageia still uses yum-style mirrorlists, and the mirrorlists are
returned in random order in each request. To make the best use of
this, we'll use fastestmirror so that somewhat fast mirrors are
selected more often.


praiskup:
Don't you have more info about this?  Is this really needed only for mageia?  How am I able to detect that it would be useful for other chroots?

Conan-Kudo:
@praiskup Mageia doesn't have a nice mirror management system deployed yet, so `fastestmirror` improves the mirror selection for users. Fedora uses a metalink powered by MirrorManager2, so it doesn't need it. EPEL uses the same system. CentOS uses a mirrorlist still, but I'm not sure if the mirrorlist file is generated in some kind of smart fashion or not. You'd have to check with them.

PR #272, opened 2019-06-12
Enable dynamic BuildRequires for Fedora Rawhide

ignatenkobrain:
RPM 4.15 is there, so let's turn it on!

Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -90,7 +90,7 @@
 # dnf-plugins-core installed
 
 # Dynamic BuildRequires, available since RPM 4.15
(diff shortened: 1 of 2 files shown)


xsuchy:
I am tempting to enable it wide in site-defaults.cfg. It should make not harm if rpm actually does not support it. It will be skipped if there is no 'rpmlib(DynamicBuildRequires)' or no such exit code from rpmbuidl (which will not happen if rpm does not support it).

ignatenkobrain:
True, I'll change a patch.

xsuchy:
Thank you.

PR #273, opened 2019-06-14
Fix compatibility with pre-4.15 RPM versions with DynamicBuildRequires

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -283,7 +283,8 @@ class Commands(object):
 
             # Check if we will have dynamic BuildRequires, but do not allow it
             hdr = next(util.yieldSrpmHeaders((rebuilt_srpm,)))
-            dynamic_buildreqs = 'rpmlib(DynamicBuildRequires)' in hdr[rpm.RPMTAG_REQUIRES]
+            requires = {util._to_text(req) for req in hdr[rpm.RPMTAG_REQUIRES]}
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
@xsuchy can you release last few commits in a new version or backport in Fedora please?

mock/py/mockbuild/backend.py
```diff
@@ -283,7 +283,8 @@ def build(self, srpm, timeout, check=True, spec=None):
 
             # Check if we will have dynamic BuildRequires, but do not allow it
             hdr = next(util.yieldSrpmHeaders((rebuilt_srpm,)))
-            dynamic_buildreqs = 'rpmlib(DynamicBuildRequires)' in hdr[rpm.RPMTAG_REQUIRES]
+            requires = {util._to_text(req) for req in hdr[rpm.RPMTAG_REQUIRES]}
```
xsuchy: Why { (dictionary) ? Shouldn't it be list or tuple?

ignatenkobrain: It is a set, not a dictionary

ignatenkobrain: And checking if element is in set is faster than in list...

xsuchy: ah, ok - this was actually new for me.

ignatenkobrain:
@xsuchy :) Could you please release new mock so that we can get it in Fedora infra instead of backporting separate patches?

PR #284, opened 2019-07-02
Allow more loop devices

schlupov:
Run as mock -r fedora-30-x86_64 shell --old-chroot and I can see the loop0-loop11 in /dev.
Doesn't work without --old-chroot

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1098,6 +1098,8 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['dynamic_buildrequires'] = True
     config_opts['dynamic_buildrequires_max_loops'] = 10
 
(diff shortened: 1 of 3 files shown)


praiskup:
It is better to cite the fixed issue in commit message.

mock/py/mockbuild/util.py
```diff
@@ -1098,6 +1098,8 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['dynamic_buildrequires'] = True
     config_opts['dynamic_buildrequires_max_loops'] = 10
 
+    config_opts['loop_max_number'] = 12
```
praiskup: Strictly speaking, "max_number" seems like loop12 will be created .. but the max is loop11.  `dev_loop_count`?

xsuchy:
Thank you.

Issue #266, opened 2019-05-28
Remove "BUILDSTDERR" from log


voxik:
I understand the motivation for #8, but it has two undesirable side-effects:

1. The log cannot be reasonably used for upstream without being questioned "what does the BUILDSTDERR means, I can't see it in my output."
2. It bloats the logs. It is 13 additional bytes for every stderr output.

voxik:
IOW, it is nice we can do this, but it should be off by default and enabled only optionally if needed.

xsuchy:
I can make it optionally, or allow to define your own string (even empty one). That is easy. 

I have problem with the default thou. Can you ask people on fedora-devel mailing list? If I will get more votes, I will flip the default, but I do not want to flip the default there and back every year.

hroncok:
I use mock heavily and when I post snippets upstream, I usually remove the BUILDSTDERR part and it is tedious. Not having it by default would work for me better but it is not a strong opinion. I.e. if differentiating between stdout and stderr is important for others, I don't mind keeping the default as is.

junaruga:
I understand BUILDSTDERR to distinguish stderr from stdout.
But I have never faced at the situation that the feature solved a issue. (Just only in my experience).

I would prefer "default: empty" (no prefix "BUILDSTDERR: ") . It's convenient to report the log to upstream project.

jaredsmith:
I prefer to have some sort of STDERR string, but as long as it's configurable, I'm happy.

Conan-Kudo:
I'd prefer to _not_ have a BUILDSTDERR string by default, as it makes it more difficult for upstream to understand logs when I send them.

jamesjer:
I also prefer to _not_ have BUILDSTDERR printed by default, for the reasons others have given above.

phmccarty:
I would prefer to see the STDERR string configurable and set to the empty string by default.

ignatenkobrain:
I do not care whether it is configurable or not, but I would prefer to not see it.

PR #287, opened 2019-07-03
Removing buildstderr from log

schlupov:
Fixes: #266

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -48,6 +48,10 @@
 # None or leaving it undefined uses the default behavior.
 # config_opts['print_main_output'] = None
 #
+# This option sets the beginning of the line with stderr output in build.log file.
+# By default it's an empty string.
+# config_opts['stderr_line_prefix'] = ""
+#
 # mock will normally set up a minimal chroot /dev.
 # If you want to use a pre-configured /dev, disable this and use the bind-mount
 # plugin to mount your special /dev
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -453,6 +453,10 @@ def setup_logging(config_path, config_opts, options):
     if options.trace:
         logging.getLogger("trace").propagate = 1
 
+    logging.getLogger("mockbuild").stderr_line_prefix = ""
+    if config_opts['stderr_line_prefix'] != "":
+        logging.getLogger("mockbuild").stderr_line_prefix = config_opts['stderr_line_prefix']
+
 
 @traceLog()
 def setup_uid_manager(mockgid):
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -453,9 +453,9 @@ def setup_logging(config_path, config_opts, options):
     if options.trace:
         logging.getLogger("trace").propagate = 1
 
-    logging.getLogger("mockbuild").stderr_line_prefix = ""
+    logging.getLogger("mockbuild")._mock_stderr_line_prefix = ""
     if config_opts['stderr_line_prefix'] != "":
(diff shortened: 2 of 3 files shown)


mock/etc/mock/site-defaults.cfg
```diff
@@ -48,6 +48,10 @@
 # None or leaving it undefined uses the default behavior.
 # config_opts['print_main_output'] = None
 #
+# This option sets the beginning of the line with stderr output in build.log file.
+# By default it's an empty string.
+# config_opts['stderr_line_beginning'] = ""
```
FrostyX: I would be slightly more inclined to call it `prefix` instead of `beginning`.

mock/py/mockbuild/util.py
```diff
@@ -582,7 +582,7 @@ def logOutput(fdout, fderr, logger, returnOutput=1, start=0, timeout=0, printOut
                         if line != '':
                             line = ansi_escape.sub('', line)
                             if fderr is s and not line.startswith('+ '):
-                                logger.debug("BUILDSTDERR: %s", line)
+                                logger.debug("{0} {1}".format(logging.getLogger('mockbuild').stderr_line_beginning, line))
```
FrostyX: As you mentioned in face to face discussion, there is going to be a space at the beginning of the line if no `stderr_line_beginning` is specified. What about having `{0}{1}` without space and letting users have a space on the end of their prefix.

mock/etc/mock/site-defaults.cfg
xsuchy: +1 to `prefix`

mock/py/mockbuild/util.py
xsuchy: Please do not use format() as it is called always, even when log is not actually used. See https://stackoverflow.com/questions/4148790/lazy-logger-message-string-evaluation

schlupov:
I changed what was discussed here. Please take a look.

FrostyX:
+1 looks good to me. Thank you!

mock/py/mock.py
```diff
@@ -453,6 +453,10 @@ def setup_logging(config_path, config_opts, options):
     if options.trace:
         logging.getLogger("trace").propagate = 1
 
+    logging.getLogger("mockbuild").stderr_line_prefix = ""
```
praiskup: It will probably work, though `stderr_line_prefix` isn't part of ` logging.Logger` api, and the solution thus isn't really clean.

mock/py/mockbuild/util.py
```diff
@@ -582,7 +582,7 @@ def logOutput(fdout, fderr, logger, returnOutput=1, start=0, timeout=0, printOut
                         if line != '':
                             line = ansi_escape.sub('', line)
                             if fderr is s and not line.startswith('+ '):
-                                logger.debug("BUILDSTDERR: %s", line)
+                                logger.debug("%s%s" % (logging.getLogger('mockbuild').stderr_line_prefix, line))
```
praiskup: Why can not we get the `config_opts['stderr_line_prefix']` value here?   It at least looks to be equivalent;  as long as we don't just do `logger.stderr_line_prefix`.

xsuchy: @praiskup There is no `config_opts` at this place of code. :(

praiskup: pff, ok, to not complicate this ... what about to "admit" this is hack by comment, and maybe namespace the attribute by `mock_` prefix so the future clash is unlikely....  Btw., why we don't use `logger.stderr_line_prefix`?

praiskup: If one wanted to have a bit more complicated life :-) there should be a way to solve this using logging.Formatter.

xsuchy:
LGTM, Thank you.

Issue #153, opened 2018-01-09
RFE: Propegate signals from the user process into the chroot


ifireball:
A typical scenario for using mock is to have some kind of an automation system (e.g. Jenkins) invoke mock to run things inside it in an isolated manner with fine control over the test environment.

An issue arises when an attempt is made to abort the test process (Either manually or due to an automated event such as a timeout). The automation system typically sends a signal to the processes it invoked to make them stop. With mock, the process the automation system sees is the consolekit wrapper that is used to run mock itself as root. When a signal is issue to it, the helper itself exits, but mock typically keeps on running as well as all the processes inside it.

I suggest a user-level executable will be added. That executable will establish some kind of a communication channel (E.g, a named pipe) and then invoke the root-level mock executable via consolekit. The root level executable would then listen on the communication channel for notifications about signals sent to the user-level process and propagate them to the processes it launched inside the chroot.

xsuchy:
Or litlte easier - mock will respond to SIGKILL and will execute --orpanskill which will kill everything started by mock.

plageat:
2 options came to mind:
- print pid of the running mock process somewhere, so that automation system can read and deal with it instead of consolehelper's pid
- there is a nice prctl.set_pdeathsig() function, so we can catch when parent is killed (consolehelper is killed in our case as I understand). But it introduce an extra dependency of prctl, cause it is not in stdlib :( For this option I even have a little patch if interested

ifireball:
@plageat - option #1 would require the automation system to have privileged access to the execution node other then being a member of the 'mock' group.

With option #2 I'm also not sure the signal can be routed across privilege boundaries.

plageat:
I don't really know how it works like, but I think it shouldn't be a problem for such a system to parse mock's sdout/stderr or read output logs otherwise it is a question how it can work at all?)

As for the second, did not understand well about privilege boundaries. Just works fine whenever you can send SIGKILL to the mock process - it kills itself and all his child processes inside it

Hojang2:
Hi, I'm just working on this issue. My solution would be to implement signal handler into the main mock thread. If handler would receive SIGTERM signal he would call orphansKill method to kill all processes related to mock. Does it fix your issue?

PR #293, opened 2019-07-18
Added signals handling by calling orphansKill for signals: SIGTERM, SIGHUP, SIGPIPE

Hojang2:


Hojang2:
Because after killing the main process, processes like dnf were still running. This PR fixes this problem.

Conan-Kudo:
This shouldn’t be the case if you’re using nspawn…

xsuchy:
This is trying to address https://github.com/rpm-software-management/mock/issues/153

.gitignore
```diff
@@ -21,3 +21,4 @@ bugzillas
 test-suite.log
 tests/runtests.sh.log
 tests/runtests.sh.trs
+.idea/
```
xsuchy: Can you put this in separate commit with the reasoning why you are adding it?

mock/py/mock.py
```diff
@@ -666,6 +674,8 @@ def main():
     util.setup_host_resolv(config_opts)
 
     buildroot = Buildroot(config_opts, uidManager, state, plugins, bootstrap_buildroot)
+    signal.signal(signal.SIGTERM, partial(handle_sigterm, buildroot))
+    print("************************* Signal handler active *********************")
```
xsuchy: log.info() would be much better here. Without those asterisks.

mock/py/mock.py
```diff
@@ -397,6 +399,12 @@ def command_parse():
     return (options, args)
 
 
+def handle_sigterm(buildroot, number, frame):
+    print("\nReceived SIGTERM activating orphansKill")
```
xsuchy: log.info() instead of print.

mock/py/mock.py
```diff
@@ -948,6 +958,7 @@ def run_command(options, args, config_opts, commands, buildroot, state):
     except (KeyboardInterrupt,):
         exitStatus = 7
         log.error("Exiting on user interrupt, <CTRL>-C")
+        #  TODO: Make this work util.orphansKill(buildroot.make_chroot_path())
```
xsuchy: can you just call it? :)

mock/py/mock.py
```diff
@@ -666,6 +674,8 @@ def main():
     util.setup_host_resolv(config_opts)
 
     buildroot = Buildroot(config_opts, uidManager, state, plugins, bootstrap_buildroot)
+    signal.signal(signal.SIGTERM, partial(handle_sigterm, buildroot))
```
xsuchy: Hmm, should we catch/process SIGABRT as well?

kdudka: On which occasions (other than internal error) does mock get `SIGABRT`?

I think it would make sense handle for example `SIGINT`, which is usually received when mock runs in the foreground process group and the user presses `^C`.

praiskup: E.g. `copr-rpmbuild` sends (on timeout) SIGINT down to mock as well.

Hojang2: Well, SIGINT is interpreted as KeyInterruptError in python and is already caught. As I mentioned below after KeyInterruptError all processes are killed, but it is not my work. If someone would test this too, it will be better.

kdudka: Indeed, `SIGINT` seems to be handled in recent versions of mock.

What about `SIGHUP` and `SIGPIPE`?

Try the following:
1. run `mock --chroot 'echo sleeping... && sleep 1024' &`
2. wait till `sleeping...` appears in your terminal
3. run `kill -SIGHUP $!` or `kill -SIGPIPE $!`
4. run `ps -A | grep sleep`

xsuchy: SIGHUP is a good idea.
SIGPIPE - I am not sure I want to encourage any use of pipes with mock. It has always been recommended to use logs with mock. Ẅith stdout, mock's code assumes human on the other side.

praiskup: But copr, koji and covscan are working with live logs...  which actually has potential to get broken pipes.

praiskup: That said - speaking about SIGPIPE - it would be nice to have at least some was to tell the mock that on stdout/stderr isn't consumed by human.

kdudka:
> SIGPIPE - I am not sure I want to encourage any use of pipes with mock. It has always been recommended to use logs with mock.

csmock has been using pipes with mock on both stdout and stderr since forever.

> With stdout, mock's code assumes human on the other side.

This could have been a valid expectation in late '90s.  In 2019 everything needs to be automated!

xsuchy:
At least since MSimacek Google Summer of Code work, there is a strong push in code to use separate logs. In Copr we use live log, but they are consumed by humans. But yeah, if csmock uses pipes, we can catch SIGPIPE as well.

kdudka:
OK, thank you for considering it!

Slightly off-topic but, at least `mock --shell` should never print anything on stdout because it could interfere with output of the command wrapped by mock.  For example, csmock uses the following command to get data out of mock's chroot:
```
mock --shell 'tar -c /builddir/cswrap-capture.err /builddir/cov ...' | tar -x ...
```

mock/py/mock.py
```diff
@@ -397,6 +399,12 @@ def command_parse():
     return (options, args)
 
 
+def handle_signals(buildroot, number, frame):
+    log.info("\nReceived SIGTERM activating orphansKill")
+    util.orphansKill(buildroot.make_chroot_path())
+    sys.exit(7)
```
praiskup: Exit status 7 isn't documented:
https://github.com/rpm-software-management/mock/blob/master/mock/py/mockbuild/exception.py#L26

kdudka: Good point.  The exit status should be `128 + signum` (to be compatible with shell interpreters).

Hojang2: So, if the signum is 13, then exit status should be 141?

praiskup: Since those exit status aren't unused, it sounds good to me.  Btw., is it safe to actually call exit in signal handler?

kdudka: In C run-time `_exit()` and `_Exit()` are async-signal-safe functions (because they do not call `atexit()`-registered callbacks).  No idea about python though...

Hojang2: > Exit status 7 isn't documented:
> https://github.com/rpm-software-management/mock/blob/master/mock/py/mockbuild/exception.py#L26

should I document all possible exit codes with signals? like 141, 143, 129

Hojang2: in the exceptions file?

kdudka: I would prefer to have them documented symbolically.  For example, this is how `bash(1)` man page documents the codes:

> The return value of a simple command is its exit status, or 128+n if the command is terminated by signal n.

xsuchy:
Yes, please document it in exception.py. I do not care if a symbolic way (128+X) or enumerate.

mock/py/mock.py
```diff
@@ -397,6 +399,20 @@ def command_parse():
...
 
 
+def handle_signals(buildroot, number, frame):
+    signal = ""
+    if number == 1:
+        signal = "SIGHUP"
+    elif number == 13:
+        signal = "SIGPIPE"
+    elif number == 15:
+        signal = "SIGTERM"
```
xsuchy: This can be better done as
```
signal_names = {1: "SIGHUP", 
               13, "SIGPIPE",
...
}
#all this can be defined at the top of file. and then
log.info("\nReceived signal {} activating orphansKill".format(signal_name[number]))
```
btw original name of variable `signal` override the name of the module. Not a good idea.

xsuchy:
LGTM, thank you

PR #305, opened 2019-08-14
Add procenv plugin for more detailed buildtime information

jcpunk:
This adds a plugin similar to hw_info.  Where hw_info produces detailed output on the hardware, this dumps a significant amount of information about the build time environment, any capabilities, and much more.

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -205,6 +205,9 @@
 # config_opts['plugin_conf']['hw_info_enable'] = True
 # config_opts['plugin_conf']['hw_info_opts'] = {}
 #
+# config_opts['plugin_conf']['procenv_enable'] = False
+# config_opts['plugin_conf']['procenv_opts'] = {}
+#
 # bind mount plugin is enabled by default but has no configured directories to
 # mount
 # config_opts['plugin_conf']['bind_mount_enable'] = True
--- a/mock/py/mockbuild/plugins/procenv.py
+++ b/mock/py/mockbuild/plugins/procenv.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+# vim:expandtab:autoindent:tabstop=4:shiftwidth=4:filetype=python:textwidth=0:
+# License: GPL2 or later see COPYING
+
+# python library imports
+import codecs
+
+# our imports
+from mockbuild.trace_decorator import getLog, traceLog
+import mockbuild.util
+
+requires_api_version = "1.1"
+
+
+# plugin entry point
+@traceLog()
+def init(plugins, conf, buildroot):
+    ProcEnv(plugins, conf, buildroot)
+
+
+class ProcEnv(object):
+    # pylint: disable=too-few-public-methods
+    """Get the runtime process environment"""
+    @traceLog()
+    def __init__(self, plugins, conf, buildroot):
+        self.buildroot = buildroot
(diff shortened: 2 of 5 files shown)


jcpunk:
Thanks pep8!  I've fixed it.

mock/py/mockbuild/plugins/procenv.py
```diff
@@ -0,0 +1,47 @@
...
+    @traceLog()
+    def _PreBuildHook(self):
+        getLog().info("enabled ProcEnv plugin")
+
+        self.buildroot.pkg_manager.install('procenv')
+
+        out_file = self.buildroot.resultdir + '/procenv.log'
+        with codecs.open(out_file, 'w', 'utf-8', 'replace') as out:
+
+            cmd = ["/usr/bin/procenv"]
```
xsuchy: This add new dependence on package `procenv`. It needs to be added either as Requires or Suggests/Recommends

mock/etc/mock/site-defaults.cfg
```diff
@@ -205,6 +205,9 @@
 # config_opts['plugin_conf']['hw_info_enable'] = True
 # config_opts['plugin_conf']['hw_info_opts'] = {}
 #
+# config_opts['plugin_conf']['procenv_enable'] = True
```
xsuchy: 1) you should set the default in util.py as well (search for `hw_info_enable` there)
2) the information in procenv is too much detailed IMHO. I would propose to not enable it by default.

mock/py/mockbuild/plugins/procenv.py
```diff
@@ -0,0 +1,47 @@
...
+    ProcEnv(plugins, conf, buildroot)
+
+
+class ProcEnv(object):
+    # pylint: disable=too-few-public-methods
+    """Get the runtime process environment"""
+    @traceLog()
+    def __init__(self, plugins, conf, buildroot):
+        self.buildroot = buildroot
+        self.hw_info_opts = conf
```
xsuchy: `hw_info_opts` - I guess this is left over from copy'n'paste from hw info plugin code :)

mock/py/mockbuild/plugins/procenv.py
jcpunk: Good call, I'll get that fixed.

mock/etc/mock/site-defaults.cfg
jcpunk: Works for me.  I'll confess I really like the extra information it provides... but I can turn it on where I'm using it easily enough.

mock/py/mockbuild/plugins/procenv.py
jcpunk: insert embarrassed emoji.

jcpunk:
In theory, this push resolves the reported issues.

mock/etc/mock/site-defaults.cfg
xsuchy: This should be False. In other words, here it states what is the default.

xsuchy:
Put before or after this line:
  https://github.com/jcpunk/mock/blob/procenv/mock/mock.spec#L80
line:
```
Suggests: procenv
```

jcpunk:
The listed review items should be fixed now.

xsuchy:
Thank you.
If you would document this plugin here https://github.com/rpm-software-management/mock/wiki#plugins you easy my life. Otherwise, I will do that during release.

PR #310, opened 2019-08-21
bootstrap-chroot: install shadow-utils by dnf as well

praiskup:
bootstrap-chroot: always explicitly install shadow-utils
    
    Previously we only did this explicitly for yum_install_command,
    apparently because yum itself (even transitively) never depended on
    shadow-utils.
    
    The 'dnf' package OTOH (mentioned in dnf_install_command and
    microdnf_install_command) used %systemd_requires before.  It means that
    also systemd package was installed as a dependency, and transitively
    /sbin/groupadd binary (shadow-utils).  So we didn't need to mention
    'shadow-utils' explicitly before.  This has changed recently in dnf [1],
    so we need to explicitly install shadow-utils for both
    {micro,}dnf_install_command.
    
    Per error:
    Finish(bootstrap): dnf install
    ERROR: Could not find useradd in chroot, maybe the install failed?
    
    [1] https://src.fedoraproject.org/rpms/dnf/c/faa199f16041940b85012fb8456b2fca349b6f9e?branch=master
    
    Fixes: rhbz#1743843

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -122,7 +122,7 @@
 # when 'use_bootstrap_container' is True, these commands are used to build
 # the minimal chroot for the respective package manager
 # config_opts['yum_install_command'] = 'install yum yum-utils shadow-utils distribution-gpg-keys'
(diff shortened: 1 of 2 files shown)


xsuchy:
Thank you

PR #315, opened 2019-08-22
mock: make only /proc and /sys "rprivate"

praiskup:
See the commit message for more info.

### DIFF
--- a/mock/py/mockbuild/mounts.py
+++ b/mock/py/mockbuild/mounts.py
@@ -142,22 +142,47 @@ class Mounts(object):
         self.essential_mounts = [] # /proc, /sys ... normally managed by systemd
         self.managed_mounts = []  # mounts owned by mock
         self.user_mounts = []  # mounts injected by user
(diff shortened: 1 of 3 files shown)


praiskup:
Though I'm not sure why e.g. `--installroot/--install` works without selinux plugin enabled.  It makes me feel that the selinux plugin is redundant entirely (at least after this PR).

praiskup:
- reworded comments and commit message
- /proc made non-recursive --bind

praiskup:
Ouch :-( neither /proc can be bind-mounted, I probably did something wrong before that it seemed to work ... :-(

praiskup:
Ok, doing the same thing for both /sys and /proc now.  I tried the mock in toolobox and --old-chroot seems to behave sane there.

praiskup:
With @xsuchy's help we added yet another verbose comment about what we are doing with `/ (shared) -> /sys (private) -> /sys (recursive, private) ...` mount tree..

PR #327, opened 2019-09-05
Changing cfg files for fedora 30 to use tpl file

schlupov:


mock-core-configs/etc/mock/fedora-30-aarch64.cfg
```diff
@@ -1,168 +1,5 @@
...
-#baseurl=http://download.fedoraproject.org/pub/fedora/linux/updates/$releasever/Modular/SRPMS/
-metalink=https://mirrors.fedoraproject.org/metalink?repo=updates-released-modular-source-f$releasever&arch=$basearch
-enabled=0
-repo_gpgcheck=0
-type=rpm
-gpgcheck=1
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
-skip_if_unavailable=False
-"""
+config_opts['baseurl'] = http://armpkgs.fedoraproject.org/repos/f30-build/latest/aarch64/
```
praiskup: This url doesn't work for me.

mock-core-configs/etc/mock/fedora-30-armhfp.cfg
```diff
@@ -1,174 +1,5 @@
...
-#baseurl=http://download.fedoraproject.org/pub/fedora/linux/updates/$releasever/Modular/SRPMS/
-metalink=https://mirrors.fedoraproject.org/metalink?repo=updates-released-modular-source-f$releasever&arch=$basearch
-enabled=0
-repo_gpgcheck=0
-type=rpm
-gpgcheck=1
-gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
-skip_if_unavailable=False
-"""
+config_opts['baseurl'] = https://kojipkgs.fedoraproject.org/repos/f30-build/latest/armhfp/
```
praiskup: You probably want to use $basearch here.

mock-core-configs/etc/mock/fedora-30.tpl
```diff
@@ -0,0 +1,184 @@
...
+[fedora-debuginfo]
+name=fedora-debuginfo
+metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-debug-$releasever&arch=$basearch
+enabled=0
+gpgkey=file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-30-primary
+gpgcheck=1
+skip_if_unavailable=False
+{% if target_arch == 'armv7hl' %}
+    failovermethod=priority
+{% endif %}
```
praiskup: Can we avoid this condition (isn't it actually bug in the old configuration)?

praiskup:
If you **only** moved one file into the template in first commit, and you made the rest of the changes in another commit it would be much easier to read and review (and you would protect the git from loosing the history, aka git blame functionality).

praiskup:
Also, what if we moved the templates into `/etc/mock/templates`?

mock-core-configs/etc/mock/fedora-30-aarch64.cfg
praiskup: You want to use https://kojipkgs.fedoraproject.org/repos/f30-build/latest/$basearch/ in all files, so you can actually avoid this.

nkadel:
@praiskup I think that moving the templates to /etc/mock/templates/ makes complete sense. I'd encourage it as a separate pull request, because it's a distinct issue and will touch a stack of config files.

Issue #199, opened 2018-06-20
RFC host configs


xsuchy:
Today mock parse:

1) site-defaults.cfg
2) chroot config (e.g. fedora-28-x86_64.cfg)
3) user configs

I want to add a new set of configs. E.g. `host-fedora-28.cfg`, `host-rhel-7.cfg`, etc. When mock will be run on RHEL7 then it will read `host-rhel-7.cfg`. When mock will be run on Fedora 28 then it will read `host-fedora-28.cfg` (if it exists).

This should solve some issue when some platform needs some different defaults.

I just wonder which order is better?:

1) first read chroot config and then host-* config or
2) first read host-* config and then read chroot config

Option 1 will allow me to override package manager for fedora chroots on RHEL7.
Option 2 .... are there some benefits?

Comments?

Conan-Kudo:
Host configs are not going to be useful except for exactly one case: bootstrap chroot. And potentially, not even then.

The assumption here is that the host configuration files will help with making package builds of newer Fedora releases on RHEL 7 work. But the answer to that is probably that it won't work even then. This is largely due to our usage of rich dependencies in the system. And this is not just in the `Requires` tag, but also in weak tags like `Recommends` and `Suggests`.

The thing is, RHEL 7's RPM and YUM is partially capable of processing weak dependencies, too, so that means that once it hits a weak tag that has a rich dep description, bad things are probably going to happen. This can definitely happen even in the bootstrap phase.

We get into ugly territory here, because RHEL 7 is too long in the tooth, and yet has been somewhat enhanced over the years to try to keep up a little bit with Fedora.

If you treat the package manager selection as an override, wouldn't that also mean that the wrong package manager would be used for selecting dependencies across the board, too? And with regular mode (no bootstrap), this would be much worse...

praiskup:
> This should solve some issue when some platform needs some different defaults.

This looks like the host-*.cfg is to some extent alternative to `site-defaults`, and thus
it looks like it should be parsed right after site-defaults.

jcpunk:
Would you be fetching the host type from `/etc/os-release` or some other detection?

xsuchy:
> Would you be fetching the host type from /etc/os-release or some other detection?

I would like to rely on python-distro detection.

voxik:
I am looking for something more generic, like configurations for nested roots.

[...]

PR #329, opened 2019-09-06
Activate use_bootstrap_container for dnf enabled config files on RHEL < 8

nkadel:
Issue 39 about bootstrap_container settings, addressed as needed in /etc/mock/ configs.

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -49,7 +49,15 @@ Config files which allow you to create chroots for:
 
 
 %build
-# nothing to do here
+
+%if 0%{?rhel} > 0 && 0%{?rhel} < 8
+# Required for dnf enabled ocnfigs on yum based hosts
+# config_opts['use_bootstrap_container'] = True
+grep -rl "config_opts\['package_manager'\] = 'dnf'" etc/mock | \
+    while read name; do
+    sed -i.bak "s/config_opts\['package_manager'\] = 'dnf'/config_opts\['package_manager'\] = 'dnf'\n# Enable bootstrap on dnf host for %%rhel %{rhel}\nconfig_opts\[\'use_bootstrap_container\'\] = True\n\n/g" $name
+done
+%endif # rhel && rhel < 8
 
 
 %install


Conan-Kudo:
@nkadel DNF exists on RHEL 7, so why is bootstrap required?

nkadel:
Because it doesn’t work for mock.

mock-core-configs/mock-core-configs.spec
```diff
@@ -49,7 +49,15 @@ Config files which allow you to create chroots for:
...
 %build
-# nothing to do here
+
+%if 0%{?rhel} > 0 && 0%{?rhel} < 8
+# Required for dnf enabled ocnfigs on yum based hosts
+# config_opts['use_bootstrap_container'] = True
+grep -rl "config_opts\['package_manager'\] = 'dnf'" etc/mock | \
+    while read name; do
+    sed -i.bak "s/config_opts\['package_manager'\] = 'dnf'/config_opts\['package_manager'\] = 'dnf'\n# Enable bootstrap on dnf host for %%rhel %{rhel}\nconfig_opts\[\'use_bootstrap_container\'\] = True\n\n/g" $name
+done
```
praiskup: I dislike this;  I don't think it is wise to change the configuration only for RHEL.  Both dnf may be installed there, and users might only want to build in RHEL <= 7 chroots on RHEL 7.  And still, it is config only - we can rely on users to adjust the configuration accordingly.

If anything, we should have host-specific configuration, and only for those configuration files where this is absolutely necessary (fedora, rhel8+, ...)

nkadel: But the problem is not host specific. It is distribution specific. It applies only to specific operating systems which lack a functional native dnf (which I've so far only aidentified as RHEL based systems before RHEL 8), and only certain target configuration files (those which use on dnf). It's therefore not appropriate to site-defaults.cfg. or any equivalent.

If there is some graceful "if config_opts['package_manager'] = dnf" logic we can embed in these .cfg files, great! But in the meantime, this logic works well.

Conan-Kudo: We have a functional DNF in RHEL 7. Mock already isn't being updated in RHEL 6, so it's irrelevant for this matter.

Conan-Kudo: The issue with RHEL 7 hosts isn't DNF, but RPM. All currently supported Fedora releases and RHEL 8 use RPM features that don't exist in RHEL 7, which breaks your ability to create buildroots of those distributions.

For example, Mageia 6 and Mageia 7 buildroots work just fine on RHEL 7 because those features aren't in use in the core, permitting bootstrap chroots to work.

nkadel: I'd thought that those failures were tied to dnf, not RPM itself. If so, can you think of any better way to tetect what the local RPM  version is set to versus the alternative explanations? Outguessing which "cores" have, use or do not use such features sounds like a losing prospect.

xsuchy:
by host configs @praiskup meant https://github.com/rpm-software-management/mock/issues/199

I dislike this change as well and I am not going to merge it. I want to rather go the host-configs way.

xsuchy:
I am going to close this one. Several people including me do not like it. This can be better solved by host configs. Or using the containers, see https://github.com/rpm-software-management/mock/pull/342

Issue #308, opened 2019-08-20
RFE: bind mount distribution-gpg-keys for bootstrap chroot


praiskup:
Instead of installing the `distribution-gpg-keys` package into bootstrap chroot, it should be enough to just bind-mount (or copy) the host's directory `/usr/share/distribution-gpg-keys/` into bootstrap chroot.

nkadel:
Only if the host's directory is sufficiently recent. This is not predictable, especially for RHEL, which is unlikely to insist on publishing up-to-date GPG keys for bleeding edge fedora entries.

praiskup:
I'd say it is fair to delegate this responsibility on user (keeping the gpg keys up2date).  Also note the benefit of slightly smaller bootstrap chroot.

nkadel:
I'd say that most of us could not care less about such a small increment in the size of the chroot tarball. Reducing the network hit against an offsite RPM repository is more important, and there are some security benefits of *not* updating such packages every time from offsite. Perhaps a small comment saying "if necessary, install this RPM to get the bleeding edge GPG keys" ?

xsuchy:
> This is not predictable, especially for RHEL, which is unlikely to insist on publishing up-to-date GPG keys for bleeding edge fedora entries.

I already do that using package distribution-gpg-keys which is released for EPEL as well.

xsuchy:
As of commit 63ab868 the configs are always copied into bootstrap chroot.
I will leave this open. And once 1.5.0 will be release (still lot of time ahead) we can remove the 'distribution-gpg-keys' package from configs and call it a day.

PR #342, opened 2019-09-24
Use podman to pull and export an image as a bootstrap chroot.

dmach:
Running mock for the latest distro on an old one doesn't always work mainly
due to a new compression format or a new checksum algorithm. Using the image
skips the installation step and allows using the latest runtime directly.

To turn this feature on, add following to the config file:
config_opts['use_bootstrap_image'] = True

You can also override image tag (defaults to 'fedora:latest'):
config_opts['bootstrap_image'] = 'fedora:latest'

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -60,6 +60,9 @@ class Buildroot(object):
         self.env.update(proxy_env)
         os.environ.update(proxy_env)
 
+        self.use_bootstrap_image = self.config['use_bootstrap_image']
+        self.bootstrap_image = self.config['bootstrap_image']
+
         self.pkg_manager = package_manager(config, self, plugins, bootstrap_buildroot)
         self.mounts = mounts.Mounts(self)
 
@@ -250,6 +253,9 @@ class Buildroot(object):
 
     @traceLog()
     def _init_pkg_management(self):
+        if self.is_bootstrap and self.use_bootstrap_image:
+            getLog().debug("Skipping package management init in the bootstrap chroot due to using bootstrap image")
+            return
         update_state = '{0} install'.format(self.pkg_manager.name)
         self.state.start(update_state)
         if 'module_enable' in self.config and self.config['module_enable']:
--- a/mock/py/mockbuild/plugins/root_cache.py
+++ b/mock/py/mockbuild/plugins/root_cache.py
@@ -138,9 +138,14 @@ class RootCache(object):
             mockbuild.util.do(cmd, printOutput=True)
 
             # start a container and detach immediately
-            cmd = ["podman", "run", "--detach", self.buildroot.bootstrap_image, "/bin/true"]
+            cmd = ["podman", "run", "-it", "--detach", self.buildroot.bootstrap_image, "/bin/bash"]
             container_id = mockbuild.util.do(cmd, returnOutput=True)
             container_id = container_id.strip()
+
+            # make sure the image contains expected packages
+            cmd = ["podman", "exec", container_id, self.buildroot.pkg_manager.command, "-y"]
+            cmd += self.buildroot.pkg_manager.install_command.split()
+            mockbuild.util.do(cmd, printOutput=True)
 
             # export container and compress it
             getLog().info("Exporting container: %s as %s", self.buildroot.bootstrap_image, self.rootCacheFile)
@@ -154,7 +159,7 @@ class RootCache(object):
             cache_file.close()
 
             # remove the container
-            cmd = ["podman", "rm", container_id]
+            cmd = ["podman", "rm", "-f", container_id]
             mockbuild.util.do(cmd)
 
         # optimization: don't unpack root cache if chroot was not cleaned (unless we are using tmpfs)
--- a/mock/py/mockbuild/plugins/root_cache.py
+++ b/mock/py/mockbuild/plugins/root_cache.py
@@ -7,6 +7,7 @@
 # python library imports
 import fcntl
 import os
+import subprocess
 import time
 
 # our imports
@@ -127,6 +128,35 @@ class RootCache(object):
         if self.rootCacheLock is None:
             self.rootCacheLock = open(os.path.join(self.rootSharedCachePath, "rootcache.lock"), "a+")
 
+        if self.buildroot.is_bootstrap and self.buildroot.use_bootstrap_image \
+                and not os.path.exists(self.rootCacheFile):
+            getLog().info("Using bootstrap image: %s", self.buildroot.bootstrap_image)
+
+            # pull the latest image
+            getLog().info("Pulling image: %s", self.buildroot.bootstrap_image)
+            cmd = ["podman", "pull", self.buildroot.bootstrap_image]
+            mockbuild.util.do(cmd, printOutput=True)
+
+            # start a container and detach immediately
+            cmd = ["podman", "run", "--detach", self.buildroot.bootstrap_image, "/bin/true"]
+            container_id = mockbuild.util.do(cmd, returnOutput=True)
+            container_id = container_id.strip()
+
+            # export container and compress it
+            getLog().info("Exporting container: %s as %s", self.buildroot.bootstrap_image, self.rootCacheFile)
+            cmd_podman = ["podman", "export", container_id]
+            podman = subprocess.Popen(cmd_podman, stdout=subprocess.PIPE)
+            cache_file = open(self.rootCacheFile, "w")
+            cmd_compressor = [self.compressProgram, "--stdout"]
+            compressor = subprocess.Popen(cmd_compressor, stdin=podman.stdout, stdout=cache_file)
+            compressor.communicate()
+            podman.communicate()
+            cache_file.close()
+
+            # remove the container
+            cmd = ["podman", "rm", container_id]
+            mockbuild.util.do(cmd)
+
         # optimization: don't unpack root cache if chroot was not cleaned (unless we are using tmpfs)
         if os.path.exists(self.rootCacheFile):
             if (not self.buildroot.chroot_was_initialized or self._haveVolatileRoot()):
@@ -150,6 +180,17 @@ class RootCache(object):
                 )
                 for item in self.exclude_dirs:
                     mockbuild.util.mkdirIfAbsent(self.buildroot.make_chroot_path(item))
+
+                if self.buildroot.is_bootstrap and self.buildroot.use_bootstrap_image:
+                     # The bootstrap image frequently lacks distribution-gpg-keys.
+                     # Copy the files from the host to avoid invoking package manager
+                     # or rebuilding the cached bootstrap chroot.
+                     keys_path = "/usr/share/distribution-gpg-keys"
+                     dest_path = os.path.dirname(keys_path)
+                     getLog().debug("Copying %s to the bootstrap chroot" % keys_path)
+                     cmd = ["cp", "-a", keys_path, self.buildroot.make_chroot_path(dest_path)]
+                     mockbuild.util.do(cmd)
+
                 self._rootCacheUnlock()
                 self.buildroot.chrootWasCached = True
                 self.state.finish("unpacking root cache")
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -126,6 +126,11 @@
 # config_opts['system_yum_command'] = '/usr/bin/yum'
 # config_opts['system_dnf_command'] = '/usr/bin/dnf'
 
+# The bootstrap chroot is installed using a package manager by default,
+# but in some cases you may want it populated from a container image instead.
+# config_opts['use_bootstrap_image'] = False
+# config_opts['bootstrap_image'] = 'fedora:latest'
+
 # anything you specify with 'bootstrap_*' will be copied to bootstrap config
 # e.g. config_opts['bootstrap_system_yum_command'] = '/usr/bin/yum-deprecated' will become
 # config_opts['system_yum_command'] = '/usr/bin/yum-deprecated' for bootstrap config
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -79,6 +79,7 @@ Recommends: btrfs-progs
 Recommends: dnf-utils
 Suggests: qemu-user-static
 Suggests: procenv
+Suggests: podman
 %else
 %if 0%{?rhel} == 7
 Requires: btrfs-progs
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -675,6 +675,9 @@ def main():
         bootstrap_buildroot_config['yum_command'] = bootstrap_buildroot_config['system_yum_command']
         bootstrap_buildroot_config['dnf_command'] = bootstrap_buildroot_config['system_dnf_command']
 
+        # disable updating bootstrap chroot
+        bootstrap_buildroot_config['update_before_build'] = False
+
         bootstrap_buildroot_state = State(bootstrap=True)
         bootstrap_plugins = Plugins(bootstrap_buildroot_config, bootstrap_buildroot_state)
         bootstrap_buildroot = Buildroot(bootstrap_buildroot_config,
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -931,6 +931,9 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['use_container_host_hostname'] = True
     config_opts['use_bootstrap_container'] = False
 
+    config_opts['use_bootstrap_image'] = False
+    config_opts['bootstrap_image'] = 'fedora:latest'
+
     config_opts['internal_dev_setup'] = True
 
     # cleanup_on_* only take effect for separate --resultdir


mock/py/mockbuild/plugins/root_cache.py
```diff
@@ -127,6 +129,35 @@ def _unpack_root_cache(self):
         if self.rootCacheLock is None:
             self.rootCacheLock = open(os.path.join(self.rootSharedCachePath, "rootcache.lock"), "a+")
 
+        if self.buildroot.is_bootstrap and self.buildroot.use_bootstrap_image \
+                and not os.path.exists(self.rootCacheFile):
+            getLog().info("Using bootstrap image: %s" % self.buildroot.bootstrap_image)
```
xsuchy: Use lazy evaluation. https://stackoverflow.com/questions/4148790/lazy-logger-message-string-evaluation

xsuchy:
podman should be added to spec file as soft dependence

mock/py/mockbuild/buildroot.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, config, uid_manager, state, plugins, bootstrap_buildroot=None
         self.env.update(proxy_env)
         os.environ.update(proxy_env)
 
+        self.use_bootstrap_image = self.config['use_bootstrap_image']
+        self.bootstrap_image = self.config['bootstrap_image']
```
xsuchy: This needs to have defaults in utils.py and documented (commented out default) in `site-defaults.cfg`

praiskup:
We should probably stop doing any work in bootstrap chroot when this feature is enabled, otherwise the stuff in bootstrap chroot gets reinstalled:
```
praiskup@host-10-0-136-252 ~]$ mock -r epel-7-x86_64 --bootstrap --shell                     
INFO: mock.py version 1.4.19 starting (python version = 3.7.4)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start(bootstrap): chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: Using bootstrap image: fedora:latest
INFO: Pulling image: fedora:latest
Trying to pull docker.io/library/fedora:latest...
Getting image source signatures
Copying blob sha256:5a915a173fbc36dc8e1410afdd9de2b08f71efb226f8eb1ebcdc00a1acbced62
Copying config sha256:e9ed59d2baf72308f3a811ebc49ff3f4e0175abf40bf636bea0160759c637999
Writing manifest to image destination
Storing signatures
e9ed59d2baf72308f3a811ebc49ff3f4e0175abf40bf636bea0160759c637999
INFO: Exporting container: fedora:latest as /var/cache/mock/epel-7-x86_64-bootstrap/root_cache/cache.tar.gz
Start(bootstrap): unpacking root cache
Finish(bootstrap): unpacking root cache
INFO: enabled yum cache
Start(bootstrap): cleaning yum metadata
Finish(bootstrap): cleaning yum metadata
INFO: enabled HW Info plugin
Start(bootstrap): yum install
BaseOS                                                                                                                                                                                                       1.0 MB/s |  10 MB     00:09    
updates                                                                                                                                                                                                      635 kB/s | 1.7 MB     00:02    
epel                                                                                                                                                                                                         2.2 MB/s |  16 MB     00:07    
extras                                                                                                                                                                                                       212 kB/s | 251 kB     00:01    
Package shadow-utils-2:4.6-8.fc30.x86_64 is already installed.
Error: 
 Problem 1: package dnf-yum-4.2.7-2.fc30.noarch conflicts with yum < 3.4.3-505 provided by yum-3.4.3-163.el7.centos.noarch
  - conflicting requests
  - problem with installed package dnf-yum-4.2.7-2.fc30.noarch
 Problem 2: package libxml2-python-2.9.1-6.el7_2.3.x86_64 requires libpython2.7.so.1.0()(64bit), but none of the providers can be installed
  - package python-libs-2.7.5-86.el7.x86_64 requires libgdbm.so.4()(64bit), but none of the providers can be installed
  - package yum-utils-1.1.31-52.el7.noarch requires libxml2-python, but none of the providers can be installed
  - installed package gdbm-libs-1:1.18-4.fc30.x86_64 obsoletes gdbm < 1:1.14.1-4 provided by gdbm-1.10-8.el7.x86_64
  - conflicting requests
  - problem with installed package gdbm-libs-1:1.18-4.fc30.x86_64
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
ERROR: Command failed: 
 # /usr/bin/yum --installroot /var/lib/mock/epel-7-x86_64-bootstrap/root/ --releasever 7 install yum yum-utils shadow-utils distribution-gpg-keys --disablerepo=sclo* --setopt=tsflags=nocontexts
BaseOS                                                                                                                                                                                                       1.0 MB/s |  10 MB     00:09    
updates                                                                                                                                                                                                      635 kB/s | 1.7 MB     00:02    
epel                                                                                                                                                                                                         2.2 MB/s |  16 MB     00:07    
extras                                                                                                                                                                                                       212 kB/s | 251 kB     00:01    
Package shadow-utils-2:4.6-8.fc30.x86_64 is already installed.
Error: 
 Problem 1: package dnf-yum-4.2.7-2.fc30.noarch conflicts with yum < 3.4.3-505 provided by yum-3.4.3-163.el7.centos.noarch
  - conflicting requests
  - problem with installed package dnf-yum-4.2.7-2.fc30.noarch
 Problem 2: package libxml2-python-2.9.1-6.el7_2.3.x86_64 requires libpython2.7.so.1.0()(64bit), but none of the providers can be installed
  - package python-libs-2.7.5-86.el7.x86_64 requires libgdbm.so.4()(64bit), but none of the providers can be installed
  - package yum-utils-1.1.31-52.el7.noarch requires libxml2-python, but none of the providers can be installed
  - installed package gdbm-libs-1:1.18-4.fc30.x86_64 obsoletes gdbm < 1:1.14.1-4 provided by gdbm-1.10-8.el7.x86_64
  - conflicting requests
  - problem with installed package gdbm-libs-1:1.18-4.fc30.x86_64
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

mock/py/mockbuild/plugins/root_cache.py
```diff
@@ -7,11 +7,13 @@
 # python library imports
 import fcntl
 import os
+import subprocess
 import time
 
 # our imports
 from mockbuild.trace_decorator import getLog, traceLog
 import mockbuild.util
+from mockbuild.util import do
```
xsuchy: Can you instead of `do()` call `mockbuild.util.do()`? It is just nitpicking, but it is used this way everywhere in the code.

mock/py/mockbuild/plugins/root_cache.py
```diff
@@ -127,6 +129,35 @@ def _unpack_root_cache(self):
...
+
+            # export container and compress it
+            getLog().info("Exporting container: %s as %s" % (self.buildroot.bootstrap_image, self.rootCacheFile))
+            cmd_podman = ["podman", "export", container_id]
+            podman = subprocess.Popen(cmd_podman, stdout=subprocess.PIPE)
+            cache_file = open(self.rootCacheFile, "w")
+            cmd_compressor = [self.compressProgram, "--stdout"]
+            compressor = subprocess.Popen(cmd_compressor, stdin=podman.stdout, stdout=cache_file)
+            compressor.communicate()
+            podman.communicate()
```
xsuchy: Are you sure this will work? I was under the impression, that communicate() cannot be used for large data and you have to use select().

xsuchy:
> We should probably stop doing any work in bootstrap chroot when this feature is enabled, otherwise the stuff in bootstrap chroot gets reinstalled:

*nod* This can be done either here https://github.com/rpm-software-management/mock/blob/devel/mock/py/mock.py#L711 or here https://github.com/rpm-software-management/mock/blob/devel/mock/py/mockbuild/buildroot.py#L263

xsuchy:
> Problem 1: package dnf-yum-4.2.7-2.fc30.noarch conflicts with yum < 3.4.3-505 provided by yum-3.4.3-163.el7.centos.noarch
On the other hand using fedora:latest to install el7 is... I really do not want to support this. And therefore not even test it. We should use the image of the target system. I.e. centos:7 for epel-7*  It can be even part of the default config as setting `bootstrap_image` will make no harm/change untill we set `use_bootstrap_image`.

Conan-Kudo:
@praiskup We still need to run them because the container isn't necessarily prepared properly to run mock. For example, `dnf` is not in the `opensuse/leap:15` or `opensuse/tumbleweed` images, and we need bootstrap to install it. Same goes for `centos:7` when wanting to use DNF with Mock.

Conan-Kudo:
@praiskup Of course, we could simply have some dedicated Mock bootstrap images, which would work in this scenario. And perhaps that makes sense. We could make some images hosted on quay.io for this.

praiskup:
Yes, if we had to run the hosts package manager the whole point of this PR would be lost.  So perhaps we need something larger then minimal image.

mock/py/mockbuild/plugins/root_cache.py
dmach: Could you be more specific on how I should use the lazy evaluation?

mock/py/mockbuild/plugins/root_cache.py
dmach: No, that's perfectly fine. I've changed that.

mock/py/mockbuild/plugins/root_cache.py
dmach: I believe this mimics pipe in shell.
Write blocks on full buffer, read blocks on reading from an empty buffer.

AFAIK select() is necessary if you need non-blocking operations.

mock/py/mockbuild/buildroot.py
dmach: > > Problem 1: package dnf-yum-4.2.7-2.fc30.noarch conflicts with yum < 3.4.3-505 provided by yum-3.4.3-163.el7.centos.noarch
> 
> On the other hand using fedora:latest to install el7 is... I really do not want to support this. And therefore not even test it. We should use the image of the target system. I.e. centos:7 for epel-7* It can be even part of the default config as setting `bootstrap_image` will make no harm/change untill we set `use_bootstrap_image`.

Do you want me set 'bootstrap_image' in all config files?
Finding and testing the images might be a bit challenging and definitely will take time.

dmach:
I've pushed requested fixes, keeping separated from the original comment for easier review.
I will squash them before the final merge.

mock/py/mockbuild/buildroot.py
xsuchy: a) it can be done in separate PR b) I and @frostyx can help with the testing

xsuchy:
We can document that the container image has to have a package manager in it.

mock/py/mockbuild/plugins/root_cache.py
```diff
@@ -150,6 +180,16 @@ def _unpack_root_cache(self):
...
                     mockbuild.util.mkdirIfAbsent(self.buildroot.make_chroot_path(item))
+
+                if self.buildroot.is_bootstrap and self.buildroot.use_bootstrap_image:
+                     # The bootstrap image frequently lacks distribution-gpg-keys.
+                     # Copy the files from the host to avoid invoking package manager
+                     # or rebuilding the cached bootstrap chroot.
+                     keys_path = "/usr/share/distribution-gpg-keys"
+                     getLog().debug("Copying '%s' to the bootstrap chroot", keys_path)
+                     cmd = ["cp", "-a", keys_path, self.buildroot.make_chroot_path(keys_path)]
+                     mockbuild.util.do(cmd)
```
praiskup: This is basically https://github.com/rpm-software-management/mock/issues/308 but not bind-mount, and into bootstrap chroot only.

Just saying it would be nice if we could have this done even without `self.buildroot.is_bootstap_image` condition, but it can be done later.  Also note that bind mount would be slightly nicer I guess.

xsuchy:
I have done some testing and it works for me. Nice work.

I will keep it open over the weekend to gather feedback. If no one raises anything critical I am going to merge this on Monday morning.

Conan-Kudo:
> We can document that the container image has to have a package manager in it.

Note that it *must* contain the `yum` or `dnf` package managers, and that's an important detail.

praiskup:
I installed podman on epel-7 box, and `mock -r fedora-rawhide-x86_64 --shell --bootstrap-chroot` worked fine for me.  Nice!

> Note that it must contain the yum or dnf package managers, and that's an important detail.

To solve zstd compression issues, it is even important to know how `rpm` is compiled in the image.

praiskup:
`--rebuild` failed for me because dnf-plugins-core isn't installed in the `fedora:rawhide`:
```
No such command: builddep. Please use /usr/bin/dnf --help
It could be a DNF plugin command, try: "dnf install 'dnf-command(builddep)'"
```

And zstd support doesn't seem to be available in `fedora:latest`:
```
ERROR You need to update rpm to handle:
rpmlib(PayloadIsZstd) <= 5.4.18-1 is needed by pcre2-10.33-14.fc32.x86_64
rpmlib(PayloadIsZstd) <= 5.4.18-1 is needed by ca-certificates-2019.2.32-3.fc31.noarch
```

dmach:
> `--rebuild` failed for me because dnf-plugins-core isn't installed in the `fedora:rawhide`:
> 
> ```
> No such command: builddep. Please use /usr/bin/dnf --help
> It could be a DNF plugin command, try: "dnf install 'dnf-command(builddep)'"
> ```
This is a problem. The dnf-utils package is not part of the image.
I don't think we want to be in the image maintenance business, we should rather focus on using already available images.

Maybe we need revert my change in _init_pkg_management() and modify it to use dnf from the bootstrap chroot to install remaining packages.

FrostyX:
> --rebuild failed for me because dnf-plugins-core isn't installed in the fedora:rawhide

I believe that this can be fixed by _chrooting_ into the bootstrap buildroot and running `_init_pkg_management` to modify the bootstrap buildroot itself. Does it make any sense? Of course this would have to be done only if `use_bootstrap_image` is `True`.

I've been trying to implement it, but I wasn't successful (yet). I would expect that something like this:

         @traceLog()
         def _init_pkg_management(self):
    -        # if self.is_bootstrap and self.use_bootstrap_image:
    -        #     getLog().debug("Skipping package management init in the bootstrap chroot due to using bootstrap image")
    -        #     return
    +        if self.is_bootstrap and self.use_bootstrap_image:
    +            self.pkg_manager.bootstrap_buildroot = self

and setting

    config_opts['chroot_additional_packages'] = ['dnf-plugins-core']

which IMHO shouldn't even be necessary, should do the trick ...

praiskup:
> I believe that this can be fixed by chrooting into the bootstrap

Using `podman exec <the_container> dnf install ....` is probably easier.

FrostyX:
> Using podman exec <the_container> dnf install .... is probably easier.

I agree, but I avoided that because I suppose that we don't want to install `dnf-plugins-core` into all containers (e.g. epel-7 chroots and opensuse chroots doesn't have DNF installed). By doing `dnf install` in podman, we would break this feature for those chroots (which may not be a problem though). Or we can do that install only conditionally and check, whether there is a DNF in the container. What do you think?

FrostyX:
> Using podman exec <the_container> dnf install .... is probably easier.

This should do the trick 

[0001-conditionally-install-dnf-plugins-core-into-podman-c.patch.txt](https://github.com/rpm-software-management/mock/files/3665417/0001-conditionally-install-dnf-plugins-core-into-podman-c.patch.txt)

praiskup:
I don't think the heuristic from inside the container isn't the best idea; and I expect similar problems with yum-based systems.  We should basically either re-use `dnf_install_command` and `yum_install_command`, or if that's considered too hacky (I don't think so) we should have a new option..  As @dmach/@xsuchy agreed above, we'll probably have different configuration for bootstrap image per chroot anyway.

FrostyX:
> We should basically either re-use dnf_install_command and yum_install_command, or if that's considered too hacky (I don't think so) 

Oooh, that's good! PTAL on this patch.
[0001-run-dnf-yum-install-command-also-in-the-podman-conta.patch.txt](https://github.com/rpm-software-management/mock/files/3665529/0001-run-dnf-yum-install-command-also-in-the-podman-conta.patch.txt)

FrostyX:
I think, that it is also a good idea to add command-line options to turn on/off this feature because I would expect it to not be enabled by default in mock configs.
[0001-add-commandline-options-for-using-bootstrap-image.patch.txt](https://github.com/rpm-software-management/mock/files/3665746/0001-add-commandline-options-for-using-bootstrap-image.patch.txt)

FrostyX:
I've also tried to set a proper podman image per mock config. It didn't go well for all configs, please look on the commit message.
[0001-configure-podman-containers-for-Fedora-EPEL-and-Mage.patch.txt](https://github.com/rpm-software-management/mock/files/3667289/0001-configure-podman-containers-for-Fedora-EPEL-and-Mage.patch.txt)

Conan-Kudo:
@berolinux Can you look into the issue with the OMV cooker container reported in the patch attached above?

xsuchy:
> I think, that it is also a good idea to add command-line options to turn on/off this feature because I would expect it to not be enabled by default in mock configs.
> [0001-add-commandline-options-for-using-bootstrap-image.patch.txt](https://github.com/rpm-software-management/mock/files/3665746/0001-add-commandline-options-for-using-bootstrap-image.patch.txt)

This would require a change in man page as well. But please create separate PR for that. And for those configs as well.

FrostyX:
> This would require a change in man page as well. But please create separate PR for that. And for those configs as well.

Well, there is a man page change. Is it missing something?
I've created a PR#352 for this change and PR#353 for the configs change. Let's discuss them separately.

I haven't submitted a PR for this one

> Oooh, that's good! PTAL on this patch.
> 0001-run-dnf-yum-install-command-also-in-the-podman-conta.patch.txt

because in case we want the change, I think, that it should be a part of this PR.

dmach:
I've force-pushed squashed commits + additional fixes:
* FrostyX's patch to install additional packages into the container
* disabled bootstrap chroot updates - they do not work due to package manager incompatibilities

I hope this PR is close to it's final shape.

xsuchy:
Thank you

PR #343, opened 2019-09-24
Add Amazon Linux 2 configs

haroldjin:
This PR adds amazon linux 2 config files with references to gpg keys provided by this [PR for distribution-gpg-keys](https://github.com/xsuchy/distribution-gpg-keys/pull/24/commits/3601e86a359892eea54e57d3e900199b9ee4ed5b).

### DIFF
--- a/mock-core-configs/etc/mock/amazonlinux-2-aarch64.cfg
+++ b/mock-core-configs/etc/mock/amazonlinux-2-aarch64.cfg
@@ -1,0 +1,5 @@
+include('/etc/mock/templates/amazonlinux-2.tpl')
+
+config_opts['root'] = 'amzn-2-aarch64'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/amazonlinux-2-x86_64.cfg
+++ b/mock-core-configs/etc/mock/amazonlinux-2-x86_64.cfg
@@ -1,0 +1,5 @@
+include('/etc/mock/templates/amazonlinux-2.tpl')
+
+config_opts['root'] = 'amzn-2-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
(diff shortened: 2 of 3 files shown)


xsuchy:
Can you please use templates and put there the common parts. Please check e.g., the recent epel-8-x86_64 config.

mock-core-configs/etc/mock/amazonlinux-2-aarch64.cfg
```diff
@@ -0,0 +1,25 @@
+config_opts['root'] = 'amzn2-aarch64'
```
xsuchy: We usually had two dashes in the name of the root. Following the patter $name-$version-$arch. I would like to see here 'amzn-2-aarch64'. This is not mandatory in mock itself. But I have seen some code based on top of mock, which assumes this schema.

haroldjin: @xsuchy, thanks for the feedback. I've made changes to to ensure $name$-version-$arch is enforced as well as using config template.

praiskup:
Can you please squash the commits?

praiskup:
mock throws error with this config:
```
$ mock -r amazonlinux-2-x86_64 --shell
  File "/etc/mock/amazonlinux-2-x86_64.cfg", line 1, in <module>
    include('/etc/mock/templates/amazonlinux-2.tpl')

  File "/etc/mock/templates/amazonlinux-2.tpl", line 1

    config_opts['chroot_setup_cmd'] = 'install @buildsys-build yum’

                                                                  ^

SyntaxError: EOL while scanning string literal

ERROR: Error in configuration
```

haroldjin:
@praiskup, thanks for the comment. I have squashed the comment and fixed the apostrophe character.

praiskup:
Mock fails for me right after
```
 # /usr/bin/yum --installroot /var/lib/mock/amzn-2-x86_64/root/ --releasever 2 install @buildsys-build yum --setopt=tsflags=nocontexts
```
If this is working fine _on amzn host_, I would say we could merge this in.  But it should be documented somewhere that it doesn't work elsewhere.

PR #352, opened 2019-09-30
add commandline options for using bootstrap image

FrostyX:
This is a follow-up for PR#342.

### DIFF
--- a/mock/docs/mock.1
+++ b/mock/docs/mock.1
@@ -383,6 +383,17 @@ build in two stages, using chroot rpm for creating the build chroot
 \fB\-\-no-bootstrap-chroot\fR
 build in a single stage, using system rpm for creating the build chroot
 
+.TP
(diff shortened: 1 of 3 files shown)


mock/docs/mock.1
```diff
@@ -383,6 +383,13 @@ build in two stages, using chroot rpm for creating the build chroot
 \fB\-\-no-bootstrap-chroot\fR
 build in a single stage, using system rpm for creating the build chroot
 
+.TP
+\fB\-\-use-bootstrap-image\fR
+create bootstrap chroot from container image
```
xsuchy: Can you be more verbose here? What it actually does.

FrostyX:
> Can you be more verbose here? What it actually does.

Like this, or even more?

xsuchy:
Thank you

PR #353, opened 2019-09-30
configure podman containers for Fedora, EPEL and Mageia

FrostyX:
This is a follow-up for PR#342

At the moment of creating this commit, all Fedora configs works
and it is possible to build in them using the podman image.

Both EPEL 6 and 7 works too, but there is no available image for
EPEL 8 yet. I've configured it though in order to have it prepared
once the image is ready.

Neither Mageia 6 nor 7 works properly. Mageia 6 fails with DNF
traceback and Mageia 7 fails because of failing to sync repos.
I guess those might fix _themselves_.

I haven't modified any other mock configs because

1) Those systems don't have Dnf/Yum (cauldron, opensuse)
2) I didn't know what are the official podman images for them
   (rhel, centos-stream)
3) It just didn't work and failed with traceback
   (amazonlinux, openmandriva-cooker)


xsuchy:
Merging even before PR#342 as it does not break anything.

PR #356, opened 2019-10-01
remove mockchain [RHBZ#1757388]

xsuchy:


xsuchy:
I will leave it here for some comments. If there will be no objections, I will merge it by end of the week.

mock/py/mockchain.py
```diff
@@ -2,422 +2,7 @@
...
-    parser.add_option(
-        '--log', default=None, dest='logfile',
-        help="log to the file named by this option, defaults to not logging")
-    parser.add_option(
-        '--tmp_prefix', default=None, dest='tmp_prefix',
-        help="tmp dir prefix - will default to username-pid if not specified")
-    parser.add_option(
-        '-m', '--mock-option', default=[], action='append',
-        dest='mock_option',
-        help="option to pass directly to mock")
```
praiskup: Except for this, the `mock --chain` should be compatible with `mockchain`, right?   Would you mind if I added small shell wrapper around `mock --chain`?

PR #358, opened 2019-10-01
ignore ./var/log when creating root_cache - fixes #309

### LINKED ISSUE #309: sparse files handling in root_cache
recently I've stumbled upon issue that for some reason (most propably some global policy, the build wasn't updated at that time), when building the root cache, /..../var/log/tallylog is written, which is a sparse file -> apparent size is 62gb in my case.
mock then creates cache tarball, for future uses. so far so good, first init goes well and the environment has ~1gb.
when we use --init again with the cache present, then the environment has 62gb, because the tallylog is filled with zeroes up to the used offset. This is invconvenniet as we have multiple users using it with their own --uniqueext -> we're running out of space right away

I've tested, that if i hack in -S in the tar call (by a script doing tar -S $@), it will then expand the sparse file properly when reusing cache. The appearent size will be 62gb, but df -h goes down by ~1GB.

is there any reason not to add the argument in plugins/root_cache.py? (should i pullrequest that)?

ghormoon:
proposed solution for https://github.com/rpm-software-management/mock/issues/309


xsuchy:
Thank you

PR #364, opened 2019-10-03
search config files in multiple possible paths

### LINKED ISSUE #350: RFE: include config templates from relative path
At this point, mock configs contain a following include

    include('/etc/mock/templates/fedora-rawhide.tpl')

While this works perfectly, it basically disallows having modified mock configs (templates) anywhere outside `/etc/mock`. For example, it is not possible to use mock configs from git repo without installing them. E.g.

    $ mock -r ./mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg --shell

fails with
    
      File "/home/jkadlcik/git/mock/mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg", line 1, in <module>
        include('/etc/mock/templates/fedora-rawhide.tpl')

    mockbuild.exception.ConfigError: Could not find included config file: /etc/mock/templates/fedora-rawhide.tpl

    ERROR: Error in configuration

This particular problem can be fixed with upgrading `mock-core-configs` package, but the main issue is that you can't simply (without installing) test changes, that you make in templates. It would be IMHO much better if instead of including

    '/etc/mock/templates/fedora-rawhide.tpl'

we would include something like

    location_of_the_mock_config + '/templates/fedora-rawhide.tpl'

FrostyX:
Fix #350

The point of this change is to use non-installed template files.

When `include` function is called on a file specified by an absolute
path, nothing changes. However, when a relative path is specified
(which for .tpl files now is), then the file is searched in multiple
possible locations:

1) The same directory where the .cfg config is
2) `templates` directory in the same location where .cfg config is
3) /etc/mock/templates

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1339,6 +1339,9 @@ def check_config(config_opts):
 
 @traceLog()
 def include(config_file, config_opts, is_statement=False):
+    if not os.path.isabs(config_file):
+        config_file = os.path.join(config_opts['config_path'], config_file)
+
     if os.path.exists(config_file):
         if is_statement and config_file in config_opts['config_paths']:
             getLog().warning("Multiple inclusion of %s, skipping" % config_file)
@@ -1461,6 +1464,7 @@ def load_config(config_path, name, uidManager, version, pkg_python_dir):
     config_opts = setup_default_config_opts(gid, version, pkg_python_dir)
 
     # array to save config paths
+    config_opts['config_path'] = config_path
     config_opts['config_paths'] = []
     config_opts['chroot_name'] = name
 
--- a/mock-core-configs/etc/mock/amazonlinux-2-aarch64.cfg
+++ b/mock-core-configs/etc/mock/amazonlinux-2-aarch64.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/amazonlinux-2.tpl')
+include('templates/amazonlinux-2.tpl')
 
 config_opts['root'] = 'amzn-2-aarch64'
 config_opts['target_arch'] = 'aarch64'
--- a/mock-core-configs/etc/mock/amazonlinux-2-x86_64.cfg
+++ b/mock-core-configs/etc/mock/amazonlinux-2-x86_64.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/amazonlinux-2.tpl')
+include('templates/amazonlinux-2.tpl')
 
 config_opts['root'] = 'amzn-2-x86_64'
 config_opts['target_arch'] = 'x86_64'
--- a/mock-core-configs/etc/mock/centos-stream-aarch64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-aarch64.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/centos-stream.tpl')
+include('templates/centos-stream.tpl')
 
 config_opts['root'] = 'centos-stream-aarch64'
 config_opts['target_arch'] = 'aarch64'
--- a/mock-core-configs/etc/mock/centos-stream-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-ppc64le.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/centos-stream.tpl')
+include('templates/centos-stream.tpl')
 
 config_opts['root'] = 'centos-stream-ppc64le'
 config_opts['target_arch'] = 'ppc64le'
--- a/mock-core-configs/etc/mock/centos-stream-x86_64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-x86_64.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/centos-stream.tpl')
+include('templates/centos-stream.tpl')
 
 config_opts['root'] = 'centos-stream-x86_64'
 config_opts['target_arch'] = 'x86_64'
--- a/mock-core-configs/etc/mock/epel-8-aarch64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-aarch64.cfg
@@ -1,5 +1,5 @@
-include('/etc/mock/templates/centos-8.tpl')
-include('/etc/mock/templates/epel-8.tpl')
+include('templates/centos-8.tpl')
+include('templates/epel-8.tpl')
 
 config_opts['root'] = 'epel-8-aarch64'
 config_opts['target_arch'] = 'aarch64'
--- a/mock-core-configs/etc/mock/epel-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/epel-8-ppc64le.cfg
@@ -1,5 +1,5 @@
-include('/etc/mock/templates/centos-8.tpl')
-include('/etc/mock/templates/epel-8.tpl')
+include('templates/centos-8.tpl')
+include('templates/epel-8.tpl')
 
 config_opts['root'] = 'epel-8-ppc64le'
 config_opts['target_arch'] = 'ppc64le'
--- a/mock-core-configs/etc/mock/epel-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-x86_64.cfg
@@ -1,5 +1,5 @@
-include('/etc/mock/templates/centos-8.tpl')
-include('/etc/mock/templates/epel-8.tpl')
+include('templates/centos-8.tpl')
+include('templates/epel-8.tpl')
 
 config_opts['root'] = 'epel-8-x86_64'
 config_opts['target_arch'] = 'x86_64'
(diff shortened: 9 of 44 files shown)


FrostyX:
If you want to be extra safe, we can add a condition and do this thing only for `*.tpl$` files.

mock/py/mockbuild/util.py
```diff
@@ -1339,6 +1339,10 @@ def check_config(config_opts):
 
 @traceLog()
 def include(config_file, config_opts, is_statement=False):
+    if not os.path.isabs(config_file):
+        config_dir = os.path.dirname(config_opts.config_file)
+        config_file = find_config_in_path(config_file, config_dir)
```
praiskup: I'd prefer not to split the argument to `config_file` and `config_dir` ... it is still just the same value filtered by `dirname`.  Why not to have `find_config_in_path(config_file)` - and make the method search through predictible set of directories (e.g. given by config option)?

mock/py/mockbuild/util.py
```diff
@@ -1356,6 +1360,16 @@ def include(config_file, config_opts, is_statement=False):
         raise exception.ConfigError("Could not find included config file: %s" % config_file)
 
 
+@traceLog()
+def find_config_in_path(config_file, config_dir):
+    path = [config_dir, os.path.join(config_dir, "templates"), "/etc/mock/templates/"]
```
praiskup: I'm not able to tell this is the best approach TBH, the fact that we include relatively to the first chosen config file (the main chroot config) brings non-determinism to mock's behavior (the includes happen differently, depending where the config resides).

Usually there's some default search path (notably $PATH but also sys.path for python modules), and you know where you include from - no matter where the input file is, or what is the $PWD.  Well the `sys.path` has flaw that it contains the `pwd`, compared to $PATH (and other *PATH vars), but still.

mock/py/mockbuild/util.py
FrostyX: It is not the same value, actually. The `config_opts.config_file` contains the value, that you passed via `-r` parameter, e.g. `../mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg`. While the `config_file` contains the currently included config/template file, e.g. `fedora-rawhide.tpl` (with this PR, otherwise the template has full etc path)

mock/py/mockbuild/util.py
FrostyX: I get the point and kind of agree. There are basically three use-cases though.

1. Mock configs are in the `/etc/mock` and I just want to use them
2. I am hacking on the configs and want to use those from git directory
3. I have my own personal configs somewhere in my system and I don't necessarily have to use `templates` directory for storing templates.

The current implementation deals with the first two cases with ease and partially solves the third one. If we would go with the proposed solution and set the `path` to some locations, not containing those relative to the specified config, we won't IMHO deal even with the case 2 - configs from git (Only if we would have `pwd` in the path and move ourselves directly to the `~/whatever/mock/mock-core-configs/etc/mock/templates`). We would have to read some environment variable like `MOCKCONFPATH` to allow us explicitly specify the possible config locations. Are you suggesting this, or am I unnecessarily complicating the issue?

mock/py/mockbuild/util.py
praiskup: Aha, ok, this isn't obvious  - good point.

mock/py/mockbuild/util.py
praiskup: > Are you suggesting this

Yes.  Either commandline option, or env var (but env var is sort of standard for *PATH use-case).

xsuchy: I would vote for `include('templates/centos-stream.tpl')` and search only in config dirs (we have several of them). This way we would not need to have `templates` directory hard coded in code and it will solve the third case really nice.

praiskup: I agree with @xsuchy, the `templates` doesn't have to be there.  But for 2 we still need to have a way to adjust the search patch at runtime.

> we have several of them

Hmm, I thought there's only /etc/mock.

xsuchy: `~/.config/mock/` as well. 

```
       -r CONFIG, --root=CONFIG
              Uses  specified  chroot  configuration  as  defined  in  ~/.config/mock/<CONFIG>.cfg or /etc/mock/<CON‐
              FIG>.cfg.  Optionally if CONFIG ends in '.cfg', it is interpreted as full path to config file. If  none
              specified, uses the chroot config linked to by /etc/mock/default.cfg.
```

xsuchy: And technically it can be in https://github.com/rpm-software-management/mock/wiki#generate-custom-config-file

FrostyX:
I may have overcomplicated this issue. I figured out, that I can throw away most of the code from this PR and fix the issue much more easily by utilizing `--configdir` parameter. See the usage and what configs are used now:

    $ mock -r fedora-rawhide-x86_64 --shell
    /etc/mock/site-defaults.cfg
    /etc/mock/fedora-rawhide-x86_64.cfg
    /etc/mock/templates/fedora-rawhide.tpl

    $ mock -r /etc/mock/fedora-rawhide-x86_64.cfg --shell
    /etc/mock/site-defaults.cfg
    /etc/mock/fedora-rawhide-x86_64.cfg
    /etc/mock/templates/fedora-rawhide.tpl

    $ mock -r ../mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg --shell
    /etc/mock/site-defaults.cfg
    /home/jkadlcik/git/mock/mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg
    /etc/mock/templates/fedora-rawhide.tpl

    $ mock -r ../mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg --shell --configdir ../mock-core-configs/etc/mock/
    # <--- No site-defaults.cfg here
    /home/jkadlcik/git/mock/mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg
    ../mock-core-configs/etc/mock/templates/fedora-rawhide.tpl

What do you think? This is good enough for me. Please note, that there is quite mess in the commits, I would like to squash them before merging. Also, in case we stick with the `--configdir`, should I document this use-case somewhere?

nkadel:
I think that if you're going to write your own customized config files, you should be prepared to do the extra configuration to hardcode the path of your own customized .tpl file. Otherwise, get the system config files. Searching unclear alternative directories for config files puts the security of the build system at risk, especially if the config files are *not* in /etc/mock/. I'm still wary of allowing mock to use config files that are not in /.etc/mock/, it would seem to create an inherent danger from people with access to mock. They have a great deal of power anyway, with permissions to build nad manipulate chroot cages., but enabling non-root-owned config files is a safety compromise that has been accepted for a while now.

mock/py/mockbuild/util.py
```diff
@@ -1339,6 +1339,9 @@ def check_config(config_opts):
 
 @traceLog()
 def include(config_file, config_opts, is_statement=False):
+    if config_file.endswith('.tpl') and not os.path.isabs(config_file):
```
praiskup: What will happen with `include('fedora-rawhide-x86_64.cfg')`?  I'm just not sure the `.tpl` condition is correct.

praiskup:
```
$ mock -r ../mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg --shell --configdir ../mock-core-configs/etc/mock/
# <--- No site-defaults.cfg here
/home/jkadlcik/git/mock/mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg
../mock-core-configs/etc/mock/templates/fedora-rawhide.tpl
```

What will happen with `mock -r fedora-rawhide-x86_64.cfg --configdir ../mock-core-configs/etc/mock`?  Btw. why the template is printed with relative path, and the cfg file is absolute?

praiskup:
Otherwise, I like we re-use the `--configdir` option here.  I sort of think we should replace it with `--configpath` _one day_, and allow mock to use more config directories when desired (colon separated list).

mock/py/mockbuild/util.py
xsuchy: I agree that we should allow *.cfg as well

FrostyX:
I've removed the `.tpl` specific condition as requested, it was not necessary. Everything still behaves exactly the same as described in the [#issuecomment-545167072](https://github.com/rpm-software-management/mock/pull/364#issuecomment-545167072).

> What will happen with mock -r fedora-rawhide-x86_64.cfg --configdir ../mock-core-configs/etc/mock

    $ mock -r fedora-rawhide-x86_64.cfg --configdir ../mock-core-configs/etc/mock --shell
    ERROR: Could not find required config file: fedora-rawhide-x86_64.cfg

Do we expect it to work it this way? 

Just for the record, this works

    $ mock -r fedora-rawhide-x86_64 --configdir ../mock-core-configs/etc/mock --shell
    /home/jkadlcik/git/mock/mock-core-configs/etc/mock/fedora-rawhide-x86_64.cfg
    ../mock-core-configs/etc/mock/templates/fedora-rawhide.tpl

mock/py/mockbuild/util.py
FrostyX: Alright, np. Done.

praiskup:
> ERROR: Could not find required config file: fedora-rawhide-x86_64.cfg

That never worked from what I remember, if there`s `cfg` on commandline - it needs to be either absolute path, or relative to `cwd`.  I don't like the behavior, but IMO it is expected.

praiskup:
+1

PR #365, opened 2019-10-03
scrub bootstrap chroot accordingly

### LINKED ISSUE #351: scrub command doesn't remove bootstrap data

`mock -r <chroot-name> --scrub all` doesn't remove bootstrap data

### Output of `rpm -q mock`

    mock-1.4.19-1.fc30.noarch

### Steps to reproduce issue

1. `mock -r fedora-rawhide-x86_64 --bootstrap-chroot --shell`
2. `mock -r fedora-rawhide-x86_64 --scrub all`
3. `ls /var/lib/mock/ /var/cache/mock/ |grep rawhide`

### Any additional notes

    [jkadlcik@localhost mock]$ mock -r fedora-rawhide-x86_64 --scrub all
    ...
    [jkadlcik@localhost mock]$ ls /var/lib/mock/ /var/cache/mock/ |grep rawhide
    fedora-rawhide-x86_64-bootstrap
    fedora-rawhide-x86_64-bootstrap

FrostyX:
Fix #351

The `Commands.scrub` method requires `bootstrap_chroot` object initialized
in order to work with it. I.e. it doesn't scrub a bootstrap chroot unless
the `--bootstrap-chroot` parameter is set. Creating its own object inside
the method would duplicate a lot of existing code and doing it without the
object and its methods and properties is IMHO not a good idea.

Since the workaround for the #351 is

    mock -r <chroot> --bootstrap-chroot --scrub=all

we can just simply fix our parser to pretend like `--bootstrap-chroot`
was set, when using `--scrub`.

### DIFF
--- a/mock/docs/mock.1
+++ b/mock/docs/mock.1
@@ -166,7 +166,7 @@ package and branch can be defined with \fB\-\-scm\-option\fP arguments,
 see site\-defaults.cfg for more information.
 .TP
 \fB\-\-scrub\fR=\fITYPE\fP
-Completely remove the specified chroot or cache dir or all of the chroot and cache.  \fITYPE\fR is one of all, chroot, cache, root\-cache, c\-cache, yum\-cache or dnf\-cache. In fact, dnf\-cache is just alias for yum\-cache, and both remove Dnf and Yum cache.
+Completely remove the specified chroot or cache dir or all of the chroot and cache.  \fITYPE\fR is one of all, chroot, bootstrap, cache, root\-cache, c\-cache, yum\-cache or dnf\-cache. In fact, dnf\-cache is just alias for yum\-cache, and both remove Dnf and Yum cache.
 .TP
 \fB\-\-shell\fP [\fICOMMAND [\fIARGS...]\fR]
 Run the specified command interactively within the chroot (no 'clean' is performed). If no command specified, /bin/sh is run. Note that COMMAND is shell expanded using the shell in chroot. Note that this command does not produce logs.
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -134,6 +134,10 @@ class Commands(object):
                         self.buildroot.root_log.info("scrubbing yum-cache and dnf-cache for %s", self.config_name)
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'yum_cache'), selinux=self.buildroot.selinux)
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'dnf_cache'), selinux=self.buildroot.selinux)
+                    elif scrub == 'bootstrap' and self.bootstrap_buildroot is not None:
+                        self.buildroot.root_log.info("scrubbing bootstrap for %s", self.config_name)
+                        self.bootstrap_buildroot.delete()
+                        util.rmtree(self.bootstrap_buildroot.cachedir, selinux=self.bootstrap_buildroot.selinux)
 
             except IOError as e:
                 getLog().warning("parts of chroot do not exist: %s", e)
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -105,10 +105,12 @@ class Commands(object):
                 self.plugins.call_hooks('clean')
                 if self.bootstrap_buildroot is not None:
                     self.bootstrap_buildroot.plugins.call_hooks('clean')
+
                 for scrub in scrub_opts:
                     self.plugins.call_hooks('scrub', scrub)
                     if self.bootstrap_buildroot is not None:
                         self.bootstrap_buildroot.plugins.call_hooks('scrub', scrub)
+
                     if scrub == 'all':
                         self.buildroot.root_log.info("scrubbing everything for %s", self.config_name)
                         self.buildroot.delete()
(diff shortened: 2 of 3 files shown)


praiskup:
`--scrub` can be called with
`all|chroot|cache|root-cache|c-cache|yum-cache|dnf-cache`.

Usually, when I don't specify --scrub=all, I don't care about bootstrap
chroot (it doesn't make sense to e.g. re-download its dnf-cache for
bootstrap chroot, or to re-install the chroot.  Bootstrap is just
mediator that user shouldn't usually know about, it shouldn't affect the
build results.

I'm not sure about this, but I'd slightly tend to just clean the bootstrap
chroot if user specifies `--scrub=all`.  Even #351 explicitly mentions 
`--scrub=all`, so perhaps that would be enough?

FrostyX:
Good thinking. I've updated the PR.

FrostyX:
Although if you look into the `Commands.scrub` method, it explicitly defines `boostrap_buildroot` deletion for all possible options.

praiskup:
Ah, good point.  This is not only about explicit `--bootstrap-chroot` option, but also about config options in /etc/mock.  So shouldn't we rather add the check for `all` into a different place?

FrostyX:
> but also about config options in /etc/mock

But you don't set _scrubbing_ in `/etc/mock`. You can set `config_opts['use_bootstrap_container']` there, but it shouldn't matter, right?

praiskup:
Btw., similar change was dda9cd7bdb33a6a87b0788a3.

FrostyX:
I moved it out of the parser. Do you like this version better?

praiskup:
Well, there's still `mock -r <chroot> --scrub=<anything-else-than-all> --bootstrap-container`.  I think this shouldn't apply to bootstrap?  What do you think?

FrostyX:
> Well, there's still mock -r <chroot> --scrub=<anything-else-than-all> --bootstrap-container. I think this shouldn't apply to bootstrap? What do you think?

So you basically want to set `config_opts['use_bootstrap_container']` to `False` when `--scrub=<anything-else-than-all>` even though you explicitly use `--bootstrap-container`?

praiskup:
I think that with `scrub` on (any param), we should set `config_opts['use_bootstrap_container'] = True` (hack to get bootstrap initialized), but we should do more fine-grained cleanup in `scrub()` method depending on what the `--scrub=ARG` argument is.

We could have also `--scrub=bootstrap` which would only scrub the bootstrap, for example.

FrostyX:
I've updated the PR and now it:

- it doesn't care whether you pass `--no-bootstrap-chroot` or set `config_opts['use_bootstrap_container'] = False`. When you use `--scrub` parameter with any value, it initializes the `bootstrap_buildroot` object.
- it adds a `--scrub=bootstrap` option to remove just bootstrap chroot
- it removes bootstrap chroot data only when `bootstrap` or `all` is passed. Otherwise, it leaves it

mock/py/mockbuild/backend.py
```diff
@@ -105,10 +105,12 @@ def scrub(self, scrub_opts):
                 self.plugins.call_hooks('clean')
                 if self.bootstrap_buildroot is not None:
                     self.bootstrap_buildroot.plugins.call_hooks('clean')
+
                 for scrub in scrub_opts:
                     self.plugins.call_hooks('scrub', scrub)
                     if self.bootstrap_buildroot is not None:
                         self.bootstrap_buildroot.plugins.call_hooks('scrub', scrub)
+
```
FrostyX: This whitespace change was intentional. Previously it was squashed so much together, that it was had to read

mock/py/mockbuild/backend.py
```diff
@@ -119,34 +121,23 @@ def scrub(self, scrub_opts):
...
                     elif scrub in ['yum-cache', 'dnf-cache']:
                         self.buildroot.root_log.info("scrubbing yum-cache and dnf-cache for %s", self.config_name)
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'yum_cache'), selinux=self.buildroot.selinux)
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'dnf_cache'), selinux=self.buildroot.selinux)
-                        if self.bootstrap_buildroot is not None:
-                            util.rmtree(os.path.join(self.bootstrap_buildroot.cachedir, 'yum_cache'),
-                                        selinux=self.bootstrap_buildroot.selinux)
-                            util.rmtree(os.path.join(self.bootstrap_buildroot.cachedir, 'dnf_cache'),
-                                        selinux=self.bootstrap_buildroot.selinux)
+                    elif scrub == 'bootstrap' and self.bootstrap_buildroot is not None:
```
praiskup: Here should be something in `self.buildroot.root_log.info('...')`.

mock/docs/mock.1
```diff
@@ -166,7 +166,7 @@ package and branch can be defined with \fB\-\-scm\-option\fP arguments,
 see site\-defaults.cfg for more information.
 .TP
 \fB\-\-scrub\fR=\fITYPE\fP
-Completely remove the specified chroot or cache dir or all of the chroot and cache.  \fITYPE\fR is one of all, chroot, cache, root\-cache, c\-cache, yum\-cache or dnf\-cache. In fact, dnf\-cache is just alias for yum\-cache, and both remove Dnf and Yum cache.
+Completely remove the specified chroot or cache dir or all of the chroot and cache.  \fITYPE\fR is one of all, chroot, bootstrap, cache, root\-cache, c\-cache, yum\-cache or dnf\-cache. In fact, dnf\-cache is just alias for yum\-cache, and both remove Dnf and Yum cache.
```
praiskup: I'd put the `bootstrap` as the last one, but this is detail.

praiskup:
Thanks, this looks nice!

praiskup:
I'd somewhat prefer to have this as one commit, IMO at least the first two commits aren't fine without each other.

mock/py/mockbuild/backend.py
FrostyX: I totally forgot on that. Fixed. Thank you!

FrostyX:
> I'd somewhat prefer to have this as one commit, IMO at least the first two commits aren't fine without each other.

I've squashed those two together

praiskup:
+1

mock/py/mockbuild/backend.py
```diff
@@ -119,34 +121,24 @@ def scrub(self, scrub_opts):
...
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'ccache'), selinux=self.buildroot.selinux)
-                        if self.bootstrap_buildroot is not None:
-                            util.rmtree(os.path.join(self.bootstrap_buildroot.cachedir, 'ccache'),
-                                        selinux=self.bootstrap_buildroot.selinux)
                     elif scrub == 'root-cache':
                         self.buildroot.root_log.info("scrubbing root-cache for %s", self.config_name)
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'root_cache'), selinux=self.buildroot.selinux)
-                        if self.bootstrap_buildroot is not None:
-                            util.rmtree(os.path.join(self.bootstrap_buildroot.cachedir, 'root_cache'),
-                                        selinux=self.bootstrap_buildroot.selinux)
```
xsuchy: I do not get why these lines are being removed for root-cache, c-cache, cache, and yum-cache and dnf-cache?

praiskup: That's what we think is actually desired;  people doing `--scrub=root-cache` unlikely want to also clean `root-cache` of bootstrap chroot (they shouldn't mind, or even shouldn't basically know there's bootstrap chroot once it is default).

xsuchy: But if they know and they want to delete it, then there will be no way

praiskup: Right, they'd have to do `--scrub=bootstrap`, but yes ... most valid would be to have `--scrub=bootstrap-root-cache`.

xsuchy: IIRC, the `--scrub=bootstrap` will not clean bootstrap root cache

praiskup:
On Tuesday, October 15, 2019 10:12:34 AM CEST Miroslav Suchý wrote:
> IIRC, the `--scrub=bootstrap` will not clean bootstrap root cache

Sure, but not _only_ that.  It will clean everything.  The question what
granularity we want WRT bootstrap.

xsuchy:
@praiskup if I am reading the code correctly https://github.com/rpm-software-management/mock/blob/7c816e7ed7092d10e7be1b8e58adddde008c9e4f/mock/py/mockbuild/buildroot.py#L681 then `bootstrap_buildroot.delete()` will not delete e.g., `/var/cache/mock/epel-7-x86_64-bootstrap/root_cache`

praiskup:
Yes, `chroot.delete()` is equal to `--clean` option, and it doesn't clean
cache dir itself.  That's why `--scrub=all` calls `chroot.delete()` **and**
`util.rmtree(self.buildroot.cachedir,...)`.  The same happens with the new
`--scrub=bootstrap` (either '--scrub=all' == full cleanup, or nothing).

I think we don't need more granularity for bootstrap cleanup, but you may
have different POV.

mock/py/mockbuild/backend.py
```diff
@@ -119,34 +121,24 @@ def scrub(self, scrub_opts):
...
                         util.rmtree(os.path.join(self.buildroot.cachedir, 'dnf_cache'), selinux=self.buildroot.selinux)
-                        if self.bootstrap_buildroot is not None:
-                            util.rmtree(os.path.join(self.bootstrap_buildroot.cachedir, 'yum_cache'),
-                                        selinux=self.bootstrap_buildroot.selinux)
-                            util.rmtree(os.path.join(self.bootstrap_buildroot.cachedir, 'dnf_cache'),
-                                        selinux=self.bootstrap_buildroot.selinux)
+                    elif scrub == 'bootstrap' and self.bootstrap_buildroot is not None:
+                        self.buildroot.root_log.info("scrubbing bootstrap for %s", self.config_name)
+                        self.bootstrap_buildroot.delete()
+                        util.rmtree(self.bootstrap_buildroot.cachedir, selinux=self.bootstrap_buildroot.selinux)
```
praiskup: ... here is the bootstrap cache cleanup.

xsuchy: I was blind.

PR #368, opened 2019-10-08
Fix 349

praiskup:


praiskup:
For myself: https://copr-be.cloud.fedoraproject.org/results/cherts/zabbix_34x_next/epel-6-i386/01051380-zabbix/builder-live.log.gz
I suspect that copr builder is broken after this problem, but backend fails to detect it.

PR #370, opened 2019-10-08
mock: detect that forcearch can not work

praiskup:
On Red Hat systems, the qemu-user-static package provides
/usr/bin/qemu-x86_64-static on all architectures.  If that file is
missing => the whole package is missing => and the feature can not work.

Also, we don't support --forcearch through --bootstrap-chroot ATM.


xsuchy:
LGTM

PR #372, opened 2019-10-10
bootstrap: fix --install and --installdeps

praiskup:
To fix --install* options for bootstrap transparently, all the files
given on command-line need to be bind-mounted into bootstrap chroot so
they can be processed _inside_.

Since the files on command-line may be relative to cwd and we don't want
to create arbitrary directory structures within bootstrap buildroot, we
simply bind-mount the files into "home" directory.  The obvious
consequence is that multiple files with the same basename could clash
inside in the homedir, so we rather raise an error as soon as possible
instead of overmounting the same path.

Fixes: rhbz#1447627

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -17,7 +17,8 @@ import uuid
 from . import mounts
 from . import uid
 from . import util
-from .exception import BuildRootLocked, Error, ResultDirNotAccessible, RootError
+from .exception import (BuildRootLocked, Error, ResultDirNotAccessible,
+                        RootError, BadCmdline)
 from .package_manager import package_manager
 from .trace_decorator import getLog, traceLog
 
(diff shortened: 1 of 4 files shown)


mock/py/mockbuild/buildroot.py
```diff
@@ -639,6 +642,41 @@ def finalize(self):
...
+            raise BadCmdline("File '{0}' can not be bind-mounted to "
+                             "bootstrap chroot twice".format(basename))
+        self._homedir_bindmounts[basename] = 1
+
+        host_filename = os.path.abspath(filename)
+        chroot_filename = os.path.join(
+            bootstrap.homedir, basename,
+        )
+        bind_path = bootstrap.make_chroot_path(chroot_filename)
+        bootstrap.mounts.add(mounts.BindMountPoint(
```
praiskup: ah, this needs to be an user mount

xsuchy:
+1

Issue #376, opened 2019-10-14
'success' file is owned by root


praiskup:
Should be owned by regular user, as any other file.

PR #378, opened 2019-10-14
chain: success/fail files shouldn't be root owned

praiskup:
Fixes: #376
Merges: #378


praiskup:
// --ignore-space-change is useful here

PR #385, opened 2019-10-25
Add configs for epel8-playground

mmathesius:
The `epel-playground-8` template already exists. This PR adds the per-architecture configs so they can easily be used.


xsuchy:
All current configs follow the patter $name-$version-$arch. While it is not technically required, some software build on top of mock assumes this pattern. Can you rename it to epelplayground-8-$arch?

mmathesius:
@xsuchy Thanks for the feedback. I rebased upon your latest commit and renamed as you suggested. To stick with the pattern, I also renamed `templates/epel-playground-8.tpl` to `templates/epelplayground-8.tpl`--but I'm happy to revert that if you like.

xsuchy:
Thank you.

PR #389, opened 2019-10-29
Add 'kernel-rpm-macros' to centos-8.tpl

tonyhutter:
This adds `kernel-rpm-macros` to the Centos 8 build root.  It is needed to provide the `%kernel_module_package_buildreqs` macro when building source RPMs for kernel modules.  This fix allows us to build the ZFS kmod source RPMs under Centos 8.

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-8.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-8.tpl
@@ -1,4 +1,4 @@
-config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep kernel-rpm-macros'
 config_opts['dist'] = 'el8'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'


xsuchy:
Centos-8 config should mimic CentOS Koji settings as much as possible. AFAIK the kernel-rpm-macros is not part of the set-up. I would like to see a discussion on centos or buildsys mailing list and general agreement that this is good.

Conan-Kudo:
@tonyhutter Wouldn't it make more sense to just add it as a BR to the kmod-zfs spec?

xsuchy:
@Conan-Kudo it depends on the spec file. Sometimes it is needed before the BuildRequires is being evaluated (e.g SCL). Thou I am not sure this is the case. Sharing the spec file will definitelly help.

tonyhutter:
@Conan-Kudo I tried adding the BuildRequires but it didn't work.  The problem is that the BuildRequires aren't pulled in when you create the SRPM, but rather when you build the SRPM to a binary.  

```
# The macros defined by these files are for things that need to be defined
# at srpm creation time when it is not feasible to require the base packages
# that would otherwise be providing the macros. other language/arch specific
# macros should not be defined here but instead in the base packages that can
# be pulled in at rpm build time, this is specific for srpm creation.
```
https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/master/f/redhat-rpm-config.spec#_26

The `%kernel_module_package_buildreqs` macro I need to create the source RPM is defined in `/usr/lib/rpm/macros.d/macros.kmp`, which is part of the `kernel-rpm-macros` package, which is a subpackage of `redhat-rpm-configs`.  Basically, `redhat-rpm-configs` defines RPM macros for userspace,  and `kernel-rpm-macros` defines them for kernel modules.  `redhat-rpm-configs` is included by mock in the build root, `kernel-rpm-macros` is not.  I'm talking specifically about Centos 8 here:

https://github.com/rpm-software-management/mock/blob/devel/mock-core-configs/etc/mock/templates/centos-8.tpl#L1

My guess is that is just a simple oversight that one is included and the other is not, since the vast majority of packages built are for userspace, not kernel modules.

@xsuchy our spec file is here:
https://github.com/zfsonlinux/zfs/blob/master/rpm/redhat/zfs-kmod.spec.in

I'll bug the koji people about including it, as it sounds like mock just mirrors whatever build environment they have.

Conan-Kudo:
This would probably be required for the Mock configs for Fedora as well (since F21)...

tonyhutter:
Fedora uses a different spec file that doesn't use `%kernel_module_package_buildreqs`

xsuchy:
@pmatilai I see in the changelog that kernel-rpm-macros has been split from redhat-rpm-configs in 2014. Can you explain to us, why it happened and why it is not in default buildroot? I think I miss something obvious.

pmatilai:
There are several reasons.

The kernel-rpm-macros content makes little sense in Fedora, and those macros and helpers are maintained by a different group of people compared to the rest of redhat-rpm-config, this is a constant pain on RHEL side. The sub-package was split off with the intention of further moving that to a separate source level component to allow for more natural maintainership over the stuff, only the move to a separate component never took place.

It has no business being in the default buildroot as it modifies global %__find_provides and %__find_requires. It should instead be updated to use modern ways of dependency generation.

xsuchy:
I had a discussion with Eugene Syromiatnikov from the kernel team. Based on this discussion I believe this package should not be included in minimal buildroot. There are several options:
1) Either redhat-rpm-config package should require kernel-rpm-macros, but that would require bringing kernel-rpm-macros to up to date.
2) or you can use `config_opts['chroot_additional_packages'] = ''` in your ~/.config/mock.cfg. This is even available in Copr webUI.

I will leave this open for a while, but likely I will close this in few months unless something changes the situation.

Issue #392, opened 2019-11-04
yum/dnf cache for mock --bootstrap-chroot is broken


RomanValov:
### Short description of the problem
In contrast to regular mode, when using `--bootstrap-chroot` option yum/dnf cache bind mount is not used, resulting yum/dnf cache re-downloaded on chroot cleanup.

### Output of `rpm -q mock`
mock-1.4.20-1.el7.noarch

### Steps to reproduce issue

1.
``` 
mock --bootstrap-chroot --root epel-6-x86_64 --init
```
2. 
```
mock --bootstrap-chroot --root epel-6-x86_64 --update
```
3.
```
mock --bootstrap-chroot --root epel-6-x86_64 --clean
```
4.
```
mock --bootstrap-chroot --root epel-6-x86_64 --update
```

Expected results: yum/dnf caches obtained on `--init` stage are reused on `--update` stages.
Actual results: yum/dnf caches are redownloaded on `--update` stages and kept inside `rootdir`.

fkooman:
This is also a problem with `use_bootstrap_container`...

praiskup:
Thanks for the report!

> In contrast to regular mode, 

In non-bootstrap scenario, your **host** cache directories are used actually because you are using dnf/yum from host -- and that's why everything works correctly _for free_.

In bootstrap scenario, this isn't resolved yet.  Doing `--clean` actually removes all the caches,
as well as `--init`.

praiskup:
Let's underline the purpose of yum_cache plugin first (that was news for me).

The yum/dnf `--installroot` doesn't use **host's** /var/cache/ directory, but it
creates a new directory below --installroot directory (given by `cachedir`
option, which is different for dnf/yum by default).  That directory would be
cleaned with each `--clean` or `--scrub=chroot` actions, which would cause repo
re-download.  So ... the plugin rather creates additional directory under
/var/cache/mock **on host**, and pre-mounts it into the chroot before executing
the pkg manager action with `--installroot`.

The thing is that nowadays it is hard for yum_cache plugin to pick the
correct `/var/cache/{dnf,yum}` variant:

1. The "target" buildroot has "package_manager" configuration, which is
   "expected" package manager that would be installed by default into buildroot.

2. It's unclear whether /bin/yum is dnf or yum (on newer Fedora boxes, /bin/yum
   is actually symlink to dnf).

3. Bootstrap chroot, when used, has package manager installed according to
[...]

PR #393, opened 2019-11-04
bootstrap: fix yum/dnf cache bind mount

RomanValov:
Proposed fix for #392

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -161,7 +161,7 @@ class Commands(object):
                 util.mkdirIfAbsent(self.buildroot.make_chroot_path())
                 self.bootstrap_buildroot.initialize(**kwargs)
                 self.buildroot.mounts.managed_mounts.append(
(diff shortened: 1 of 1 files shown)


praiskup:
Thank you for the patch, and for the issue report!

Unfortunately, this isn't correct fix.  The yum cache plugin isn't
supposed to be used for this purpose (its real purpose is to re-use yum
caches for users working _manually_ with yum/dnf inside chroot, e.g.  with
`mock --shell`).  The plugin should actually be disabled for bootstrap
chroot since there it has no effect, and we should handle the cache
directory explicitly in bootstrap code.  I'll take a look at the patch.

praiskup:
I was wrong in previous comment, the problem is that host's pkg manager is different than the bootstrap chroot package manager.  Let's move the discussion to the original issue.

praiskup:
More info;  the recursive bind mount is really needed but it's better to make that mount tree hidden from host.

Issue #396, opened 2019-11-05
missing metadata_expire=0 in epel chroots


praiskup:
This is originally intentional:
https://bugzilla.redhat.com/show_bug.cgi?id=1230508

But I think we should add it everywhere.  E.g. in copr mock behaves differently for epel, and it likely triggers weird issues (https://bugzilla.redhat.com/show_bug.cgi?id=1768955).

PR #399, opened 2019-11-07
 Adds missing metadata_epire=0 in epel chroots

hrnciar:
Fix: #396


xsuchy:
I believe that this was supposed to be added to mock-core-configs/etc/mock/templates/centos-8.tpl only.

praiskup:
@hrnciar can you please test that everything works fine even after this change, for all changed chroots (and with yum as well)?  E.g. for each affected chroot run `mock --bootstrap-chroot -r $CHROOT --install gcc`?

praiskup:
+1

Issue #395, opened 2019-11-05
bootstrap chroot: opensuse tumbleweed  doesn't have shadow-utils


praiskup:
After 8f8863d6a, we install shadow-utils into every bootstrap chroot, but opensuse tumbleweed doesn't have that package.  So builds with --bootstrap fail.

xsuchy:
let's overwrite it in opensuse config then.

Conan-Kudo:
@praiskup The package name is `shadow` on openSUSE.

Conan-Kudo:
@praiskup You also could request `/sbin/useradd`, which works on RH/Fedora and SUSE systems.

praiskup:
It seems there's no need to install it to bootstrap chroot, though.  So we removed it from all configs.

PR #400, opened 2019-11-07
bootstrap: drop shadow-utils from dnf_install_command for opensuse chroots

hrnciar:
Fix: #395

### DIFF
--- a/mock-core-configs/etc/mock/opensuse-tumbleweed-aarch64.cfg
+++ b/mock-core-configs/etc/mock/opensuse-tumbleweed-aarch64.cfg
@@ -8,6 +8,7 @@ config_opts['useradd'] = '/usr/sbin/useradd -o -m -u %(uid)s -g %(gid)s -d %(hom
 config_opts['releasever'] = '0'
 config_opts['macros']['%dist'] = '.suse.tw%(sh -c ". /etc/os-release; echo \$VERSION_ID")'
 config_opts['package_manager'] = 'dnf'
+config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core distribution-gpg-keys'
 
 config_opts['yum.conf'] = """
 [main]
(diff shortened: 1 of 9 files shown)


mock-core-configs/etc/mock/opensuse-tumbleweed-aarch64.cfg
```diff
@@ -8,6 +8,7 @@ config_opts['useradd'] = '/usr/sbin/useradd -o -m -u %(uid)s -g %(gid)s -d %(hom
 config_opts['releasever'] = '0'
 config_opts['macros']['%dist'] = '.suse.tw%(sh -c ". /etc/os-release; echo \$VERSION_ID")'
 config_opts['package_manager'] = 'dnf'
+config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core shadow-utils distribution-gpg-keys'
```
praiskup: The problem of tumbleweed is that there's no `shadow-utils` package, there's `shadow` package (which is installed automatically, even when it is not in `dnf_install_command`).

hrnciar: Oh, got it now. So solution should be that I just simply remove shadow-utils from this `dnf_install_command`and it will take precedence over `dnf_install_command` in `util.py`. So it won't fail on no package found with this name. Am I right?  

```suggestion
config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core distribution-gpg-keys'
```

praiskup: That should be good enough.

praiskup:
Meh, by `mock -r opensuse-leap-15.0-x86_64 --install shadow-utils` failure I now realized that this is needed for all opensuse chroots, not only tubleweed.

PR #401, opened 2019-11-08
Changing cfg files for custom 1 to use tpl file

hrnciar:


praiskup:
To make `git blame` easier, it's better to move one (the oldest ideally) config file into the template first in first commit, and then do the changes in second one.

praiskup:
(try `git blame -C mock-core-configs/etc/mock/templates/fedora-rawhide.tpl`)

praiskup:
Ok, was wrong, I meant `git blame mock-core-configs/etc/mock/templates/fedora-30.tpl`.  The point is to make blame working without `-C`.

Nevermind ... the previous PR with epel-6 is already merged (requres `blame -C`), and this is incorrect as well; please either first do 'git mv ...' in one commit, and the rest, or just squash the commits to not over-complicate this.

praiskup:
> The point is to make blame working without -C.

Not only that of course.  It is for example pretty hard to follow the patch as well: You need to check that all the removed lines exactly match the added lines in different file manually.  OTOH commit with `git mv` says that the file was moved without changes...

praiskup:
+1, thanks

PR #402, opened 2019-11-08
bootstrap: create gpg key directory before copying there

praiskup:
```
$ mock -r epel-6-x86_64 --scrub=all
$ mock -r epel-6-x86_64 --bootstrap-chroot --shell
...
ERROR: Command failed:
 # cp -a /usr/share/distribution-gpg-keys /var/lib/mock/epel-6-x86_64-bootstrap/root/usr/share
```


xsuchy:
BTW https://docs.python.org/3.9/distutils/apiref.html#distutils.dir_util.copy_tree would be probably better.

PR #409, opened 2019-11-11
configs: drop cost=2000 from fedora-31+-i386

praiskup:
Those configs don't have more repos, so the cost override doesn't make
sense.


xsuchy:
Well, technically speaking - they have noarch packages which are also presented in x86_64 repos.

praiskup:
Hm, can you elaborate on this?  The build repo may contain both packages, but cost shouldn't be used in this case..

Issue #380, opened 2019-10-17
Missing epel-8 repos in bootstrap chroot 


mkutlak:
### Short description of the problem
Mock doesn't copy-in epel repos into bootstrap chroot when the `use_bootstrap_container` is used.

The epel repos are needed to install `distribution-gpg-keys`.

### Output of `rpm -q mock`

`mock-1.4.20-1.el8.noarch`

### Steps to reproduce issue
```
config_opts['use_bootstrap_container'] = True
config_opts['use_bootstrap_image'] = True
```

```
mock --init -r epel-8-x86_64
INFO: mock.py version 1.4.20 starting (python version = 3.6.8)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start: clean chroot
Finish: clean chroot
Start(bootstrap): chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: Using bootstrap image: centos:8
INFO: Pulling image: centos:8
Trying to pull registry.redhat.io/centos:8...Failed
Trying to pull quay.io/centos:8...Failed
Trying to pull docker.io/centos:8...Getting image source signatures
Copying blob sha256:729ec3a6ada3a6d26faca9b4779a037231f1762f759ef34c08bdd61bf52cd704
Copying config sha256:0f3e07c0138fbe05abcb7a9cc7d63d9bd4c980c3f61fea5efa32e7c4217ef4da
Writing manifest to image destination
Storing signatures
0f3e07c0138fbe05abcb7a9cc7d63d9bd4c980c3f61fea5efa32e7c4217ef4da
CentOS-8 - AppStream                            2.2 MB/s | 6.0 MB     00:02    
CentOS-8 - Base                                 2.1 MB/s | 7.9 MB     00:03    
CentOS-8 - Extras                               1.9 kB/s | 2.1 kB     00:01    
Last metadata expiration check: 0:00:01 ago on Thu Oct 17 13:13:45 2019.
Package dnf-4.0.9.2-5.el8.noarch is already installed.
Package shadow-utils-2:4.6-7.el8.x86_64 is already installed.
No match for argument: distribution-gpg-keys
Error: Unable to find a match
exit status 1
ERROR: Command failed: 
[...]

PR #413, opened 2019-11-13
bootstrap: don't install distribution-gpg-keys pkg

praiskup:
We copy the gpg keys from host nowadays, and after 63ab868920e8c0 for
all bootstrap chroots (not only in bootstrap image case).

So make the '<PKGMANAGER>_install_command' config options simpler, and
lower repo-prerequisites (no need to have distribution-gpg-keys in
repository)

Complements: 63ab868920e8c0
Fixes: #380, #308


praiskup:
Ah, RHEL configs need fix as well.

praiskup:
Fixed.

PR #414, opened 2019-11-13
bootstrap: don't install shadow-utils

praiskup:
We don't actually need the mockbuild user in bootstrap, so stop doing
useradd (etc.) and drop 'shadow-utils' from *_install_command config
option.  This speeds the installroot transaction a bit, but also
the opensuse chroots don't have this package, so we needed to override
the configs before.

Relates: #395

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -128,8 +128,8 @@
 
 # when 'use_bootstrap_container' is True, these commands are used to build
 # the minimal chroot for the respective package manager
-# config_opts['yum_install_command'] = 'install yum yum-utils shadow-utils'
-# config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core shadow-utils'
+# config_opts['yum_install_command'] = 'install yum yum-utils'
+# config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core'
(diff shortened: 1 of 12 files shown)


praiskup:
Hm, `--shell` doesn't work.

praiskup:
Shell works now.

mock/etc/mock/site-defaults.cfg
```diff
@@ -128,8 +128,8 @@
 
 # when 'use_bootstrap_container' is True, these commands are used to build
 # the minimal chroot for the respective package manager
-# config_opts['yum_install_command'] = 'install yum yum-utils shadow-utils distribution-gpg-keys'
-# config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core shadow-utils distribution-gpg-keys'
+# config_opts['yum_install_command'] = 'install yum yum-utils distribution-gpg-keys'
+# config_opts['dnf_install_command'] = 'install dnf dnf-plugins-core distribution-gpg-keys'
```
xsuchy: But if we **use** bootstrap, then we need it. Right?

praiskup: I don't think so, that's why I submitted this PR...  but I may be wrong.  Any idea why the user would be needed in bootstrap chroot?

PR #415, opened 2019-11-13
mock: explicitly require distribution-gpg-keys

praiskup:
There's transitive dependency through mock-core-configs, but it's better
to be explicit because --bootstrap-chroot now depends on this package
(mock-core-configs OTOH has the requires because it wants to enforce
proper version).


praiskup:
With @xsuchy we decided that it would be better to not add additional requirement,
because any additional Requires: complicates a bit new architecture bootstrapping.
So if anything, we could add a new runtime check with proper error.

PR #423, opened 2019-11-20
Initialize config_paths in funtion util.setup_default_config_opts 

sergiomb2:
If we want call directly function util.update_config_from_file, we need set
config_paths, so config_paths should set in setup_default_config_opts,
is better than in function load_config


praiskup:
https://github.com/rpm-software-management/mock/pull/423/commits/66ab067933d2889098f9a522d1592f47ee65e5f5

Only cosmetics, could you fix "header" in commit message to something shorter, like
`initialize config_paths in funtion util.setup_default_config_opts`

Otherwise looks good, thanks.

sergiomb2:
done

praiskup:
+1

xsuchy:
Thank you.

PR #334, opened 2019-09-15
mount /dev/loop-control and /dev/loopX into systemd-nspawn container

FrostyX:
Fix #298

Issue #298, opened 2019-07-24
/dev/loop-control and /dev/loop? not created in nspawn chroot


sharkcz:
### Short description of the problem
/dev/loop-control and /dev/loop? device nodes are not created in the nspawn chroot, but they are in the classic chroot.

### Output of `rpm -q mock`
mock-1.4.16-2.fc30.noarch

### Steps to reproduce issue

1. mock -r fedora-rawhide-s390x --init
2. ll /dev/loop*

Do not forget to mention full commandline with the mock command you executed.

### Any additional notes

Consider adding output of `mock --debug-config` this can help developers to reproduce the issue.

praiskup:
Can you please provide an example package which needs this feature?  I'd like to test this.

praiskup:
@sharkcz ^

sharkcz:
@praiskup, IIRC it came from https://bugzilla.redhat.com/show_bug.cgi?id=1707389#c17 when trying to run lorax from a rawhide chroot (nspawn vs classic). Lorax uses loop devices internally for manipulating with the installer image.

PR #430, opened 2019-11-28
mock: loop devices in nspawn

praiskup:
We would ideally --bind mount
    /var/lib/mock/fedora-rawhide-x86_64/root/dev/loop-control
into
    /dev/loop-control

But, there are currently two issues with nspawn:
- the src path is sub-directory of 'systemd-nspawn -D', and thus
  systemd-nspawn thinks the file doesn't exist (I suppose it overmounts
  the /dev subidrectory before trying to do the bind-mount)
- there's the default DeviceAllow=/dev rule

Ideas?

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -88,6 +88,7 @@ class Buildroot(object):
         self.final_rpm_list = None
 
         self._homedir_bindmounts = {}
+        self._setup_nspawn_loop_devices()
 
     @traceLog()
     def make_chroot_path(self, *paths):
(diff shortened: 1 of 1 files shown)


praiskup:
> here's the default DeviceAllow=/dev rule

I tried to create directory under /var/cache/mock e.g., and it failed anyways:
```
Failed to register machine: DeviceAllow= requires device node or pattern
```

praiskup:
@msekletar any idea how I could do `--bind=/var/lib/mock/fedora-rawhide-x86_64/root/dev/loop-control:/dev/loop-control`?

praiskup:
Btw., this tries to solve #298, differently than #334.

mock/py/mockbuild/buildroot.py
```diff
@@ -538,7 +538,14 @@ def _setup_devices(self):
             self.root_log.debug("kernel version == %s", kver)
             for i in devFiles:
                 src_path = "/" + i[2]
-                chroot_path = self.make_chroot_path(i[2])
+
+                if util.USE_NSPAWN:
+                    chroot_path = os.path.join('/dev/mock/',
+                                               self.config['root'],
+                                               i[2])
```
praiskup: WARN WARN! This is **the hack**, we shouldn't maintain another set of directories under `/dev/mock`.  But since `/dev/mock/...` "startswith" `/dev`, systemd-nspawn doesn't object...

msekletar:
I am a bit confused. Shouldn't this be actually the other way around, i.e. `/dev/loop-control` bind mount into `/var/lib/mock/fedora-rawhide-x86_64/root/dev/loop-control` ?

praiskup:
I think that we want to have separated set of loop devices bind-mounted into chroot, not re-using (potentially already used) device files from host.

msekletar:
In that case, I don't understand why you want to bind mount `loop-control`. By bind mounting loop-control you also provide the container a way to manage (add, remove) loop devices on the host.

praiskup:
> By bind mounting loop-control you also provide the container a way to manage
> (add, remove) loop devices on the host.

Indeed, I missed that this is not namespaced, how it could with the
bindmount.  Thanks!  And obviously I neither know how this is supposed to
be used by rpm-build (or what people expect from this feature).

Perhaps this is not about isolation, but only about allowing users to
control loop devices on host.  Which by itself looks like something we
ideally shouldn't allow by default.

praiskup:
The question is whether it makes sense to behave differently for `--old-chroot`
or whether it is better to create `/dev/loop*` files also on host.

Issue #437, opened 2019-12-10
bootstrap image is broken for `yum` based chroots


praiskup:
When playing with #436, I noticed that we don't execute `yum` in container to get yum-utils installed:
```
Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
ERROR: Command failed: 
 # podman exec d16f4857a3dff4229bd2d06f607893560609622cf771bc7c2de73d2e81481ef6 /usr/bin/dnf -y install dnf dnf-plugins-core subscription-manager
```

PR #438, opened 2019-12-10
mock: don't use buildroot.pkg_manager in podman

praiskup:
That attribute is used for 'host -> bootstrap' installation, not for
'bootstrap -> bootstrap'.

Fixes: #437


praiskup:
I had a chance to test EPEL7 again few minutes ago....  and it is needed to allow
builds for Fedora on EL7 as well (before this patch, mock installs yum into the
from-container-image fedora bootstrap chroot, which is wrong).

praiskup:
If anyone wants, can be tried with `mock-1.5.0-1.git.97.da7be97.el7.noarch` from praiskup/mock-fixes copr.

PR #442, opened 2019-12-17
don't override defaults for USE_NSPAWN

praiskup:
So the configuration is respected, and the obsoleted options
--old-chroot and --new-chroot still work.


xsuchy:
Nice catch

PR #443, opened 2020-01-06
mock: release allocated FDs in mock --chain

jcajka:
Currently mock --chain allocates 3 FDs per build and doesn't release them, making it able to run build ~300 times. This releases them prior to the next build being executed, keeping FD use ~constant.


xsuchy:
LGTM, Thank you.

PR #449, opened 2020-01-13
mock-core-configs: fix epel-6.tpl config bug

praiskup:
The indent broke dnf, and 'basearch' config option doesn't exist.

Complements: 564552e441afca0b6dcca651ff7f868421ccbafd


praiskup:
The ppc64 version still fails with:
Error: Failed to download metadata for repo 'base': Cannot prepare internal mirrorlist: No URLs in mirrorlist

Issue #432, opened 2019-11-29
new systemd-nspawn doesn't bypass stdin as pipe to container 


praiskup:
This doesn't work on F31+, while it worked fine on F30 (with systemd
systemd-241-12.git323cdf4.fc30.x86_64).

```
$ echo input | mock --shell 'cat > /file'
... hangs indefinitely, since there's no EOF on stdin ...
```

Systemd on 'F31' has `systemd-nspawn --console=pipe` for that.  But we
don't use that option now, and the feature is documented with security
"warning".  Basically, not everyone wants this....  but e.g. it breaks
[copr rpmbuild](https://pagure.io/copr/copr/issue/1123).

So we probably need a new mode, or some toggle option.

bryteise:
I ended up patching mock to always pass the --console=pipe option via the default config_opts. I don't particularly view this as a large risk but our build infrastructure isn't public compared to fedora.

praiskup:
... not only --shell, per
https://github.com/rpm-software-management/rpm/issues/978.  So we could
perhaps "risk" that, and do:

```patch
diff --git a/mock/py/mockbuild/util.py b/mock/py/mockbuild/util.py
index f7972ea4..dc3bdd56 100644
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -940,7 +940,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['online'] = True
     config_opts['use_nspawn'] = True
     config_opts['rpmbuild_networking'] = False
-    config_opts['nspawn_args'] = ['--capability=cap_ipc_lock']
+    config_opts['nspawn_args'] = ['--capability=cap_ipc_lock', '--pipe']
     config_opts['use_container_host_hostname'] = True
     config_opts['use_bootstrap_container'] = False
```

But we'd have to know that `--pipe` is suppported by systemd-nspawn.

praiskup:
Even the live logs in copr seem to be different because of this:
```
rootdir: /builddir/build/BUILD/pip-19.3.1, inifile: setup.cfg
collected 1655 items / 137 deselected / 1518 selected                          [0m

tests/functional/test_broken_stdout.py [32m.[0m[32m.[0m[32m.[0m[36m  
```
(dunno why those special characters aren't filtered out, those should be...)

I'd prefer to flip back mock to use `--pipe` rather soon.  @xsuchy wdyt?

PR #453, opened 2020-01-15
mock: use systemd-nspawn --pipe argument

praiskup:
Fixes: #432

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -59,6 +59,7 @@ import shutil
 import sys
 import time
 import copy
+import subprocess
 
 # pylint: disable=import-error
 from functools import partial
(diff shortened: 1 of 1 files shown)


xsuchy:
Would it be better to put this into `py/mockbuild/util.py:setup_default_config_opts()`? Just saying.

praiskup:
In the `setup_default_config_opts` I'm not yet sure whether user used --isolation=chroot or not, and calling systemd-nspawn binary when it's not needed isn't desired.  What about `set_config_opts_per_cmdline`?

praiskup:
(i've implemented your f2f comment, so I'm not adding --pipe if it is already there)

Issue #381, opened 2019-10-18
Bind mount local repos to bootstrap chroot when use_bootstrap_image is used


mkutlak:
```
ERROR: Command failed: 
 # /usr/bin/systemd-nspawn -q -M 1cc147f14ac049e4a7cc7c64a85c6758 -D /var/lib/mock/187917158-bootstrap/root -a --capability=cap_ipc_lock --bind=/tmp/mock-resolv.d3n75yyg:/etc/resolv.conf --setenv=TERM=vt100 --setenv=SHELL=/bin/bash --setenv=HOME=/var/lib/mock/187917158/rpmconfig --setenv=HOSTNAME=mock --setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin --setenv=PROMPT_COMMAND=printf "\033]0;<mock-chroot>\007" --setenv=PS1=<mock-chroot> \s-\v\$  --setenv=LANG=en_US.UTF-8 --setenv=LC_MESSAGES=C.UTF-8 /usr/bin/dnf --installroot /var/lib/mock/187917158/root/ --releasever 31 --setopt=deltarpm=False --allowerasing --disableplugin=local --disableplugin=spacewalk install will-crash shadow-utils abrt-addon-ccpp rpm gdb --setopt=tsflags=nocontexts
DEBUG: Unsharing. Flags: 134217728
Failed to set locale, defaulting to C.UTF-8
No matches found for the following disable plugin patterns: local, spacewalk
Failed to download metadata for repo 'retrace_fedora'
Ignoring repositories: retrace_fedora
No match for argument: will-crash
No match for argument: shadow-utils
No match for argument: abrt-addon-ccpp
No match for argument: rpm
No match for argument: gdb
Error: Unable to find a match: will-crash shadow-utils abrt-addon-ccpp rpm gdb
```

`default.cfg`:
```
config_opts['root'] = '187917158'
config_opts['target_arch'] = 'x86_64'
config_opts['chroot_setup_cmd'] = 'install will-crash shadow-utils abrt-addon-ccpp rpm gdb'
config_opts['package_manager'] = 'dnf'
config_opts['releasever'] = '31'
config_opts['bootstrap_image'] = 'fedora:31'
config_opts['plugin_conf']['ccache_enable'] = False
config_opts['plugin_conf']['yum_cache_enable'] = False
config_opts['plugin_conf']['root_cache_enable'] = True
config_opts['plugin_conf']['bind_mount_enable'] = True
[...]

PR #454, opened 2020-01-20
mock: bind mount local repos to bootstrap chroot

dturecek:
Fixes: #381

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -8,12 +8,14 @@ import shutil
 import sys
 import time
 from textwrap import dedent
+from configparser import ConfigParser
 
 from distutils.dir_util import copy_tree
 
 from . import util
 from .exception import BuildError, Error, YumError
 from .trace_decorator import traceLog, getLog
+from .mounts import BindMountPoint
 
 fallbacks = {
     'dnf': ['dnf', 'yum'],
@@ -295,6 +297,34 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
             self.copy_distribution_gpg_keys()
         self.initialize_config()
 
+        try:
+            self._bind_mount_repos_to_bootstrap()
+        except Exception as e:
+            getLog().warning(e)
+
+    def _bind_mount_repos_to_bootstrap(self):
+        if not self.buildroot.is_bootstrap:
+            return
+
+        config = ConfigParser(strict=False)
+        config.read_string(self.pkg_manager_config)
+
+        for section in config.sections():
+            if 'baseurl' not in config[section]:
+                continue
+            baseurl = config[section]['baseurl']
+            # triple slash, we only accept absolute pathnames
+            if not baseurl.startswith('file:///'):
+                continue
+            srcdir = baseurl[7:]
+            destdir = self.buildroot.make_chroot_path(srcdir)
+
+            bind_mount_point = BindMountPoint(srcpath=srcdir, bindpath=destdir)
+
+            # we need to use user mounts as essential mounts are used
+            # only for installing into bootstrap chroot
+            self.buildroot.mounts.add_user_mount(bind_mount_point)
+
     def initialize_config(self):
         # there may be configs we get from container image
         util.rmtree(self.buildroot.make_chroot_path('etc', 'yum.repos.d'))
@@ -369,6 +399,7 @@ class Yum(_PackageManager):
                            plugins=1
                            pluginconfpath={0}""".format(pluginconf_dir)))
 
+        self.pkg_manager_config = config_content
         check_yum_config(config_content, self.buildroot.root_log)
 
         # write in yum.conf into chroot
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +296,24 @@ def initialize(self):
         self.copy_certs()
         if self.buildroot.is_bootstrap:
             self.copy_distribution_gpg_keys()
-        self.initialize_config()
+        config_content = self.initialize_config()
+
+        config = ConfigParser()
+        config.read_file(StringIO(config_content))
```
praiskup: there's read_string()

mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +296,24 @@ def initialize(self):
...
+
+        for section in config.sections():
+            if 'baseurl' not in config[section]:
+                continue
+            srcdir = config[section]['baseurl'].split('file://')[-1]
+            util.mkdirIfAbsent(srcdir)
```
praiskup: You probably don't want to create the srcdir.  If the directory doesn't exist, it's an user error.

mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +296,24 @@ def initialize(self):
...
-        self.initialize_config()
+        config_content = self.initialize_config()
+
+        config = ConfigParser()
+        config.read_file(StringIO(config_content))
+
+        for section in config.sections():
+            if 'baseurl' not in config[section]:
+                continue
+            srcdir = config[section]['baseurl'].split('file://')[-1]
```
praiskup:
```python
baseurl = config[section]['baseurl']
# triple slash, we only accept absolute pathnames 
if not baseurl.startswith('file:///'):
    continue
srcdir = srcdir[7:]
```
```

File: mock/py/mockbuild/package_manager.py [hunk 4]
```diff
@@ -293,7 +296,24 @@ def initialize(self):
...
+            util.mkdirIfAbsent(srcdir)
+            destdir = srcdir
+
+            bind_mount_point = BindMountPoint(
+                    srcpath=srcdir,
+                    bindpath=self.buildroot.make_chroot_path(destdir)
```
praiskup: otoh, you are probably responsible for creating `self.buildroot.make_chroot_path(destdir)`

mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +296,24 @@ def initialize(self):
...
+                )
+            self.buildroot.mounts.essential_mounts.append(bind_mount_point)
+            bind_mount_point.mount()
```
praiskup: this is not valid, the mount point should mount automatically when (and only when) needed

mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +295,32 @@ def initialize(self):
         self.copy_certs()
         if self.buildroot.is_bootstrap:
             self.copy_distribution_gpg_keys()
-        self.initialize_config()
+        config_content = self.initialize_config()
```
praiskup: nit, instead of returning the variable - could we e.g. set `self.pkg_manager_config` variable and use it?  (it's somewhat weird that the method sets some attributes, and some are returned)

mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +295,32 @@ def initialize(self):
...
+            # triple slash, we only accept absolute pathnames
+            if not baseurl.startswith('file:///'):
+                continue
+            srcdir = baseurl[7:]
+            destdir = self.buildroot.make_chroot_path(srcdir)
+
+            bind_mount_point = BindMountPoint(
+                    srcpath=srcdir,
+                    bindpath=destdir
+                )
```
praiskup: weird indent

mock/py/mockbuild/package_manager.py
```diff
@@ -293,7 +295,32 @@ def initialize(self):
         self.copy_certs()
         if self.buildroot.is_bootstrap:
             self.copy_distribution_gpg_keys()
-        self.initialize_config()
+        config_content = self.initialize_config()
+
+        if not self.buildroot.is_bootstrap:
+            return
+
+        config = ConfigParser()
```
praiskup: Since dnf allows duplicate `baseurl` statements, while the last seems to win.

praiskup: And, per f2f discussion, we should also probably ignore all errors which can happen
during the attempt to parse the config.  The `baseurl=file://` support is rather rarely
used thing which shouldn't even by accident break normal `--bootstrap` use-cases.

dturecek:
Fixed.

praiskup:
lgtm

praiskup:
FTR, @dturecek documented a limitations here:
https://github.com/rpm-software-management/mock/wiki/Feature-bootstrap/_compare/0dc3116b801e90debc32161d9c80829ceed5fe32...7ae2a630e0d19b5224ec6b7bd3850a4d8b246c4d

PR #455, opened 2020-01-21
Update epel-8 config template to include modular repos

mmathesius:
Also updated to include missing non-modular source repo


praiskup:
Looks good, thank you.

praiskup:
(disabled by default)

mmathesius:
@praiskup,

> (disabled by default)

Yes,  just like the modular repos are disabled by default in the fedora config template files.

Thanks for the quick review and merge!

Issue #459, opened 2020-01-24
problems with terminal with nspawn and --shell 


praiskup:
```
$ mock -r fedora-rawhide-x86_64 --shell
...
Start: shell
sh: cannot set terminal process group (-1): Inappropriate ioctl for device
sh: no job control in this shell
<mock-chroot> sh-5.0# 
```

Pressing Ctrl-C:
```
<mock-chroot> sh-5.0# Container e698917bd45e4fa1a07e7b00d38fab7e terminated by signal KILL.
Finish: shell
ERROR: Exiting on user interrupt, <CTRL>-C
```

... the terminal stays broken, `reset` doesn't help.

PR #460, opened 2020-01-25
nspawn: better handle --console=pipe

praiskup:
We can not use `--pipe` globally to not break interactive shell in
--shell mode.  So with new systemd-nspawn mimic the behavior of old
systemd-nspawn.

Complements: b1b5730c6d584f98abc901693c547d7d0910e92b
Relates: #432
Fixes: #459


praiskup:
@FrostyX @xsuchy @dturecek can anyone PTAL?

dturecek:
+1

praiskup:
thanks for the review!

PR #463, opened 2020-01-30
host overrides

xsuchy:
directories in `etc/host-overrides` folder should be named:
  rhel-%{rhel}
  fedora-%{fedora}
  ...
and are in SPEC files appended to config. E.g
  etc/host-override/fedora-30/fedora-31-x86_64.cfg
is appended to
  etc/mock/fedora-31-x86_64.cfg
if we are building mock-core-configs package for fedora-30

### DIFF
--- a/mock-core-configs/etc/host-overrides/README
+++ b/mock-core-configs/etc/host-overrides/README
@@ -1,0 +1,9 @@
+directories in this folder should be named:
+  rhel-%{rhel}
+  fedora-%{fedora}
+  ...
+and are in SPEC files appended to config. E.g
+  host-override/fedora-%{fedora}/fedora-31-x86_64.cfg
+is appended to
+  mock/fedora-31-x86_64.cfg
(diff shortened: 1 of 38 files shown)


Conan-Kudo:
@xsuchy Why not just name the file `<os>-<ver>.cfg` and have all the `<os>-<ver>-<arch>.cfg` ones symlink to that?

praiskup:
I in particular dislike the maintenance POV here;  the number of configs can easily
explode (each host variant x each target variant).  But I don't see a  better way now,
so +1.

xsuchy:
@Conan-Kudo the `<os>-<ver>-<arch>.cfg` are symlinks to first one config (aarch64). If I put `<os>-<ver>.cfg` there, the code in spec file would be more complicated.

praiskup:
Closed by 6dabae8319e1ed54c4d0fbc397df794660dbdeec

Issue #458, opened 2020-01-24
rename config_opts[use_bootstrap_container] to config_opts[bootstrap]?


praiskup:
Or any better name?

The thing is that we want to flip this ON by default, which means that e.g. Koji might
likely want to turn this OFF.  And, despite the misleading cmdline option is
`--bootstrap-chroot`, they'd need to use even more misleading option
`config_opts['use_bootstrap_container'] = False`.

Neither _chroot_ nor _container_ is correct term 
because it depends on `use_nspawn` option which of them is used.  Also,
since there`s now `--use-bootstrap-image`, people tend to mix-up "containers"
with "images".  So they think `--bootstrap-chroot` is equivalent to
`--use-bootstrap-image` and vice versa.

This isn't probably material for v1.5, but we should do something about it,
ideas?

xsuchy:
Yeah, `use_bootstrap_container` is clumsy name. I think changing it to simple `bootstrap_chroot` will be nice.

The `chroot` in the name is with the outlook that one day we will have to address bootstrapping RPMs  https://docs.fedoraproject.org/en-US/packaging-guidelines/#bootstrapping

I think we can do that for 1.5 without a problem.

PR #464, opened 2020-02-02
mock: use_bootstrap_container=True default

praiskup:
.. again, after some time.  The residual issues should be fixed now, and
the mock caches were enhanced to make this less obtrusive (PR#365).

Switching per discussion/announcement in:
https://lists.fedoraproject.org/archives/list/\
buildsys@lists.fedoraproject.org/thread/XHTQD4AAAEGPTDPQ2TCGXUCY3MAC52G2/

### DIFF
--- a/mock/etc/mock/site-defaults.cfg
+++ b/mock/etc/mock/site-defaults.cfg
@@ -123,7 +123,7 @@
 # that chroot to make the actual build chroot, set this to True.
 # This is useful when the target may require newer RPM features
 # than are available on the host.
-# config_opts['use_bootstrap_container'] = False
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@praiskup What will happen when there isn't a container (like with openSUSE)? And what if `podman` isn't available?

praiskup:
I guess you are hitting the issue #458, this only enables bootstrap - it doesn't depend on podman.

xsuchy:
@Conan-Kudo the podman is needed only when `config_opts['use_bootstrap_image'] = True` but this still default to False.

Issue #461, opened 2020-01-29
dynamic buildrequires error is only visible with --verbose


praiskup:
```
DEBUG: DEBUG: Unsharing. Flags: 134217728
DEBUG: error: Missing rpmlib features for python-copr-1.100-1.git.28.6557bf7.fc31.noarch:
DEBUG: error:   rpmlib(DynamicBuildRequires) <= 4.15.0-1
DEBUG: error: /builddir/build/originals/python-copr-1.100-1.git.28.6557bf7.fc31.src.rpm cannot be installed
```
Should be visible without --verbose.

PR #466, opened 2020-02-04
mock: print stdout/stderr if 'install_srpm' failed

praiskup:
It is kind of strange to use returnOutput=True here when we don't
actually use the output in caller, but this parameter also causes that
the raised exception (in case of command failure) contains both stderr
and stdout.  As a result, mock ends with very clear error:

    /usr/bin/systemd-nspawn -q -M 0f25244d38ab4d32a46a109813628224 -D
    ... all the nspawn args ....
    /bin/rpm -Uvh --nodeps /builddir/build/originals/python-copr-1.100-1.git.36.dff5221.fc31.src.rpm
    error: Missing rpmlib features for python-copr-1.100-1.git.36.dff5221.fc31.noarch:
    error:  rpmlib(DynamicBuildRequires) <= 4.15.0-1
    error: /builddir/build/originals/python-copr-1.100-1.git.36.dff5221.fc31.src.rpm cannot be installed

Instead of:

    /usr/bin/systemd-nspawn -q -M 0f25244d38ab4d32a46a109813628224 -D
    ... all the nspawn args ....
    /bin/rpm -Uvh --nodeps /builddir/build/originals/python-copr-1.100-1.git.36.dff5221.fc31.src.rpm

Fixes: #461

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -646,7 +646,8 @@ class Commands(object):
                                 shell=False, uid=self.buildroot.chrootuid, gid=self.buildroot.chrootgid,
                                 user=self.buildroot.chrootuser,
                                 nspawn_args=self._get_nspawn_args(),
-                                unshare_net=self.private_network)
+                                unshare_net=self.private_network,
+                                returnOutput=True)
 
     @traceLog()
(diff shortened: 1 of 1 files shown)


praiskup:
Seems easy enough, merging.

PR #467, opened 2020-02-04
mock: rename use_bootstrap_container to use_bootstrap

praiskup:
And update the in-code documentation to better describe the difference
between use_bootstrap and use_bootstrap_image, and when one may want to
use them.

Fixes: #458


praiskup:
thanks for review!

PR #468, opened 2020-02-05
maint: bump to v2.0, and declare 'master' is the main branch

praiskup:
This should simplify our workflow with merging devel<->master, and also
it should make the development more clear (random contributors
intuitively want to work with master branch).

Update release workflow.


praiskup:
thanks for the review!

Issue #469, opened 2020-02-06
mock chain seems to be broken, per testsuite


praiskup:
Looking at release notes I totally forgot about this :-(.  I'll have a look first at this
and fix before announcement and official builds.

AdamWill:
This was a bit of a rude change to push to stable releases, because it breaks any config that inherits from one of the changed configs and adds settings to the `[yum.conf]` section - because now that section doesn't exist any more.

This actually broke openQA update testing, because we do exactly that as part of it.

praiskup:
This PR has nothing to do with the status, right?  The yum.conf ->
dnf.conf default change was done long time ago, and we had to push it
eventually one day (or revert).  The only new thing in this PR is that if you specify
both config options mock rants -- that's because (especially with the
changed in mock-core-configs) we need to warn users that they have broken
configuration, instead of not-using and ignoring the additions to
`yum.conf`.

Do you have idea how to mitigate this in mock?  How to make both
config_opts['yum.conf'] / config_opts['dnf.conf'] equivalent?  Patches are
welcome.

I guess you are using `config_opts['yum.conf'] += `.  Now you can do
`config_opts[config_pts['package_manager'] + '.conf'] += """`.

praiskup:
> This actually broke openQA update testing, because we do exactly that as part of it.

Can I help with this somehow?

AdamWill:
I fixed it already, it's just...kinda unexpected to have to do it.

PR #471, opened 2020-02-07
chain: fix mockchain with --bootstrap-chroot

praiskup:
The chroot mount into bootstrap chroot was added to 'managed_mounts' for
each build.  So the final managed_mounts array grow and got mixed order,
.. and the main chroot directory become invisible eventually
(overmounted by empty tmpfs filesystem).

So now, move the managed_mounts.append() call to place where it is going
to be executed only once, even for --chain mode.

I tested mockchain before - and I thought that mockchain works fine with
--bootstrap-chroot, but I was confused because I had
`plugin_conf.tmpfs_opts.keep_mounted=True` configuration and this
effectively hides this bug.

Fixes: #469

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -64,7 +64,7 @@ import subprocess
 # pylint: disable=import-error
 from functools import partial
 from mockbuild import util
-from mockbuild.mounts import BindMountPoint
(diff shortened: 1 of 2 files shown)


praiskup:
Fixed by 4658933b12f405681b29a3fbda224a499511810e

PR #472, opened 2020-02-07
Cleanup working with `dnf.conf` and `yum.conf`

praiskup:


sergiomb2:
Hi , I think this PR is a problem , I have to check it better to understand what is the goal , but I should have configurations that works with dnf and with yum .

mock/py/mockbuild/util.py
```diff
@@ -1698,7 +1700,7 @@ def add_local_repo(config_opts, baseurl, repoid=None, bootstrap=None):
 best=1
 """.format(repoid=repoid, baseurl=baseurl)
 
-    config_opts['yum.conf'] += localyumrepo
+    config_opts['{0}.conf'.format(config_opts['package_manager'])] += localyumrepo
```
sergiomb2: here , it beaks configuration if we want use dnf when we have yum.conf , for example epel7 , I guess

praiskup: Indeed.  Can you please consider #486, would that help?

praiskup: That said, it should work just fine if `package_manager` is in sync with `{yum,conf}.conf`.  It should
be the case in the latest mock-core-configs.

sergiomb2: why not revert these commits ? yum.conf for all ! :)

sergiomb2: what are the advantages, perspectives of having dnf.conf ?

Issue #474, opened 2020-02-11
mock --install /usr/bin/time: No match for argument: /builddir/time


hroncok:
### Short description of the problem

When I try to install `/usr/bin/time` into mock, it attempts to install `/builddir/time` instead.

### Output of `rpm -q mock`

mock-2.0-2.fc31.noarch

### Steps to reproduce issue

```
$ mock -r fedora-rawhide-x86_64 --install /usr/bin/time
...
No match for argument: /builddir/time
Error: Unable to find a match: /builddir/time
```

### Any additional notes

<details>
<summary>

`mock -r fedora-rawhide-x86_64 --debug-config`

</summary>

```
config_opts['bootstrap_image'] = 'fedora:rawhide'
config_opts['build_log_fmt_str'] = '%(message)s'
config_opts['cache_alterations'] = False
config_opts['chroot_name'] = 'fedora-rawhide-x86_64'
config_opts['chroot_setup_cmd'] = 'install @buildsys-build'
config_opts['cleanup_on_failure'] = False
config_opts['cleanup_on_success'] = False
config_opts['config_file'] = '/etc/mock/fedora-rawhide-x86_64.cfg'
config_opts['config_path'] = '/etc/mock'
config_opts['config_paths'] = ['/etc/mock/site-defaults.cfg',
 '/etc/mock/fedora-rawhide-x86_64.cfg',
 '/etc/mock/templates/fedora-rawhide.tpl',
 '/home/churchyard/.config/mock.cfg']
config_opts['dist'] = 'rawhide'
config_opts['dnf.conf'] = ('\n'
 '[main]\n'
 'keepcache=1\n'
 'debuglevel=2\n'
 'reposdir=/dev/null\n'
 'logfile=/var/log/yum.log\n'
 'retries=20\n'
 'obsoletes=1\n'
 'gpgcheck=0\n'
 'assumeyes=1\n'
 'syslog_ident=mock\n'
 'syslog_device=\n'
 'install_weak_deps=0\n'
 'metadata_expire=0\n'
 'best=1\n'
 'module_platform_id=platform:f{{ releasever }}\n'
 'protected_packages=\n'
 '\n'
 '{%- macro rawhide_gpg_keys() -%}\n'
 'file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-$releasever-primary\n'
 '{%- for version in [releasever|int, releasever|int - 1]\n'
[...]

PR #475, opened 2020-02-11
allow --install /usr/bin/time [GH#474]

xsuchy:
do not modify argument if the file exist, but argument does not have *rpm extension.

It works now for:

--install zsh
--install mock.rpm
--install /usr/bin/time

Fixes: #474

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -706,8 +706,11 @@ class Buildroot(object):
         if not bootstrap:
             return filename
 
-        if not os.path.exists(filename):
+        # optimized of
+        # not os.path.exists(filename) or not filename.lower().endswith('.rpm')
+        if not (os.path.exists(filename) and filename.lower().endswith('.rpm')):
             # probably just '--install pkgname'
+            # or '--install /usr/bin/time'
             return filename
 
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/buildroot.py
```diff
@@ -706,8 +706,11 @@ def file_on_cmdline(self, filename):
         if not bootstrap:
             return filename
 
-        if not os.path.exists(filename):
+        # optimized of
+        # not os.path.exists(filename) or not filename.lower().endswith('.rpm')
```
praiskup: This comment isn't necessary, but feel free to keep it...

praiskup:
+1

hroncok:
Thanks for the fix. How fast can we anticipate a Fedora update? Th underlying regression has broken our rawhide CI and I was fine to ignore that for a while, but now it was merged also to f31 stable. We can workaround it on our side, but if you are working on a release already, we can wait ~1 week.

xsuchy:
@hroncok We are busy this week with Copr migration. Next week we likely focus on the feedback from the release. I do not want to promise anything, but the release is quite probable.

hroncok:
Thanks. Good luck with the migration.

PR #477, opened 2020-02-14
don't emit unneeded warning

remicollet:
See https://bugzilla.redhat.com/1802930

In boostrap this seems really unneeded (especially for quiet mode)

An information is already there, ans seems enough

     INFO: Using 'dnf' instead of 'yum' for bootstrap chroot

And for non boostrap case

     WARNING! WARNING! WARNING!
    You are building package for distribution which uses {0}. However your system
    does not support {0}. You can continue with {1}, which will likely succeed,

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -34,7 +34,7 @@ def package_manager_from_string(name):
     raise Exception('Unrecognized package manager "{}"'.format(name))
 
 
-def package_manager_exists_on_host(name, config_opts):
+def package_manager_exists_on_host(name, config_opts, bootstrap=False):
     option = '{}_command'.format(name)
     pathname = config_opts[option]
     if not os.path.isfile(pathname):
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/package_manager.py
```diff
@@ -47,8 +47,9 @@ def package_manager_exists_on_host(name, config_opts):
     real_pathname = os.path.realpath(pathname)
     # resolve symlinks, and detect that e.g. /bin/yum doesn't point to /bin/dnf
     if name not in real_pathname:
-        getLog().warning("Not using '%s', it is symlink to '%s'", pathname,
-                         real_pathname)
+        if not bootstrap:
+            getLog().warning("Not using '%s', it is symlink to '%s'", pathname,
+                             real_pathname)
```
praiskup: I'm not sure.  Isn't it better to uninstall yum if the warning irritates you?

praiskup:
Indeed, the only warning in `--quiet` mode is this one.. I don't think it is that important
thing when bootstrap isn't disabled.

The more I think about this, the more I tend to +1 this PR, thank you.

remicollet:
Thanks

PR #478, opened 2020-02-17
config: move rawhide-i386 to f33 as well

praiskup:
configs: templatize F31+ i386
    
    The rawhide i386 config was de-synced from rawhide.tpl by mistake, this
    patch should assure that such mistake never happens in future.
    
    The line:
        config_opts['mirrored'] = config_opts['target_arch'] != 'i686'
    though means that the config file isn't order-independend anymore, and
    that the 'target_arch' needs to be defined _before_ include()'ing the
    template.  We could do this instead:
        config_opts['mirrored'] = '{{ target_arch != i686 }}'
    Which would make it order-independend, but the variable wouldn't be
    of "bool" type, but string.  That wouldn't allow us to do convenient
    '{% if mirrored %}...{% endif %}.
    
    Relates: rhbz#1804415


praiskup:
Maybe we should enhance the templates a bit, so we don't have to bump the releasever in two files during branching time.

Issue #434, opened 2019-12-02
mock should fail build early when dynamic build requires are not satisfied


praiskup:
Commit ba5e4ef2d16d389383207430019f292457eb7f02 added a hack
which causes that even if `rpmbuild` returns 11 (some dynamic buildrequires
are missing) mock continues as if rpm succeeded.  This has negative that
the build can fail a very late during the %build process.

The problem is that the current released rpm with dynamic BR
support is broken and we can not drop the hack from mock;  at
least not till we can expect broken RPM.  @hrnciar please link the
RPM bug once it is filled.

praiskup:
We should at least throw a huge warning that we continue to %build even though
we shouldn't.  Perhaps we could detect whether RPM is fixed... and stop doing the
hack in such case.

hrnciar:
https://github.com/rpm-software-management/rpm/issues/963

ignatenkobrain:
I don't see what is the hack in mentioned comment. You can have multiple levels of dynamic BRs, so you want to iterate through generating them over and over until all fail or all get installed.

praiskup:
```
+                   if packages_after == packages_before:
+                        success = True
```

This is the hack I had in mind.  Do you think it is correct, ie that we should ignore rpm 
retval?

ignatenkobrain:
This is definitely not a hack. The problem is that mock uses `--nodeps`, so the rpmbuild is not sure whether all dependencies are satisfied or not. Obviously it can exit with `1`, but that would indicate that nosrc.rpm would not be written.

praiskup:
I don't get it.  Can you interpret what the exit status 11 means, then?

ignatenkobrain:
that buildreqs.nosrc.rpm has been generated

praiskup:
Is this documented somewhere? I read `RPMRC_MISSINGBUILDREQUIRES`.

ignatenkobrain:
I'm not sure if it is documented, however if you pass `--nodeps` you can't ever have MISSINGBUILDREQUIRES, because you omit checking it. And the only way RPM can tell you this is to return 11 always and you will have to try installing Requires of nosrc.rpm and checking whether something was installed or not (or fail).

hrnciar:
@pmatilai could you please take a look?
Is it really correct to return 11 even though all dependencies were satisfied?

hrnciar:
@pmatilai ping

pmatilai:
[...]

PR #482, opened 2020-02-20
mock: fix dynamic buildrequires unnecesarry loop

hrnciar:
per discussion here: https://github.com/rpm-software-management/rpm/issues/963
Mock should accept return code 0 from rpm.

Relates: #434

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -708,6 +708,7 @@ class Commands(object):
                 except Error as e:
                     if e.resultcode != 11:
                         raise e
+                finally:
                     max_loops -= 1
                     self.buildroot.root_log.info("Dynamic buildrequires detected")
                     self.buildroot.root_log.info("Going to install missing buildrequires")


mock/py/mockbuild/backend.py
```diff
@@ -699,7 +699,6 @@ def get_command(mode):
                 # * it fails
                 # * installSrpmDeps does nothing
                 # * or we run out of dynamic_buildrequires_max_loops tries
-                packages_before = self.buildroot.all_chroot_packages()
```
praiskup: Yes, we can drop this (no need to access the RPM db by `rpm -qa`), but we
need a replacement here.  Most likely you need to maintain a `set()` of
real build requirements across the iteration...  in each step you can get
the list from `installSrpmDeps` with small enhancements.

mock/py/mockbuild/backend.py
```diff
@@ -708,18 +711,19 @@ def get_command(mode):
...
                         raise e
+                finally:
                     max_loops -= 1
                     self.buildroot.root_log.info("Dynamic buildrequires detected")
                     self.buildroot.root_log.info("Going to install missing buildrequires")
                     buildreqs = glob.glob(bd_out + '/SRPMS/*.buildreqs.nosrc.rpm')
-                    self.installSrpmDeps(*buildreqs)
-                    packages_after = self.buildroot.all_chroot_packages()
-                    if packages_after == packages_before:
+                    new_deps = self.installSrpmDeps(*buildreqs, *deps)
```
praiskup: Shouldn't we use just `deps` argument here, without star?

praiskup: I'd perhaps use `ignored_deps=deps` to underline the purpose of the argument.

mock/py/mockbuild/backend.py
```diff
@@ -180,24 +180,27 @@ def remove(self, *rpms):
         self.buildroot.root_log.info(output)
 
     @traceLog()
-    def installSrpmDeps(self, *srpms):
+    def installSrpmDeps(self, *srpms, ignored_deps=None):
         """Figure out deps from srpm. Call package manager to install them"""
```
praiskup: Please state what `ignored_deps` means in the method documentation.

praiskup:
The commit message doesn't correspond to what is really changing.

mock/py/mockbuild/backend.py
```diff
@@ -708,6 +708,7 @@ def get_command(mode):
                 except Error as e:
                     if e.resultcode != 11:
```
praiskup: Please document the fact that we accept exit_status 0 and exit_status 11, and why.

hrnciar:
I am sorry, the PR is not yet finished. I forgot to add WIP tag.

PR #486, opened 2020-02-23
mock: 'dnf.conf' is equivalent to 'yum.conf'

praiskup:
mock: 'dnf.conf' is equivalent to 'yum.conf'
    
    Resolves: rhbz#1806039
    Relates: https://pagure.io/copr/copr/pull-request/1275
    Related discussions:
    https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2020-4f9536de5b
    https://pagure.io/koji/issue/2026
    https://lists.fedoraproject.org/archives/list/\
    devel@lists.fedoraproject.org/thread/IRMM6UYHRWOI3YM6IE4DTTV4YWF3J7G2/

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -112,13 +112,29 @@ class commandTimeoutExpired(exception.Error):
 # pylint: disable=no-member,unsupported-assignment-operation
 class TemplatedDictionary(MutableMapping):
     """ Dictionary where __getitem__() is run through Jinja2 template """
-    def __init__(self, *args, **kwargs):
-        '''Use the object dict'''
+    def __init__(self, *args, alias_spec=None, **kwargs):
+        '''
+        Use the object dict.
+
+        Optional parameter 'alias_spec' is dictionary of form:
+        {'aliased_to': ['alias_one', 'alias_two', ...], ...}
+        When specified, and one of the aliases is accessed - the
+        'aliased_to' config option is returned.
+        '''
         self.__dict__.update(*args, **kwargs)
+
+        self._aliases = {}
+        if alias_spec:
+            for aliased_to, aliases in alias_spec.items():
+                for alias in aliases:
+                    self._aliases[alias] = aliased_to
+
     # The next five methods are requirements of the ABC.
     def __setitem__(self, key, value):
+        key = self._aliases.get(key, key)
         self.__dict__[key] = value
     def __getitem__(self, key):
+        key = self._aliases.get(key, key)
         if '__jinja_expand' in self.__dict__ and self.__dict__['__jinja_expand']:
             return self.__render_value(self.__dict__[key])
         return self.__dict__[key]
@@ -944,7 +960,7 @@ def find_non_nfs_dir():
 @traceLog()
 def setup_default_config_opts(unprivUid, version, pkgpythondir):
     "sets up default configuration."
-    config_opts = TemplatedDictionary()
+    config_opts = TemplatedDictionary(alias_spec={'dnf.conf': ['yum.conf']})
     config_opts['config_paths'] = []
     config_opts['version'] = version
     config_opts['basedir'] = '/var/lib/mock'  # root name is automatically added to this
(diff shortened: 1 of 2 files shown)


sergiomb2:
well this PR also works for me

mock/py/mockbuild/util.py
```diff
@@ -944,7 +955,7 @@ def find_non_nfs_dir():
 @traceLog()
 def setup_default_config_opts(unprivUid, version, pkgpythondir):
     "sets up default configuration."
-    config_opts = TemplatedDictionary()
+    config_opts = TemplatedDictionary(alias_spec=[('dnf.conf', 'yum.conf')])
```
xsuchy: I would propose alias_spec=[{'dnf.conf': ['yum.conf',]}]

mock/py/mockbuild/util.py
```diff
@@ -112,13 +112,24 @@ def __init__(self, msg):
 # pylint: disable=no-member,unsupported-assignment-operation
 class TemplatedDictionary(MutableMapping):
     """ Dictionary where __getitem__() is run through Jinja2 template """
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, alias_spec=None, **kwargs):
         '''Use the object dict'''
```
xsuchy: Documentation here?

mock/py/mockbuild/util.py
praiskup: Done, PTAL.

sergiomb2:
why we want replace config_opts['yum.conf'] with config_opts['dnf.conf'] in [1] ? TBH I'd revert all to 
config_opts['yum.conf'] ... I don't see any advantage 

[1]
https://github.com/rpm-software-management/mock/pull/472/files

praiskup:
PR #472 tried to align with 15519fb7d35e19b0747a651a21aa54a393347c1a.

> I don't see any advantage

Everything is switching from Dnf to Yum.  Yum is dead.  This is rather cosmetic change
indeed.  It is really weird to see `package_manager = dnf` and `yum.conf = '...'`.  That
said, if we had no other chance to solve this - I'd prefer the revert as well as you prefer.

But this PR should solve the wide problems, and we can keep cleaner configs. 

But

sergiomb2:
OK Thanks for the reply, I have to think , I'm very busy with my day job ... 
Best regards,

sergiomb2:
After thinking, I have this problem to solve, what configurations I generate for use in EPEL 7 and EPEL6 (in mock-rpmfusion) [1] ? , now we have to supply 2 types of configurations , configurations for mock 1.x and confirurations for mock 2.0 ...


[1] 
https://github.com/rpmfusion-infra/mock-rpmfusion

praiskup:
The best is to keep `yum.conf` config option for `package_manager=yum` configs.  New mock
shouldn't care and accept both variants so if you really need to keep compatibility please
elaborate on the problem.

hroncok:
This is awesome \o/

sergiomb2:
> if you really need to keep compatibility please elaborate on the problem.

Not really , it is  just curiosity in knowing what is the plan for mock in EL6 and EL7 and reduce my workload.
In practice mock-rpmfusion-free depends on mock-core-configs [1] or mock 
 
I'm sorry for not helping more and be little away ... 

[1] 
https://pkgs.rpmfusion.org/cgit/free/mock-rpmfusion-free.git/tree/mock-rpmfusion-free.spec

praiskup:
> Not really , it is just curiosity in knowing what is the plan for mock in EL6 and EL7

The plan is to submit update on EL7 (v2.1) as soon as possible.  Do you really use mock
to build on EL6?

xsuchy:
>  it is just curiosity in knowing what is the plan for mock in EL6 and EL7 

EL6 got only 1.3.x version. El6 never got mock-1.4.x
I originally wanted to stop building the next major version for EL7. @praiskup had a different opinion and decided otherwise. 
EL7 is still supported, but I would highly recommend to start planning the move to EL8.

Issue #480, opened 2020-02-19
not-only bind-mount baseurl=file:// files into bootstrap, but also baseurl=/file/


praiskup:
Per discussion in:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-8c3f06d776

PR #488, opened 2020-02-23
bootstrap: bind-mount also baseurl=/absolute/dir repos

praiskup:
Per discussion in:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-8c3f06d776

Also skip directories that do not exist, which leads to unexpected
bind-mounts.

Fixes: #480


praiskup:
Thanks for review.  The tests are still broken though, I'll fix them and merge.

PR #493, opened 2020-02-26
tests: add xfail test for jinja problem with nested fields

praiskup:


praiskup:
I have potential fix (plus others ideas, which need to be in separate commits) in  https://github.com/praiskup/mock/commit/f848c5fd94c3fd8a6b3f1ab0bbc561565ff11a93.

The point is that the fix allows us to never modify the dict content when jinja rendering
is enabled, and then we can basically (somehow) copy sub-dicts of the dict to plugins
anytime (resp. access `self.buildroot.config` from plugin, temporarily disable expansion,
inherit `pugin_confs.ignd_opts` unexpanded, create new TemplatedDictionary from it,
and work with that in isolation).

PR #495, opened 2020-02-26
Add 'rendered.raw' attribute (research state)

praiskup:
This is mostly research patch.  If it proves to be wrong idea, we can calmly close this,
but I wanted to have the idea dumped somewhere.

The signd plugin could then access the values through `self.conf['opts'].raw`,
create new instance of TemplatedDictionary, and define `rpms` correctly.


praiskup:
I'm closing this for now, we don't necessarily need this.

PR #504, opened 2020-03-02
lvm_root: drop hack related to bootstrap

praiskup:
Remove one old hack that tried to secure the chroot root mount point
re-mounted into bootstrap.  This isn't needed nowadays because we
always pre over-mount that directory by tmpfs with private flag, so
there's no risk that umount in bootstrap will also umount the main mount
point.

This commit complements 4658933b12f405681b29a3fbda224a499511810e,
b87244a88f39d856f303732eca09434a6d9da4c6 and friends.

Fixes: rhbz#1805179

### DIFF
--- a/mock/integration-tests/20-lvm-plugin.tst
+++ b/mock/integration-tests/20-lvm-plugin.tst
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+
+: "${MOCKCMD=mock}"
+if test -z "$TESTDIR"; then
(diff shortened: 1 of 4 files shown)


praiskup:
@dturecek, plaese can you take a look?

praiskup:
The second patch doesn't work (testsuite doesn't pass), I need to update it.

dturecek:
+1

praiskup:
thanks for the review

PR #506, opened 2020-03-02
mock: add config options for optionally copying /etc/pki files from host

### LINKED ISSUE #492: mock-2.0 broken on Gentoo: "cannot copy tree '/etc/pki/ca-trust/extracted'"
Since many years I'm packaging and using mock on my Gentoo box to create and test software packages for EPEL and Fedora. This was working fine until release 1.4.21, but the recent release 2.0, especially commit 2be432603733ce19ee0467bb0534daefd5bf86ed (fix for #397) broke my workflow.

I don't have any requirement to load my local system certificates to the build environment therefore I would welcome if this feature of copying the certificates optionally could be disabled (and/or the path to the certificate directory be specified). Gentoo manages the system certificates a bit different from Fedora/RHEL.

### Output of `rpm -q mock`
Obviously I didn't install mock from the RPM but from source (via ebuild):
```
$ eix -e dev-util/mock
[I] dev-util/mock [1]
     Available versions:  (~)1.4.21-r1 (~)2.0 {PYTHON_SINGLE_TARGET="python3_6 python3_7"}
     Installed versions:  2.0(10:20:37 PM 02/25/2020)(PYTHON_SINGLE_TARGET="python3_7 -python3_6")
     Homepage:            https://github.com/rpm-software-management/mock
     Description:         Builds RPM packages inside chroots

[1] "linuxmonk-overlay" /var/lib/layman/linuxmonk-overlay
```
The ebuild is available in my [linuxmonk-overlay](https://github.com/ganto/linuxmonk-overlay)

FrostyX:
Fixes #492

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -11,6 +11,7 @@ from textwrap import dedent
 from configparser import ConfigParser
 
 from distutils.dir_util import copy_tree
+from distutils.errors import DistutilsFileError
 
 from . import util
 from .exception import BuildError, Error, YumError
@@ -296,7 +297,11 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
     def copy_certs(self):
         cert_path = "/etc/pki/ca-trust/extracted"
         pki_dir = self.buildroot.make_chroot_path(cert_path)
-        copy_tree(cert_path, pki_dir)
+        try:
+            copy_tree(cert_path, pki_dir)
+        except DistutilsFileError:
+            warning = "Couldn't copy certs from host, %s doesn't exist or is not readable"
+            self.buildroot.root_log.debug(warning, cert_path)
 
     def initialize(self):
         self.copy_gpg_keys()


FrostyX:
Do we want to have separate config options

    config_opts['copy_certs_from_host'] = True
    config_opts['copy_gpg_keys_from_host'] = True

Or just 

    config_opts['copy_pki_from_host'] = True

praiskup:
I'd personally avoid adding another knob TBH, but I don't raise strong
vote against this (those can be added).  Just saying that the
configuration just becomes more and more complicated.

What I think is more important here is to make things work (not traceback,
or fail) by default - so people **don't have to** adjust configuration.
Potential solution could be not to copy stuff which doesn't exist.

If we were about to make things work, not skip, we could have a config
option which says where the certificates/gpg keys/... could be.  Something
like `config_opts['extracted_ssl_certs_search_path'] = pathA:pathB`.  But
I don't think this is necessary ATM...

WRT gpg keys;  those are pretty much mandatory - without copying them even
basic fedora repositories won't work.  So basically almost nobody wants to
disable copying them (and the `glob.glob` solution used for gpg_keys is
pretty robust, I don't think users have problems with that).  OTOH, pretty
limited set of users wants to copy the ssl certs, and the copy mechanism
is brittle ATM.  I think we have to concentrate on certs only.

WDYT?  Thank you for working on this!

FrostyX:
@praiskup makes sense. Do you like this version better? Once we finish the changes, I will squash the commits together.

praiskup:
I like this, +1.

FrostyX:
Rebased

praiskup:
Thank you!

Issue #483, opened 2020-02-21
mock --chain doesn't use package/repo cache properly


fkooman:
### Short description of the problem

"Normal" mock uses package / repo cache properly, "chain" mock does not. It downloads all packages not part of the root every time, and worse, the dnf repodata as well which takes a long time...

### Output of `rpm -q mock`

mock-2.0-2.fc31.noarch

### Steps to reproduce issue

1. use mock --chain instead of mock

PR #507, opened 2020-03-03
mockchain: don't run buildroot.finalize() for each package

praiskup:
That invalidates mount points, and probably cleans the buildroot too
much without reason.  There's automatic buildroot.clean() call before
each package anyways.

Fixes: #479

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -475,10 +475,6 @@ class Commands(object):
                             do_rebuild(self.config, self, buildroot, options, [pkg])
                     except Error:
                         build_ret_code = 1
-                    finally:
-                        buildroot.finalize()
-                        if buildroot.bootstrap_buildroot is not None:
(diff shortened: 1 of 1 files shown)


praiskup:
@xsuchy wdyt?  I tried to look at the code before 628cfbeea9af690e3c929abf4e9345215c211e83, and the finalize() call wasn't there before.

praiskup:
After f2f discussion, this was initially added because the finalize()
was called through `mockchain -> main() [mock] -> finalize()` before.

But since we call `buildroot.delete()` anyways before each package build
(which itself may be too much, let's see what will happen with the #483),
the only thing that needs our attention in finalize is buildroot.tmpdir.
But that path is only used for nosync plugin ATM ..  and that plugin only
uses the directory for installing one so file.  This change looks safe
to me, and I plan to merge it tomorrow.

Btw., I also tested `*chain*.tst` tests, and they passed.

PR #508, opened 2020-03-04
mock: don't clean between consecutive builds

FrostyX:
Fixes: #483

The issue doesn't seem to be --chain specific. It is reproducible by both

    mock -r <chroot> --chain foo.src.rpm bar.src.rpm
    mock -r <chroot> foo.src.rpm bar.src.rpm --resultdir=/foobar

Unfortunatelly the fix can't be done in some common code, but separately
for both of these cases. Basically we need to make sure to not call
`commands.clean()` between consecutive builds. Interesting is, that
if --chain is not used, we can clean before the first build and still
go from cache, but when using --chain, we can't clean before any of them.

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -472,7 +472,9 @@ class Commands(object):
                         if os.path.exists(success_file):
                             build_ret_code = 2
                         else:
-                            do_rebuild(self.config, self, buildroot, options, [pkg])
+                            config_opts = self.config.copy()
+                            config_opts["clean"] = False
+                            do_rebuild(config_opts, self, buildroot, options, [pkg])
                     except Error:
                         build_ret_code = 1
                     finally:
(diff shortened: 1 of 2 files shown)


praiskup:
Well, this is weird.  Should `commands.clean()` clean dnf/yum caches?

praiskup:
This doesn't seem to be reproducible without bootstrap.

FrostyX:
> Well, this is weird. Should commands.clean() clean dnf/yum caches?

I don't know. But I tried to dissect what line exactly causes the issue and it seems to be this one in `buildroot.py:delete`:

    util.rmtree(self.basedir, selinux=self.selinux)

praiskup:
There's IMO some wrong order of umount/mount pair withing bootstrap chroot;
that the caches are not wiped away, but are just invisible from within bootstrap.

praiskup:
Closing in favor of #512

PR #509, opened 2020-03-04
config: explicitly disable bootstrap for OpenSUSE

praiskup:
First, doing bootstrap to install `dnf` stack, to later install `zypper`
distro doesn't make sense to me.

Second, per issue #500, ssl certificates inside OpenSUSE bootstrap
chroot work differently from Fedora (at least OpenSUSE uses different
paths).

That said, to make everything work smoothly, we would have to have at
least another `class Zypper(_PackageManager)` abstraction, and better
handle certificates so we have compatibility for Fedora <-> OpenSUSE.

Fixes: #500

### DIFF
--- a/mock-core-configs/etc/mock/opensuse-leap-15.0-aarch64.cfg
+++ b/mock-core-configs/etc/mock/opensuse-leap-15.0-aarch64.cfg
@@ -8,6 +8,7 @@ config_opts['useradd'] = '/usr/sbin/useradd -o -m -u {{chrootuid}} -g {{chrootgi
 config_opts['releasever'] = '15.0'
 config_opts['macros']['%dist'] = '.suse.lp150'
 config_opts['package_manager'] = 'dnf'
+config_opts['use_bootstrap'] = False
 
 config_opts['dnf.conf'] = """
 [main]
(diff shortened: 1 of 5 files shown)


praiskup:
I worry a bit that building Fedora from OpenSUSE box using mock with bootstrap
is broken as well, though I have no data that anyone ever used this scenario.

praiskup:
@brianjmurrell, is this sufficient solution to #500?  I'm afraid we don't have time
now for more debugging OpenSUSE chroots.

Conan-Kudo:
@praiskup I plan to work with @simotek to fix openSUSE patterns so that zypper isn't pulled in when we don't use it. This is something that I definitely want fixed.

praiskup:
On Wednesday, March 4, 2020 12:30:32 PM CET Neal Gompa (ニール・ゴンパ) wrote:
> @praiskup I plan to work with @simotek to fix openSUSE patterns so that zypper
> isn't pulled in when we don't use it. This is something that I definitely want
> fixed.

Can you elaborate on this thought?  Do you think that it makes any sense
to have bootstrap enabled for opensuse?

Conan-Kudo:
@praiskup It will be mandatory, since openSUSE is moving to `ndb` rpmdb (and removing bdb backend at the same time) and I can't get it enabled in Fedora to maintain compatibility (see [RH#1795450](https://bugzilla.redhat.com/show_bug.cgi?id=1795450)).

praiskup:
Thanks for the info.  Closing this in favor of #511 then.

PR #512, opened 2020-03-05
bootstrap: bind-mount the inner root mount privately

praiskup:
This caused problems with dnf caches before when we called
    commands.clean() -> commands.init()
sequence (typically mock --chain).

That's because commands.clean() calls buildroot.mounts.umountall(),
which according to the order of bind-mounts first "recursively" (lazily)
umounted the inner-chroot from bootstrap chroot.  That though affected
the original chroot mount as well (umounted caches).  The following
particular dnf cache unmount attempts failed since the mount points were
already unmounted...  this kept the mount points in 'mounted = True'
state.

Having the 'mounted = True' caused that the 'init()' command did not
re-mount the caches.

Fixes: #483
Supersedes: #508

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -755,7 +755,7 @@ def main():
                                  options="rprivate"))
         buildroot.mounts.managed_mounts.append(
             BindMountPoint(buildroot.make_chroot_path(), inner_mount,
-                           recursive=True))
+                           recursive=True, options="rprivate"))
 
     signal.signal(signal.SIGTERM, partial(handle_signals, buildroot))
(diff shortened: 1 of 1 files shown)


FrostyX:
Makes sense, LGTM.

FrostyX:
> Makes sense, LGTM.

The issue investigation ^^
And the fix works for me.

I don't particularly understand why `options="rprivate"` fixed it, though.

praiskup:
That is just the `mount -o rprivate` option for the mount point.  Maybe we should
reference `man mount_namespaces(7)` somewhere.

praiskup:
Fixed by 00bb3546dd4a2f16e3f9ff0cf028c7137d90a4c1

PR #517, opened 2020-03-06
Better support for bootstrap_package_manager config option

pghmcfc:
The bootstrap buildroot is constructed before mapping `bootstrap_*` options into `bootstrap_buildroot_config`, but the constructor of the buildroot wants to know what the desired package manager to set up the bootstrap buildroot is. If a `bootstrap_package_manager` option is specified, this should be used as the desired package manager rather than what's been requested for set-up of the actual main buildroot.

This is not a terribly important option to get right since mock will check what's available on the host system and fall back sensibly. However, this will usually result in a warning message appearing in
the logs, which is nice to avoid.

For example, if building on Fedora (with dnf) for a target that uses yum, mock currently outputs a message about yum being symlink to dnf and hence using dnf instead, right at the start of construction of the bootstrap buildroot, even if `config_opts['bootstrap_package_manager'] = 'dnf'`.

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -717,6 +717,11 @@ def main():
         # disable updating bootstrap chroot
         bootstrap_buildroot_config['update_before_build'] = False
 
+        # check if a different package manager is wanted for the bootstrap buildroot
+        if 'bootstrap_package_manager' in bootstrap_buildroot_config:
+            bootstrap_buildroot_config['package_manager'] = \
+                bootstrap_buildroot_config['bootstrap_package_manager']
+
         bootstrap_buildroot_state = State(bootstrap=True)
         bootstrap_plugins = Plugins(bootstrap_buildroot_config, bootstrap_buildroot_state)
         bootstrap_buildroot = Buildroot(bootstrap_buildroot_config,


pghmcfc:
Looks like commit 6c9682cc fixes this actually, though I think this way is cleaner.

praiskup:
Thanks for the patch.

> For example, if building on Fedora (with dnf) for a target that uses yum, mock
> currently outputs a message about yum being symlink to dnf and hence using dnf
> instead, right at the start of construction of the bootstrap buildroot, even
> if config_opts['bootstrap_package_manager'] = 'dnf'.

Is there any other reason to add this new option?  If you are about to
only silence the info message, I'd rather enhance it a bit so it isn't
printed when you feel it shouldn't.

IMO, with new mock there shouldn't be any reason to set `package_manager`
differently from `bootstrap_package_manager`.

praiskup:
> Looks like commit 6c9682c fixes this actually, though I think this way is cleaner.

Ah, this ...  I forgot about this since I'm running pre-release mock for
some time.

Well, the thing that the semantics of `package_manager` is
"what package manager to use to install stuff into the chroot".
If you use `use_bootstrap=False`, not having the right package
manager is serious problem.  We need to warn.

If you `use_bootstrap=True`, it isn't that important to warn the user ...
because that's exactly why the bootstrap chroot was introduced (to install
the right package manager first).

Neither of those cases ^^ should motivate you to replace the
value of package_manager value.  WDYT?

pghmcfc:
I'm happy enough if the warning isn't printed. I found it somewhat confusing and expected to be able to silence it by being explicit about the package manager that I was using for the bootstrap setup, but it didn't work for me (using stable mock in Fedora 31), hence this PR.

praiskup:
Ok, thanks for the feedback and PR, I'm closing for now as this is probably not needed.
 If you had any other fight with the warning, please reopen.

PR #518, opened 2020-03-06
Use configured yum commands in bootstrap

pghmcfc:
When the target buildroot is constructed using yum, the commands to invoke yum are hard-coded as '/usr/bin/yum' and '/usr/bin/yum-builddep' regardless of what may have been configured in config_opts. This was a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1446294

Since these are the default values of `config_opts['yum_command']` and `config_opts['yum_builddep_command']` anyway, it should be safe to use the values of these options instead of hard-coding them.

Without this change, it's not possible to use yum as package manager when the target system has yum in /usr/bin/yum-deprecated rather than /usr/bin/yum. With the change, the user can specify
`config_opts['yum_command'] = '/usr/bin/yum-deprecated'` and it works as expected.

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -395,10 +395,9 @@ class Yum(_PackageManager):
         self.install_command = config['yum_install_command']
         self.builddep_command = [config['yum_builddep_command']]
         if bootstrap_buildroot is not None:
-            # we are in bootstrap so use old names
-            self.command = '/usr/bin/yum'
-            yum_deprecated_path = '/usr/bin/yum'
-            yum_builddep_deprecated_path = '/usr/bin/yum-builddep'
+            # we are in bootstrap so use configured names
+            yum_deprecated_path = config['yum_command']
+            yum_builddep_deprecated_path = config['yum_builddep_command']
             yum_deprecated_path = bootstrap_buildroot.make_chroot_path(yum_deprecated_path)
             yum_builddep_deprecated_path = bootstrap_buildroot.make_chroot_path(yum_builddep_deprecated_path)
         else:


praiskup:
Hmm, thanks for the patch!

> Without this change, it's not possible to use yum as package manager when the
> target system has yum in /usr/bin/yum-deprecated rather than /usr/bin/yum.
> With the change, the user can specify config_opts['yum_command'] =
> '/usr/bin/yum-deprecated' and it works as expected.

Interesting indeed, do you have example use-case of this, or is this just
theoretical use-case?

If yum is `yum-deprecated` in the chroot, you actually want to install dnf
into the buildroot, or?

pghmcfc:
It's not theoretical. I'm trying to build packages in Fedora 22 and 23 chroots (whether there's good reasons for that or not is a separate issue...). Anyway, the default package manager there is dnf and hence we have /usr/bin/yum-deprecated for real yum. However, when packages were originally being built in Fedora's koji for Fedora 22 and 23, yum was used rather than dnf - koji only switched to dnf for Fedora 24.

Now I could probably live with using dnf for these Fedora releases except that in Fedora 22, the version of dnf there is unable to import GPG keys whilst running in a chroot (I get an error like the one in https://bugzilla.redhat.com/show_bug.cgi?id=1224908) so I need either to use yum (which works OK) or not use bootstrap mode so that the buildroot is set up using a modern dnf. I would much rather use the former as it's cleaner, more representative of how packages were built for that release and probably more future-proof.

praiskup:
Meh, I was afraid of something like this..., I need to think about this a bit because
I'm not able to rule-out possible consequences.

praiskup:
Ok, I don't see any real problem in this patch.  If it helps you, I'm +1.  Nobody objected
till now, so merging.  Thank you!

PR #526, opened 2020-03-11
add rpkg spec preprocessing capability

clime:
Hello!

This pull request adds spec preprocessing capability to mock by using `preproc` and `rpkg-macros` packages (and also `preproc-rpmspec` which is a wrapper around these two). The whole rationale for this pull request is described here: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/FSVFSQJ5R4WKLH7WIYZQ22CUXAEYZPNK/

I have manually tested it from `epel-6` up to `fedora-rawhide` under `x86_64` architecture. The needed packages are not yet in stable Fedoras (and EPELs) so using copr repos, as described below, is needed. Rawhide update is probably going to come sooner.

The preprocessing is enabled only if new config option `config_opts['rpkg_preprocessing']` is `True` (by default it is set to `False`). There are two other new config options `config_opts['rpkg_preprocessing_requires']` and `config_opts['rpkg_preprocessing_cmd']` which specify what package is needed to perform the preprocessing task and how the command should be called.

Even if the the feature is enabled in mock, one still needs to have `rpkg.conf` file in the repository with the following content to get the spec file preprocessed:
```
[rpkg]
preprocess_spec = True
```

If this is satisfied as well, the rpkg preprocessing requirements are installed into the chroot and the preprocessing command is run which takes the spec file under `/buildir/build/SOURCES/`, pipes it to `preproc` that has `rpkg-macros` loaded and the output spec file is then put under `/buildir/build/SPECS/` instead of the original unprocessed spec file.

Please, let me know if any tests or further improvements are needed. I noticed a small issue that under Fedora chroots, dnf always refreshes repos twice, once at the beginning when `dnf update` is invoked and then slightly after when `self.buildroot.pkg_manager.install(*rpkg_preprocessing_requires, check=True)` is called. Maybe this could be optimized but I didn't figure how.

An example spec file that contains macros to be processed by `preproc` may look like this:
https://pagure.io/hello_rpkg_release/blob/master/f/hello_rpkg.spec.rpkg

I have run the tests by manually adding the following copr repos into mock configuration files (together with creating the needed `rpkg.conf` file with the above mentioned content):
```
[copr:copr.fedorainfracloud.org:clime:preproc]
name=Copr repo for preproc owned by clime
baseurl=https://download.copr.fedorainfracloud.org/results/clime/preproc/fedora-$releasever-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/clime/preproc/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1

[copr:copr.fedorainfracloud.org:clime:rpkg-macros]
name=Copr repo for rpkg-macros owned by clime
baseurl=https://download.copr.fedorainfracloud.org/results/clime/rpkg-macros/fedora-$releasever-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/clime/rpkg-macros/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1

[copr:copr.fedorainfracloud.org:clime:preproc-rpmspec]
name=Copr repo for preproc-rpmspec owned by clime
baseurl=https://download.copr.fedorainfracloud.org/results/clime/preproc-rpmspec/fedora-$releasever-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/clime/preproc-rpmspec/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1
```

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -559,6 +559,7 @@ class Commands(object):
         try:
             self.uid_manager.becomeUser(self.buildroot.chrootuid, self.buildroot.chrootgid)
             self.state.start("buildsrpm")
+            host_chroot_sources = None
 
             # copy spec/sources
             shutil.copy(spec, self.buildroot.make_chroot_path(self.buildroot.builddir, "SPECS"))
@@ -567,14 +568,24 @@ class Commands(object):
                 # Resolve any symlinks
                 sources = os.path.realpath(sources)
                 if os.path.isdir(sources):
-                    file_util.rmtree(self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES"))
+                    host_chroot_sources = self.buildroot.make_chroot_path(
+                        self.buildroot.builddir, "SOURCES")
+                    file_util.rmtree(host_chroot_sources)
+
                     shutil.copytree(sources,
-                                    self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES"),
+                                    host_chroot_sources,
                                     symlinks=(not follow_links))
                 else:
-                    shutil.copy(sources, self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES"))
+                    host_chroot_sources = self.buildroot.make_chroot_path(
+                        self.buildroot.builddir, "SOURCES", os.path.basename(sources))
+                    shutil.copy(sources, host_chroot_sources)
 
             spec = self.buildroot.make_chroot_path(self.buildroot.builddir, "SPECS", os.path.basename(spec))
+
+            self.plugins.call_hooks('pre_srpm_build',
+                                    spec,
+                                    host_chroot_sources)
+
             # get rid of rootdir prefix
             chrootspec = spec.replace(self.buildroot.make_chroot_path(), '')
 
--- a/mock/py/mockbuild/plugins/rpkg_preprocessor.py
+++ b/mock/py/mockbuild/plugins/rpkg_preprocessor.py
@@ -1,0 +1,123 @@
+# -*- coding: utf-8 -*-
+# vim:expandtab:autoindent:tabstop=4:shiftwidth=4:filetype=python:textwidth=0:
+# License: GPL2 or later see COPYING
+# Written by clime
+# Copyright (C) 2020 clime <clime@fedoraproject.org>
+
+# python library imports
+import os
+import os.path
+import re
+import subprocess
+import configparser
+import shlex
+
+# our imports
+import mockbuild.util
+from mockbuild.trace_decorator import getLog, traceLog
+from mockbuild.exception import PkgError
+
+requires_api_version = "1.1"
+
+
+# plugin entry point
+@traceLog()
+def init(plugins, conf, buildroot):
+    RpkgPreprocessor(plugins, conf, buildroot)
+
+
+class RpkgPreprocessor(object):
+    """preprocess spec file by using rpkg utilities"""
+    # pylint: disable=too-few-public-methods
+    @traceLog()
+    def __init__(self, plugins, conf, buildroot):
+        self.buildroot = buildroot
+        self.config = buildroot.config
+        self.state = buildroot.state
+        self.opts = conf
+        self.log = getLog()
+        plugins.add_hook("pre_srpm_build", self._preprocess_proxy)
+        self.log.info("rpkg_preprocessor: initialized")
+
+    @traceLog()
+    def _install_requires(self, requires):
+        try:
+            self.buildroot.uid_manager.becomeUser(0, 0)
+            self.buildroot.pkg_manager.install(*requires, check=True)
+        finally:
+            self.buildroot.uid_manager.restorePrivs()
+
+    @traceLog()
+    def _preprocess_proxy(self, host_chroot_spec, host_chroot_sources):
+        if not host_chroot_sources or not os.path.isdir(host_chroot_sources):
+            self.log.debug("Sources not specified or not a directory. "
+                           "Skipping rpkg preprocessing step.")
+            return
+
+        self._preprocess(host_chroot_spec, host_chroot_sources)
+
+    @traceLog()
+    def _preprocess(self, host_chroot_spec, host_chroot_sources):
+        rpkg_conf_path = os.path.join(host_chroot_sources, 'rpkg.conf')
+
+        if not os.path.isfile(rpkg_conf_path):
+            self.log.info("rpkg.conf not found. "
+                          "Skipping rpkg preprocessing step.")
+            return
+
+        parser = configparser.ConfigParser(
+            interpolation=configparser.ExtendedInterpolation())
+
+        try:
+            parser.read(rpkg_conf_path)
+        except configparser.ParsingError as e:
+            raise PkgError("Parsing of %s failed with error: %s" % (rpkg_conf_path, repr(e)))
+
+        try:
+            preprocess_spec = parser.getboolean('rpkg', 'preprocess_spec')
+        except (configparser.Error, ValueError):
+            self.log.warning(
+                "Could not get boolean value of rpkg.preprocess_spec option from rpkg.conf.")
+            preprocess_spec = False
+
+        if not preprocess_spec:
+            self.log.info("preprocess_spec not enabled in rpkg.conf. "
+                          "Skipping rpkg preprocessing step.")
+            return
+
+        # try to locate spec file in SOURCES, which will be our input
+        host_chroot_sources_spec = os.path.join(host_chroot_sources,
+                                                os.path.basename(host_chroot_spec))
+
+        if not os.path.isfile(host_chroot_sources_spec):
+            raise PkgError("%s is not a file. Spec file needs to be among sources." %
+                           host_chroot_sources_spec)
+
+        self.log.info("Installing rpkg preprocessing requires...")
+        self._install_requires(self.opts.get('requires', []))
+
+        # get rid of host rootdir prefixes
+        rootdir_prefix = self.buildroot.make_chroot_path()
+        chroot_spec = host_chroot_spec.replace(rootdir_prefix, '')
+        chroot_sources = host_chroot_sources.replace(rootdir_prefix, '')
+        chroot_sources_spec = host_chroot_sources_spec.replace(rootdir_prefix, '')
+
+        command_str = self.opts.get('cmd') % {'source_spec': chroot_sources_spec,
+                                              'target_spec': chroot_spec}
+        command = shlex.split(command_str)
+
+        # determine whether to use private network or not based on rpmbuild_networking
+        private_network = (not self.config.get('rpmbuild_networking', False))
+
+        self.buildroot.doChroot(
+            command,
+            shell=False,
+            cwd=chroot_sources,
+            logger=self.buildroot.build_log,
+            uid=self.buildroot.chrootuid,
+            gid=self.buildroot.chrootgid,
+            user=self.buildroot.chrootuser,
+            unshare_net=private_network,
+            nspawn_args=self.config.get('nspawn_args', []),
+            printOutput=self.config.get('print_main_output', True)
+        )
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -318,6 +318,11 @@
 # the rpms in the results folder.
 # config_opts['plugin_conf']['sign_opts']['opts'] = '--addsign %(rpms)s -D "%%_gpg_name your_name" -D "%%_gpg_path /home/your_name/.gnupg"'
 
+# Enable preprocessing step before srpm build by using rpkg utilities
+# config_opts['plugin_conf']['rpkg_preprocessor_enable'] = False
+# config_opts['plugin_conf']['rpkg_preprocessor_opts']['requires'] = ['preproc-rpmspec']
+# config_opts['plugin_conf']['rpkg_preprocessor_opts']['cmd'] = '/usr/bin/preproc-rpmspec %(source_spec)s --output %(target_spec)s'
+
 #############################################################################
 #
 # environment for chroot
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -28,7 +28,7 @@ from .util import USE_NSPAWN, setup_operations_timeout
 PLUGIN_LIST = ['tmpfs', 'root_cache', 'yum_cache', 'mount', 'bind_mount',
                'ccache', 'selinux', 'package_state', 'chroot_scan',
                'lvm_root', 'compress_logs', 'sign', 'pm_request',
-               'hw_info', 'procenv', 'showrc']
+               'hw_info', 'procenv', 'showrc', 'rpkg_preprocessor']
 
 
 def nspawn_supported():
@@ -191,7 +191,11 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
         'compress_logs_opts': {
             'command': 'gzip',
         },
-
+        'rpkg_preprocessor_enable': False,
+        'rpkg_preprocessor_opts': {
+            'requires': ['preproc-rpmspec'],
+            'cmd': '/usr/bin/preproc-rpmspec %(source_spec)s --output %(target_spec)s',
+        },
     }
 
     config_opts['environment'] = {


mock/py/mockbuild/util.py
```diff
@@ -1202,6 +1202,11 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['rpmbuild_command'] = '/usr/bin/rpmbuild'
     config_opts['dnf_disable_plugins'] = ['local', 'spacewalk']
 
+    # rpkg preprocessing configuration options
+    config_opts['rpkg_preprocessing'] = False
+    config_opts['rpkg_preprocessing_requires'] = ['preproc-rpmspec']
+    config_opts['rpkg_preprocessing_cmd'] = '/usr/bin/preproc-rpmspec %(source_spec)s --output %(target_spec)s'
```
xsuchy: Instead of %() - can you use the built-in jinja2 expansion

mock/py/mockbuild/backend.py
```diff
@@ -536,6 +546,62 @@ def chain(self, args, options, buildroot):
...
+            self.buildroot.root_log.info(
+                "rpkg.conf not found. Skipping rpkg preprocessing step.")
+            return
+
+        parser = configparser.ConfigParser(
+            interpolation=configparser.ExtendedInterpolation())
+
+        try:
+            parser.read(rpkg_conf_path)
+        except Exception as e:
```
xsuchy: Does it really need to be wide general Exception? The same few lines below.

mock/py/mockbuild/util.py
clime: Hello, if I understand correctly, the jinja2 expansion can be only used to reference other configuration values. For example, config option `resultdir` uses `{{basedir}}` and `basedir` is defined earlier in `site-defaults.cfg`. But in this case, the values `source_spec` and `target_spec` cannot be completely preconfigured because they depend on name of the spec file being built into sprm which varies package by package. Am I understanding this correctly?

mock/py/mockbuild/backend.py
clime: Yes, I think I could be using `configparser.ParsingError` here and on the few lines below:
```
        try:
            preprocess_spec = parser.getboolean('rpkg', 'preprocess_spec')
        except:
            preprocess_spec = False
```
I can use more generic `configparser.Error`. I will prepare a patch. I will also fix the linting errors.

clime: Except for `configparser.Error`, I also needed to check for `ValueError` which is raised when `rpkg.preprocess_spec` is not a boolean. I also added a log message when we couldn't read the value from `rpkg.conf`. Please, see the new commits.

praiskup:
> If this is satisfied as well, the rpkg preprocessing requirements are
> installed into the chroot ...

This is somewhat fundamental problem in my opinion.

Why do you have to install rpkg-util inside the chroot, and evaluate everything
inside?  This makes mock much, much more brittle.  But it also makes the overall
spec preprocessor concept less useful (you can not guarantee that rpkg-util is
available say in opensuse chroots).

That being said, why not to run `rpkg-util` from host?

mock/py/mockbuild/util.py
praiskup: Yes.  Similar problem has arisen in sign plugin, so we went with #497.
You can use `{{ rpkg_preprocessing.spec }}`.  But it is less nice, especially
considering that we should move this into mock plugin IMO and that would
add yet another level in the dict ... I'm not completely sure about the plugin,
yet.

mock/py/mockbuild/backend.py
```diff
@@ -567,9 +636,21 @@ def buildsrpm(self, spec, sources, timeout, follow_links):
...
+                self.buildroot.root_log.info("Rpkg preprocessing enabled.")
+
+                if not host_chrootsources:
+                    self.buildroot.root_log.info(
+                        "Sources not present. Skipping rpkg preprocessing step.")
+                else:
+                    self.rpkgPreprocess(host_chrootspec, host_chrootsources, timeout)
+
             # get rid of rootdir prefix
-            chrootspec = spec.replace(self.buildroot.make_chroot_path(), '')
```
praiskup: If you did not `s/spec/host_chrootspec/`, the commit would be easier read.

mock/py/mockbuild/backend.py
```diff
@@ -567,9 +636,21 @@ def buildsrpm(self, spec, sources, timeout, follow_links):
                 else:
                     shutil.copy(sources, self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES"))
 
-            spec = self.buildroot.make_chroot_path(self.buildroot.builddir, "SPECS", os.path.basename(spec))
+                host_chrootsources = self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES")
```
praiskup: Having "chrootsources" in hand, you could pretty easily call something like
`call_hooks("preprocess_sources", chrootsources)`.  That way it would be much
easier to add any other source preprocessors.

clime:
Dne pá 27. bře 2020 16:39 uživatel Pavel Raiskup <notifications@github.com>
napsal:

> If this is satisfied as well, the rpkg preprocessing requirements are
> installed into the chroot ...
>
> This is somewhat fundamental problem in my opinion.
>
> Why do you have to install rpkg-util inside the chroot, and evaluate
> everything
> inside? This makes mock much, much more brittle. But it also makes the
> overall
> spec preprocessor concept less useful (you can not guarantee that
> rpkg-util is
> available say in opensuse chroots).
>
> That being said, why not to run rpkg-util from host?
>
Hello,

I am on my phone now so there might be some mistypes.

It's not actually rpkg-util which is being installed. Only rpkg-macros
(macro definitions), preproc (simple preprocessing engine) and
preproc-rpmspec which is a tiny wrapper around those two (it invokes
preproc-rpmspec with rpkg macros loaded).

There are, I guess, two reasons to do it inside chroot:

1) it is safer for users because macro evaluation is at least run in a
chrooted environment - probably the main reason, i think it would be
unexpected for users when mock would suddenly start to do a kind of "code
injection" evaluation directly on host.

2) user doesn't need to install more tooling on the host than already
present - additional tooling is installed only when needed and in chroot
when it is certain to be available. This has an advantage that people can
build the templated fedora/epel packages for fedora/epel chroots while
being e. g. on Mageia or Opensuse host. They only need to import the new
tooling if they want to start using it for their own chroots.

It's true that you can't then build the new fedora package (i.e package in
src.fp.o) for e.g. OpenSuse but the question is if people do it for other
than testing purposes.

Good thing is that this wouln't affect Copr which uses a different
host-based mechanism for building srpms!

That all being said, adoption of this tooling should be relatively easy
because the main requirements currently are: git or git-core, coreutils,
python (for preproc which written in python), bash but it will also take
time.

mock/py/mockbuild/util.py
clime: Alright, I thought the `{{ }}` braces can only be used to reference another configuration variable but they cannot be used to reference a variable, which is known only later during mock processing (when configuration has already been parsed). And when I looked into mock code, it didn't seem to me that you have some kind of special magic for it :). (Maybe I overlooked it).

mock/py/mockbuild/backend.py
clime: Yes, I know but I was getting confused myself with the variable names there so I thought I should make the names of variables more understandable.

mock/py/mockbuild/backend.py
clime: Right, but I think I would also need access to the specfile parameter (which was specified by `--spec`) and also access to the chroot environment. I guess this wouldn't be a problem to get to those variables as well, however.

clime: I mean access to the environment unless you will want me to switch to evaluation on host :).

clime:
Idk if I should say it now but probably yes. If this PR was accepted, I also wanted to add support for working with `.spec.rpkg` extension so that the `.rpkg` suffix is stripped when the spec is evaluated. This might require slight tweaks to the code, so I didn't want to do it in this PR because it could add some confusion but eventually, I would probably like to do it. I think people will want to differentiate between normal spec files and templated spec files by an extension (but I also think the extension shouldn't required). I would do it in a next PR unless you would like me to do it here.

mock/py/mockbuild/backend.py
praiskup: AFAIK plugins have everything important in hand, namely reference to buildroot object.

praiskup:
Ad 1) mock [doesn't provide security](https://github.com/rpm-software-management/mock/wiki#scope).  It
sort of protects users against accidental mistakes, only.  But you need to
trust the RPM/spec file and friends that there are good intentions.  As
such, there's no problem in running the rpkg-util stuff on host.  Calling
the rpkg util stuff inside chroot doesn't make sense from this POV.  Also,
I believe running rpkg-util stuff is safe even without chroot! :-)

Ad 2) I understand, and I don't want you to push to that (you just
complicate your life, and the code in plugin).  I don't think additional
Reuires: is anyhow a problem, wrap the code to `mock-rpkg-util`
sub-package and add any runtime dep you need.

> I also wanted to add support for working with .spec.rpkg extension so
> that the .rpkg

This code indeed doesn't belong to mock.  Mock reads src.rpm, spec files,
and can call pre-processor.  I'm all for the additional rpkg-util plugin.
But please don't complicate mock basecode.

IMO, what we need to find is proper place to call some hook(s), and
rpkg-util plugin should react on that hook(s) call(s).  Whether you want
to call the hook before or after (or both) the sources are copied to
chroot, it doesn't matter too much.  My guess though is that "before" it
is copied is much cleaner, so user are 100% sure what happened with the
source dir on host (and that both the host directory, and in-chroot
directory match each other).  Btw., note the difference between SOURCES
and BUILD directories.

Important thing, the code place where we look for **spec** file should
stay untouched - and we shouldn't try to find some random file suffixes.

Some small advices WRT proposal -- I wouldn't call the rpkg-util
constructs "macros".  RPM itself has (real) macros, or is macro language.
Not only we use wrong naming, but it looks weird to invent another macros
on top of macro language.  Also, use `rpkg-util` or `rpkg_util` prefixes
where you can, not just `rpkg`, that would make false assumption that
the configuration relates to `python-rpkg`.

praiskup:
Regarding my proposal, can you please:
1. move the logic to `plugins/rpkg_util.py` file, 
2. invent some hook call
3. optionally, add sub-package bringing the deps (depending on where you
plan to call the code
4. document it on wiki page
5. package the stuff to fedora, so we can merge the code?
ẂDYT?

clime:
> Ad 1) mock [doesn't provide security](https://github.com/rpm-software-management/mock/wiki#scope). It
> sort of protects users against accidental mistakes, only. But you need to
> trust the RPM/spec file and friends that there are good intentions. As
> such, there's no problem in running the rpkg-util stuff on host. Calling
> the rpkg util stuff inside chroot doesn't make sense from this POV. Also,
> I believe running rpkg-util stuff is safe even without chroot! :-)

At least the protection against accidental mistakes is useful. But I also feel that running the macro evaluation inside chroot is safer because it would take a non-trivial amount of effort to create an attacking sequence that would work. I wouldn't know how to do it, personally. Maybe after spending weeks or months on it but I don't know if I should spend time on that.

But generally, I consider `mock --buildsrpm` a safer/more secure operation than `fedpkg srpm` operation just for the reason that the second one is easy to take an advantage of.

> 
> Ad 2) I understand, and I don't want you to push to that (you just
> complicate your life, and the code in plugin). I don't think additional
> Reuires: is anyhow a problem, wrap the code to `mock-rpkg-util`
> sub-package and add any runtime dep you need.

Alright, it's possible but I would still prefer to do the evaluation in chroot if you don't completely disagree.

> 
> > I also wanted to add support for working with .spec.rpkg extension so
> > that the .rpkg
> 
> This code indeed doesn't belong to mock. Mock reads src.rpm, spec files,
> and can call pre-processor. I'm all for the additional rpkg-util plugin.
> But please don't complicate mock basecode.
> 
> IMO, what we need to find is proper place to call some hook(s), and
> rpkg-util plugin should react on that hook(s) call(s). Whether you want
> to call the hook before or after (or both) the sources are copied to
> chroot, it doesn't matter too much. My guess though is that "before" it
> is copied is much cleaner, so user are 100% sure what happened with the
> source dir on host (and that both the host directory, and in-chroot
> directory match each other). Btw., note the difference between SOURCES
> and BUILD directories.

I am okay with moving the stuff into a plugin. It seems the plugin mechanism is robust enough for it. I mean, I'll try it and in case I hit some issues, I would contact you to discuss them.

> 
> Important thing, the code place where we look for **spec** file should
> stay untouched - and we shouldn't try to find some random file suffixes.

Is there a place where we look for a spec file in mock by using `*.spec` wildcard? I am asking because I haven't found it.

I think my note about `.spec.rpkg` doesn't mean much for mock code. I probably mentioned it because I am still thinking whether it is a good idea to differentiate between templated spec files and normal spec files by an extension and I would like to collect some feedback :).

> 
> Some small advices WRT proposal -- I wouldn't call the rpkg-util
> constructs "macros". RPM itself has (real) macros, or is macro language.
> Not only we use wrong naming, but it looks weird to invent another macros
> on top of macro language. 

I don't know what else to call them because I understand them as macros. They are just a different kind of macros that are evaluated before rpm macros. I could call them tags but rpm also already has "tags". Idk about other possibilities.

> Also, use `rpkg-util` or `rpkg_util` prefixes
> where you can, not just `rpkg`, that would make false assumption that
> the configuration relates to `python-rpkg`.

Well, I would like to introduce the support for rpkg macros also to python-rpkg library so that `fedpkg local` or `fedpkg srpm` will work with the macros as well. I don't consider rpkg-macros to be specific for rpkg-util project. Maybe I could get an agreement that I can use the rpkg namespace here from rpkg maintainers first? I.e. I could contact Ondřej Nosek about it.

clime:
> Regarding my proposal, can you please:
> 
> 1. move the logic to `plugins/rpkg_util.py` file,
> 2. invent some hook call
> 3. optionally, add sub-package bringing the deps (depending on where you
>    plan to call the code)
> 4. document it on wiki page
> 5. package the stuff to fedora, so we can merge the code?
> 
> ẂDYT?

Sounds ok to me. Thank you for the approach you are taking here. I will try to move the code to plugin and in case I will hit some problems, I'll let you know. Documentation is not a problem.

The needed packages should already be both in Fedora and EPEL:
https://bodhi.fedoraproject.org/updates/?packages=preproc-rpmspec

I would probably prefer to call the plugin `plugins/rpkg.py` or `plugins/rpkg_preprocessing.py` (although I would prefer a shorter name). Maybe we could agree on this together with rpkg maintainers?

praiskup:
> Is there a place where we look for a spec file in mock by using *.spec wildcard?

I meant the place you edited.  That should stay basically untouched except for calling
the plugin hook.

> Alright, it's possible but I would still prefer to do the evaluation in chroot if you don't completely disagree.

I don't think it is wise choice, but it's up to you.  Think of `mock-scm`,
what we speak about here isn't much different.

> I think my note about .spec.rpkg doesn't mean much for mock code. I
> probably mentioned it because I am still thinking whether it is a good
> idea to differentiate between templated spec files and normal spec files
> by an extension and I would like to collect some feedback :).

Mock can not be responsible for reading anything else than final spec
file, ideally s/much/none/ code. :-)

My feedback wouldn't be positive here .. I'm not in favor of implementing
this in Fedora (at all), but I don't see any reason not to experiment here
if you'll maintain it.  Especially if it will bring possibility to use the same
logic for other spec preprocessors (the same hooks).

clime:
> > Is there a place where we look for a spec file in mock by using *.spec wildcard?
> 
> I meant the place you edited. That should stay basically untouched except for calling
> the plugin hook.
> 
> > Alright, it's possible but I would still prefer to do the evaluation in chroot if you don't completely disagree.
> 
> I don't think it is wise choice, but it's up to you. Think of `mock-scm`,
> what we speak about here isn't much different.
> 
> > I think my note about .spec.rpkg doesn't mean much for mock code. I
> > probably mentioned it because I am still thinking whether it is a good
> > idea to differentiate between templated spec files and normal spec files
> > by an extension and I would like to collect some feedback :).
> 
> Mock can not be responsible for reading anything else than final spec
> file, ideally s/much/none/ code. :-)
> 
> My feedback wouldn't be positive here .. I'm not in favor of implementing
> this in Fedora (at all),

I am sad to hear that. I think it's the best option for automatic release and changelog generation out there.

> but I don't see any reason not to experiment here
> if you'll maintain it. Especially if it will bring possibility to use the same
> logic for other spec preprocessors (the same hooks).

Fair enough, I still need to investigate the plugin mechanism. Hopefully, I won't hit any walls.

clime:
I've redone the solution completely to use the mock plugin mechanism. I tested it manually and it seems to work well. Please, take a look.

mock/py/mockbuild/backend.py
```diff
@@ -583,14 +584,25 @@ def buildsrpm(self, spec, sources, timeout, follow_links):
...
+                    host_chroot_sources = self.buildroot.make_chroot_path(
+                        self.buildroot.builddir, "SOURCES", os.path.basename(sources))
+                    shutil.copy(sources, host_chroot_sources)
 
             spec = self.buildroot.make_chroot_path(self.buildroot.builddir, "SPECS", os.path.basename(spec))
+
+            self.plugins.call_hooks('pre_srpm_build',
+                                    spec,
+                                    host_chroot_sources,
+                                    self.private_network)
```
praiskup, line 604: I'd drop the `self.private_network` parameter, you can use equivalent config `rpmbuild_networking`.

mock/py/mockbuild/plugins/rpkg_preprocessor.py
```diff
@@ -0,0 +1,120 @@
...
+        plugins.add_hook("pre_srpm_build", self._preprocess_proxy)
+        self.log.info("rpkg_preprocessor: initialized")
+
+    @traceLog()
+    def _install_requires(self, requires):
+        try:
+            self.buildroot.uid_manager.becomeUser(0, 0)
+            self.buildroot.pkg_manager.install(*requires, check=True)
+        finally:
+            self.buildroot.uid_manager.restorePrivs()
```
praiskup, line 48: Please take a look at `buildroot.preexisting_deps`.  Adding the packages there should be enough.

praiskup:
Please take a look at something like `https://github.com/rpm-software-management/mock/wiki/Plugin-rpkg-preprocess`

mock/py/mockbuild/backend.py
clime: I went for the `self.private_network` parameter because it is used everywhere in operations like `buildroot.doChroot`. I was worried that you may change the definition of `self.private_network` to be something else than `not config['rpmbuild_networking']` in future and the change wouldn't propagate to plugin.

praiskup: I wouldn't be afraid of that;  if that happened, there would have to be
compat fix for the plugin.  And making the plugin api as simple as
possible has preference here (the less parameters in hook the better).

mock/py/mockbuild/plugins/rpkg_preprocessor.py
clime: It doesn't seem to work for me. I tried to follow examples of other plugins like `ccache`. I looked at the code paths and it seems the `buildroot.preexisting_deps` is only used from `installSrpmDeps`, which doesn't seem to be invoked for `backend.py:buildsrpm` method.

mock/py/mockbuild/backend.py
clime: I added the new commit not to pass `private_network` into the plugin.

clime:
> Please take a look at something like `https://github.com/rpm-software-management/mock/wiki/Plugin-rpkg-preprocess`

Thank you, i'll try to write something up.

mock/py/mockbuild/plugins/rpkg_preprocessor.py
praiskup: Indeed, that seems to be a bug.  We should call it on universal place...
The `ccache` is not srpm dependency.  Do you want me to fix this first,
or can you write a patch also for this?  The patch should be applied before
this patch.

clime: I would prefer to finish this pr in some working state and perhaps do some optimizations later.

clime:
The last build failed, but I think it might now be my fault: https://travis-ci.com/github/rpm-software-management/mock/jobs/319189905

clime:
I wrote a description of the plugin as you asked me here: https://github.com/rpm-software-management/mock/wiki/Plugin-rpkg-preprocessor

Can you, please, review?

There is a link to rpkg macros library but the documentation on that link is for an older version of `rpkg-macros`. I am going to update the documentation and fix the link but if you can give me feedback to the text meanwhile, it would be great!

Thanks

clime:
Do you know if I can make a non-breaking space in the wiki edit window?

praiskup:
> I wrote a description of the plugin as you asked me here: https://github.com/rpm-software-management/mock/wiki/Plugin-rpkg-preprocessor
> Can you, please, review?

Please don't speak about macros there.  It is confusing, (a) the `{{{ }}}`
aren't macros (macro language keeps re-expanding, or is it?), and when we
speak in RPM/Mock world about macros, `%macro` syntax is what have people
immediately in mind (if you need macros, why on world would you invent another
macro language instead of reusing the existing one)..  This always blows my
mind when I see it.  Otherwise looks fine, thank you for writing it!

mock/py/mockbuild/backend.py
```diff
@@ -583,14 +584,24 @@ def buildsrpm(self, spec, sources, timeout, follow_links):
...
+
+                    util.rmtree(host_chroot_sources)
                     shutil.copytree(sources,
-                                    self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES"),
+                                    host_chroot_sources,
                                     symlinks=(not follow_links))
                 else:
-                    shutil.copy(sources, self.buildroot.make_chroot_path(self.buildroot.builddir, "SOURCES"))
+                    host_chroot_sources = self.buildroot.make_chroot_path(
+                        self.buildroot.builddir, "SOURCES", os.path.basename(sources))
```
praiskup, line 596: I'm missing something here.  Can you document why `SOURCES` and now you need one more field `os.path.basename(sources)`?

mock/py/mockbuild/backend.py
```diff
@@ -583,14 +584,24 @@ def buildsrpm(self, spec, sources, timeout, follow_links):
...
+                    shutil.copy(sources, host_chroot_sources)
 
             spec = self.buildroot.make_chroot_path(self.buildroot.builddir, "SPECS", os.path.basename(spec))
+
+            self.plugins.call_hooks('pre_srpm_build',
+                                    spec,
+                                    host_chroot_sources)
```
praiskup, line 603: This looks like nice API.

clime:
> > I wrote a description of the plugin as you asked me here: https://github.com/rpm-software-management/mock/wiki/Plugin-rpkg-preprocessor
> > Can you, please, review?
> 
> Please don't speak about macros there. It is confusing, (a) the `{{{ }}}`
> aren't macros (macro language keeps re-expanding, or is it?), and (b)when we
> speak in RPM/Mock world about macros, `%macro` syntax is what have people
> immediately in mind (if you need macros, why on world would you invent another
> macro language instead of reusing the existing one).. This always blows my
> mind when I see it. Otherwise looks fine, thank you for writing it!

I don't have better name to call them. `{{{ }}}` are just tags but what is inside are macros - i.e. predefined set of functions that do something. They are not rpm macros, but rpkg macros (or rpkg-util macros). I really don't see what else to call them.

mock/py/mockbuild/backend.py
clime: I think this else branch is for case when the `sources` is a tarball. Then we want to maintain the tarball name. `shutil.copy` in that case would also maintain the tarball name.

mock/py/mockbuild/backend.py
clime: Thanks.

clime:
Also note that I am using the term macros for them everywhere in docs and in man pages - it would be just very confusing to use something else here.

mock/py/mockbuild/backend.py
praiskup: > I think this else branch is for case when the sources is a tarball.

Yes.

> Then we want to maintain the tarball name. shutil.copy in that case
> would also maintain the tarball name.

I see, `shutil.copy('/file', '/dir')` is equivalent to
`shutil.copy('/file', '/dir/file')`, and you want to pass down tarball
name to the `pre_srpm_build`.

Why to do that?  It means that every `pre_srpm_build` needs to
check whether the path points to filename or directory because it is
uncertain.

Could we make it consistent, and always pass down to path to sources
directory?  If not, can you document the reasons?

praiskup:
> I don't have better name to call them.

Actually, anything else is better than "macro".  Perhaps variable, template,
procedure, callback, ... when I hear macro, I hear `cpp`, `m4`, `rpm`, `tex`,...

>  I am using the term macros for them everywhere in docs and in man pages

Yeah, I supposed that you could fix that misconcept too.  I'm stopping here..
it's your design so you should decide.  Can you please provide some link to
what "rpkg macro" means?  We change the first use of term "macro" into
hyperlink pointing to the definition.
page.

mock/py/mockbuild/backend.py
clime: Well, i think it depends on what is being passed to `--sources <param>`. Whether a file or a subdirectory. I can for sure document this but can you point me where, please?

praiskup: Well, I mean ... why do you want to let `pre_srpm_build` depend on the
argument that way when we can design more clear API?  I don't see any
benefits, only negatives (I'd prefer passing down the path to SOURCES dir
always).

But if there _are_ some benefits or needs which are not really obvious,
we have to document so we don't have to ask this question again.

clime: I am not really sure if there is a significant benefit. A potential benefit could be that if user specifies a specific tarball and there is an srpm preprocessing plugin, which does certain preprocessing on tarballs (i.e. license checks) or that reads headers from the tarball to use them for spec preprocessing, then it is handy to know what tarball was specified on command-line by user because mock afaik doesn't really cleanup SOURCES so there might be more tarballs for some other packages too at this point. By not passing the source file user specified explicitly, we could therefore potentially lose some information. I would probably prefer just a comment in documentation of the hooks (if there is any) but let me know what you think and if this is ok with you. rpkg_preprocessor currently doesn't need that use-case when a user specifies `--sources <file>`.

clime:
> > I don't have better name to call them.
> 
> Actually, anything else is better than "macro". Perhaps variable, template,
> procedure, callback, ... when I hear macro, I hear `cpp`, `m4`, `rpm`, `tex`,...
> 
> > I am using the term macros for them everywhere in docs and in man pages
> 
> Yeah, I supposed that you could fix that misconcept too. I'm stopping here..
> it's your design so you should decide. Can you please provide some link to
> what "rpkg macro" means? We could change the first use of term "macro" into
> hyperlink pointing to the definition..

Yes, I will do it. I am still working on documentation and I would like to have it completed before this is merged. I have found out that I should change the way in which the rpkg "macros" sort annotated tags, which will require release of a new version. I will write here when everything is ready for merge but meanwhile we can work on other details if there are any. Thank you for your time you invested into this so far. I appreciate it a lot.

mock/py/mockbuild/backend.py
praiskup: Thanks for the response,

> By not passing the source file user specified explicitly, we could
> therefore potentially lose some information.

When anyone ever needed the additional info, I think we could add
additional argument, --source argument could make a note in buildroot
object, or something like that.

>  rpkg_preprocessor currently doesn't need that use-case when a user
>  specifies --sources <file>.

Cool, can you please make the API simpler then, and always pass the source
directory path down to the hook?

praiskup:
>  I will write here when everything is ready for merge

ACK

praiskup:
> I will write here when everything is ready for merge

How is it going?

clime:
> > I will write here when everything is ready for merge
> 
> How is it going?

Hey, sorry for the delay. I've finished the docs: https://docs.pagure.org/rpkg-util/v3/macro_reference.html. I now need to update packaging of the components, then do final testing and then push updates to the repos. The push of the components might take some time (i will also need FeSCo exception).

clime:
Hello, in the recent commits, I have rebased to the current master and fixed two problems in code:

1) fixed reference to `self.config` when getting `rpmbuild_networking` value
2) included `rpkg_preprocessor` in PLUGINS list so that it gets default values specified in `utils.py`

I have also released `rpm-git-tag-sort-1.0` and opened https://pagure.io/fesco/issue/2488 to get an exception to update `rpkg-macros` and `preproc-rpmspec` to their latest versions. If I get the exception and the process is completed, I think this is ready for merge. I tested that the plugin works with the newest packages in epel7 and fedora32. I will probably do another, final testing when everything is really ready.

clime:
The build failed but I don't see why. Please, let me know if you have that info.

praiskup:
Thanks for the update.

> The build failed but I don't see why. Please, let me know if you have that info.

I think I should remove fedora-33-i386 from CI - since it is not mirrored it is more error-prone:
https://download.copr.fedorainfracloud.org/results/@mock/mock-pull-requests/fedora-33-i386/01700214-mock/

clime:
Small update: i've got the fesco exception and the required packages are now waiting to get to stable (e.g. https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2020-90d5176dcb)

Looks like some conflicts appeared meanwhile. Will try to sort them about before the final testing. Thanks for the patience.

clime:
Hello!

I've rebased the commits and tested they work as expected with the current packages in EPEL and fedora releases. I believe this is finally ready for merge...

There is a build failure: https://travis-ci.com/github/rpm-software-management/mock/jobs/443928347 but I think it might rather be an infrastructure issue.

clime:
I've rebased once more as I didn't have the right committer set on the commits (`git config user.name` and `user.email`).

praiskup:
Thanks, can you please squash (or allow me to do so) and rebase once more?   Otherwise looks good.

clime:
> Thanks, can you please squash (or allow me to do so) and rebase once more? Otherwise looks good.

Done

praiskup:
Thank you!

praiskup:
Merged as bda814a743ef74e6dfbee261a4db6e2f219382d1

Issue #183, opened 2018-05-07
Tmpfs Plugin does not Umount correctly.


jmcgrath207:
### Short description of the problem

Keep on Getting a `Device is Busy` When using  the Tmpfs Plugin

### Output of `rpm -q mock`
Mock Version 1.4.9 on Centos 7

### Any additional notes
From the output of `df -h` it looks like It creates 2  tmpfs mounts when running but after  the command executes  it only removes one.

### During Run

```
[vagrant@BuildUtil ~]$ df -h
Filesystem                       Size  Used Avail Use% Mounted on
/dev/mapper/VolGroup00-LogVol00   38G  2.4G   36G   7% /
devtmpfs                         1.4G     0  1.4G   0% /dev
tmpfs                            1.5G     0  1.5G   0% /dev/shm
tmpfs                            1.5G  8.4M  1.4G   1% /run
tmpfs                            1.5G     0  1.5G   0% /sys/fs/cgroup
/dev/sda2                       1014M   63M  952M   7% /boot
tmpfs                            287M     0  287M   0% /run/user/1000
mock_chroot_tmpfs                2.0G  973M  1.1G  48% /var/lib/mock/epel-6-x86_64/root
tmpfs                            1.5G     0  1.5G   0% /var/lib/mock/epel-6-x86_64/root/dev/shm
```

### After Run

```
[vagrant@BuildUtil ~]$ df -h
Filesystem                       Size  Used Avail Use% Mounted on
/dev/mapper/VolGroup00-LogVol00   38G  2.4G   36G   7% /
devtmpfs                         1.4G     0  1.4G   0% /dev
tmpfs                            1.5G     0  1.5G   0% /dev/shm
tmpfs                            1.5G  8.4M  1.4G   1% /run
tmpfs                            1.5G     0  1.5G   0% /sys/fs/cgroup
/dev/sda2                       1014M   63M  952M   7% /boot
tmpfs                            287M     0  287M   0% /run/user/1000
mock_chroot_tmpfs                2.0G     0  2.0G   0% /var/lib/mock/epel-6-x86_64/root
```

Consider adding output of `mock --debug-config` this can help developers to reproduce the issue.

```
[vagrant@BuildUtil ~]$ sudo mock --debug-config -r Util_mock_config_epel-6-x86_64.cfg --buildsrpm --spec Util.spec --sources . --resultdir=.
INFO: mock.py version 1.4.9 starting (python version = 2.7.5)...
Start: init plugins
[...]

PR #534, opened 2020-03-13
Call orphansKill also "postyum" time

praiskup:
Some packages may keep leftover processes on background after it's
installation, triggered by scriptlets.

Even though this is definitely packaging issue (bug in package) the
background processes keep using the chroot's root filesystem and refuse
us to umount the filesystem later (e.g. when tmpfs is enabled).

Since the dnf commands are not run under systemd-nspawn (when bootstrap
is disabled), we need to add new option which ignores the USE_NSPAWN
flag.

Fixes: #183

### DIFF
--- a/mock/integration-tests/001-orphanskill-explicit.tst
+++ b/mock/integration-tests/001-orphanskill-explicit.tst
@@ -6,27 +6,38 @@
 # Test orphanskill feature (explicit)
 #
 header "Test orphanskill feature (explicit)"
+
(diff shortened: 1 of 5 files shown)


praiskup:
I am trying to prepare a testcase for this;  but for some reason my daemon processes
aren't able survive %posttrans scriptlet runtime...   So once I have test, I'll merge this.
Stop me if you disagree with the code.

Issue #447, opened 2020-01-13
Can't build rpm for CentOS 8 using Docker


karljohns0n:
### Short description of the problem

I use CentOS 7 docker image to test my builds for CentOS 6/7/8 and mock fail to build el8 packages (using docker only):

```
INFO: Mock Version: 1.4.21
Start: dnf update
Finish: dnf update
Finish: chroot init
Start: build phase for aeris-release-1.0-6.el8.src.rpm
Start: build setup for aeris-release-1.0-6.el8.src.rpm
Finish: build setup for aeris-release-1.0-6.el8.src.rpm
Start: Outputting list of installed packages
ERROR: Exception(/tmp/aeris-release/SRPMS/aeris-release-1.0-6.el8.src.rpm) Config(aeris-8-x86_64) 0 minutes 52 seconds
INFO: Results and/or logs in: /tmp/aeris-release/RPMS
ERROR: state finish mismatch: current: Outputting list of installed packages, state: build phase for aeris-release-1.0-6.el8.src.rpm

The only errors in root.log:

DEBUG util.py:689:  Executing command: rpm -qa --root '/var/lib/mock/aeris-8-x86_64/root' --qf '%{nevra} %{buildtime} %{size} %{pkgid} installed\n' > /tmp/aeris-release/RPMS/installed_pkgs.log with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'} and shell True
DEBUG util.py:689:  Executing command: rpm -qa --root '/var/lib/mock/aeris-8-x86_64/root' --qf '%{nevra} %{buildtime} %{size} %{pkgid} installed\n' > /tmp/aeris-release/RPMS/installed_pkgs.log with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'} and shell True
[...]

PR #535, opened 2020-03-13
Run several stuff from bootstrap instead of from host, if appropriate

praiskup:
~~This is random thought, not a material for review ... I just don't want to loose the idea so opening PR for now.~~

### DIFF
--- a/mock/docs/mock.1
+++ b/mock/docs/mock.1
@@ -104,8 +104,18 @@ the specified destination path.
 Copies the source paths (files or directory trees) from the chroot to
 the specified destination path.
 .TP
-\fB\-\-chroot\fP \fICOMMAND [\fIARGS...\fR]
-Run the specified command non\-interactively within the chroot (no 'clean' is performed). Command output will be sent to the log files. Note that COMMAND is shell expanded only if there are no ARGS. Otherwise, there is no shell expansion.
+\fB\-\-chroot\fP [\fI\-\-\fR] \fICOMMAND [\fIARGS...\fR]
+Run the specified command non\-interactively within the chroot (no
+\fB\-\-clean\fR is performed).
+
+This mode is similar to \fishell\fR mode, except that the output \fBis logged\fR
+and the \fICOMMAND\fR and \fIARGS\fR are not shell expanded in chroot when
+the variant with \fIARGS\fR is used, see the difference:
+    \fI$ \fBmock --quiet --chroot -- echo '*'\fR
+    *
+    \fI$ \fBmock --quiet --chroot -- 'echo *'\fR
+    bin boot builddir dev etc home lib lib64 media mnt opt ...
+See also \fB\-\-shell\fR.
 .TP
 \fB\-\-debug-config\fP
 Print all options in config_opts.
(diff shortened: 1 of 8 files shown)


praiskup:
Anyone, please take a look.  Some of those patches (better to review
separately) are very important with upcoming changes in Fedora, and some
are less important but were somewhat natural to fix when I was working the
additional `doOutChroot` abstraction.

praiskup:
Potentially related to #447, but more likely just 60b2b34f8f325d0b90dbfc7d0d294a4d1edac155.

mock/docs/mock.1
```diff
@@ -104,8 +104,13 @@ the specified destination path.
...
 the specified destination path.
 .TP
-\fB\-\-chroot\fP \fICOMMAND [\fIARGS...\fR]
-Run the specified command non\-interactively within the chroot (no 'clean' is performed). Command output will be sent to the log files. Note that COMMAND is shell expanded only if there are no ARGS. Otherwise, there is no shell expansion.
+\fB\-\-chroot\fP [\fI\-\-\fR] \fICOMMAND [\fIARGS...\fR]
+Run the specified command non\-interactively within the chroot (no
+\fB\-\-clean\fR is performed).
+
+This mode is similar to \fishell\fR mode, except that the output \fBis logged\fR
+and the \fICOMMAND\fR and \fIARGS\fR are not shell expanded in chroot.  See also
```
xsuchy: This is the case I spoke about. Some people still expects that `--chroot 'ls *'` will be shell expanded in chroot.

praiskup: And it stays expanded...  but the documentation fix is wrong.  Right?

praiskup: I updated it to:
```
       --chroot [--] COMMAND [ARGS...]
              Run the specified command non-interactively  within  the  chroot
              (no --clean is performed).

              This  mode  is  similar to shell mode, except that the output is
              logged and the COMMAND and ARGS are not shell expanded in chroot
              when the variant with ARGS is used, see the difference:
                  $ mock --quiet --chroot -- echo '*'
                  *
                  $ mock --quiet --chroot -- 'echo *'
                  bin  boot  builddir dev etc home lib lib64 media mnt opt ...
              See also --shell.

```

mock/py/mockbuild/backend.py
```diff
@@ -179,17 +179,30 @@ def remove(self, *rpms):
         output = self.buildroot.pkg_manager.remove(*rpms, returnOutput=1)
         self.buildroot.root_log.info(output)
 
+    @traceLog()
+    def getPreconfiguredDeps(self, srpms):
```
xsuchy, line 183: Can I ask here for docstring, please? I do not get from the name what preconfigured deps are.

mock/py/mockbuild/util.py
```diff
@@ -714,6 +714,7 @@ def do_with_status(command, shell=False, chrootPath=None, cwd=None, timeout=0, r
             command = _prepare_nspawn_command(chrootPath, user, command,
                                               nspawn_args=nspawn_args,
                                               env=env, cwd=cwd, shell=shell)
+            env['SYSTEMD_NSPAWN_TMPFS_TMP'] = '0'
```
xsuchy, line 717: can we rather pass it to _prepare_nspawn_command. Or even better - set this inside of _prepare_nspawn_command().

mock/py/mockbuild/util.py
```diff
@@ -892,7 +880,13 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None,
...
         # EL < 7.5 does not support the nspawn -a option. See BZ 1417387
         nspawn_argv += ['-a']
-    nspawn_argv.extend(nspawn_args)
+
+    if user:
+        # user can be either id or name
+        nspawn_argv += ['-u', str(user)]
+
+    if nspawn_args:
+        nspawn_argv += nspawn_args
```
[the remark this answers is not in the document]
praiskup: Yeah, makes sense (nspawn_args can not be None).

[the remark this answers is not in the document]
praiskup: Well, can be None, so the check needs to stay ... I'll flip to extend then, the change
wasn't intentional, even though it shouldn't make any change (not even performance) here.

mock/py/mockbuild/util.py
praiskup: I can rewrite this to return tuple, `command` + `env_overrides`.  I'm not sure
it is wise to edit `env` that is passed by reference ... wdyt?

mock/py/mockbuild/util.py
```diff
@@ -903,12 +897,15 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None,
...
-        return cmd
-    else:
-        return " ".join(cmd)
 
+    # /bin/sh -c wrapper explicitly requested
+    if shell and isinstance(cmd, list):
+        cmd = ' '.join(cmd)
+
+    if isinstance(cmd, str):
+        cmd = ['/bin/sh', '-c', cmd]
```
xsuchy, line 907: ??? this is not based on `shell` variable?

praiskup: Yes, that's the hack needed for `--chroot 'one command'`, which needs to stay expanded.
And we always drop the `shell=True` after calling _prepare_nspawn ... so :-( this is 
not something I'm happy about either.

praiskup: I'll at least document the reasons here.

praiskup:
merged manually as 60f7bf750b6c26c672ddfb61cb3e76275293838f -> 42df0034f0046fb84f989af83a9d2f2aa7085901.

mock/py/mockbuild/util.py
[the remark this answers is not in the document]
xsuchy: Yes, after your change it can be None. But originally it is checked at the start of function.

mock/py/mockbuild/util.py
xsuchy: What about setting this inside of _prepare_nspawn_command?

mock/py/mockbuild/util.py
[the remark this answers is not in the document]
praiskup: Indeed, there was before:
```
    if nspawn_args is None:
    nspawn_args = []
```
... with `nspawn_args=None` default.

mock/py/mockbuild/util.py
praiskup: That would mean passing `env` down into _prepare_nspawn_command
by reference, and editing it there.  Perhaps yes, ... I wanted to do this but
then I realized that we have to flip `shell` from `True` to `False` after calling
this method anyways - and we can not pass immutable down by reference
(bool).  So to make it nice, it only makes sense to fix both env and shell 
variables.

PR #536, opened 2020-03-13
mock: use 'private' mount option, not 'rprivate'

praiskup:
The 'mount --rbind -o rprivate' on EL7 does weird things;  it looks like
it actually reverses the '--rbind' option and does only '--bind'.  Using
just 'private' seems to do the same (good) thing on both EL7 and modern
Fedora.

This fixes yum_chache plugin on EL7.

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -747,15 +747,15 @@ def main():
         # add the extra bind mount to the outer chroot
         inner_mount = bootstrap_buildroot.make_chroot_path(buildroot.make_chroot_path())
 
-        # Hide re-mounted chroot from host by rprivate tmpfs.
+        # Hide re-mounted chroot from host by private tmpfs.
         buildroot.mounts.managed_mounts.append(
             FileSystemMountPoint(filetype='tmpfs',
                                  device='hide_root_in_bootstrap',
                                  path=inner_mount,
-                                 options="rprivate"))
+                                 options="private"))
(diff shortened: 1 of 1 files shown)


xsuchy:
I believe this was introduced as PR to make this work under Fedora Toolbox. Check discussion in that PR.

praiskup:
Hmm, I don't think we discussed `private` vs. `rprivate`.  I'll take a look...

praiskup:
I think that we discussed `--bind` vs `--rbind`, but we talked about `/proc` and `/sys` mountpoints.  Not about the `/root -> /bootstrap/root` mountpoint.  WDYT?

praiskup:
Hm, it actually looks like we never tested bootstrap chroot with toolbox.

praiskup:
Ok, I tested:
`$ mockchain --old-chroot --no-bootstrap-chroot python-copr-999-1.src.rpm dep-on-python-copr-999-1-0.src.rpm ` from [testing repo](https://github.com/rpm-software-management/mock-test-data) and it succeded in toolbox:
```
...
INFO: Packages built: 2
INFO: Packages successfully built in this order:
INFO: python-copr-999-1.src.rpm
INFO: dep-on-python-copr-999-1-0.src.rpm
```

praiskup:
Merged as 34fe0d5b80a12d3a7ebdcbe4f977dbf136dc54fa

PR #538, opened 2020-03-14
Add the Devel repo for CentOS 8 and CentOS Stream

### LINKED ISSUE #537: Add new CentOS 8 "devel" repo
RHEL 8 does not ship certain devel packages, such as "quota-devel", which they use to compile other software. CentOS is now publishing a "devel" repo for these.

### Output of `rpm -q mock`

1.4.21.1

### Steps to reproduce issue

mock install quota-devel.

### Any additional notes

The new repo is at http://mirror.centos.org/centos/8/Devel/x86_64/os/Packages/ . I expect it to be added ot the CentOS mirrors ASAP, at which point it should be added to the centos-8 template. I'll be happy to ship a pull request as soon as the mirrors are published.

Conan-Kudo:
This repository is disabled by default because we should offer
equivalent repository sources to RHEL by default and there is no
guarantee on the usability of the content shipped here.

In addition, a trivial commit adding the PowerTools source repo
(disabled like other source repos) has been included for completeness.

Fixes #537.

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream.tpl
@@ -89,6 +89,13 @@ gpgcheck=1
 enabled=0
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 
+[Stream-PowerTools-source]
+name=CentOS-Stream - PowerTools Sources
+baseurl=http://vault.centos.org/centos/8-stream/PowerTools/Source/
+gpgcheck=1
+enabled=0
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+
 [Stream-extras-source]
 name=CentOS-Stream - Extras Sources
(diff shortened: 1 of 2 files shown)


mock-core-configs/etc/mock/templates/centos-stream.tpl
```diff
@@ -75,6 +75,13 @@ gpgcheck=1
 enabled=1
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 
+[Stream-Devel]
+name=CentOS-Stream - Devel WARNING! FOR BUILDROOT USE ONLY!
```
praiskup: Do you want to keep the large warning here?  Saying "Buildroot only." would be appropriate to me :-).  Otherwise no problem, ... no strong objection, the PR looks good to me.

Conan-Kudo: I copied it from [the official repo file](https://git.centos.org/rpms/centos-release/blob/c8/f/SOURCES/CentOS-Devel.repo), so I think it's appropriate to leave as-is.

PR #544, opened 2020-03-25
mock: conflict with old core-configs

praiskup:
.. instead of Require'ing them.  Since core-configs are optional thing,
this is acceptable use of Conflicts tag:
https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts/#_optional_functionality

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -25,7 +25,14 @@ Requires: usermode-consoleonly
 Requires: usermode
 %endif
 Requires: createrepo_c
-Requires: mock-core-configs >= 32.4
+
+# We know that the current version of mock isn't compatible with older variants,
+# and we want to enforce automatic upgrades.
+Conflicts: mock-core-configs < 32.4
+
+# Requires 'mock-core-configs', or replacement (GitHub PR#544).
+Requires: mock-configs
+
 Requires: systemd
 %if 0%{?fedora} || 0%{?rhel} >= 8
 Requires: systemd-container
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -16,6 +16,10 @@ URL:        https://github.com/rpm-software-management/mock/
 Source:     https://github.com/rpm-software-management/mock/releases/download/%{name}-%{version}-1/%{name}-%{version}.tar.gz
 BuildArch:  noarch
 
+# The mock.rpm requires this.  Other packages may provide this if they tend to
+# replace the mock-core-configs.rpm functionality.
+Provides: mock-configs
+
 # distribution-gpg-keys contains GPG keys used by mock configs
 Requires:   distribution-gpg-keys >= 1.36
 
--- a/mock/docs/release-instructions.txt
+++ b/mock/docs/release-instructions.txt
@@ -29,7 +29,7 @@ Release checklist overview:
    $ tito tag
    When you release both mock and mock-core-configs together, you
    likely want to (a) first tag 'mock-core-configs' package, (b) bump
-   'Requires: mock-core-configs >= ??' dependency and (c) then tag
+   'Conflicts: mock-core-configs < ??' in mock.spec and (c) then tag
    new mock version.
 5) push to main git repo (only from master branch):
    $ git push


xsuchy:
Than you want to at least Recommends those configs. Otherwise people will complain that they install mock package and does not have configs.

praiskup:
Absolutely, thanks for taking a look.

praiskup:
Well, but I wanted to achieve different thing in this PR, see updated commit message:
```
     mock: conflict with old core-configs
    
    We want to make the mock-core-configs package "optional", in a sense
    that users may replace it with their own package.  This should be
    acceptable use of Conflicts tag:
    https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts/#_optional_functionality
    
    Keep the Requires tag -- as probably all users expect to have mock
    configured by default.  And also to not cause wave of updates in all the
    packages that depend on mock.  To ease the `mock-core-configs`
    replacement, flip the 'Requires:' against new explicit 'Provides:'
    from mock-core-configs that we don't Conflict with.
    
    Fixes: #544
```

praiskup:
@xsuchy, can you please have another look?

mock/mock.spec
```diff
@@ -25,7 +25,14 @@ Requires: usermode-consoleonly
...
 %endif
 Requires: createrepo_c
-Requires: mock-core-configs >= 32.4
+
+# We know that the current version of mock isn't compatible with older variants,
+# and we want to enforce automatic upgrades.
+Conflicts: mock-core-configs < 32.4
+
+# Requires 'mock-core-configs', or replacement (GitHub PR#544).
+Requires: mock-configs
```
Conan-Kudo: What do you expect to replace this with?

praiskup: There are few mock users that don't use mock-core-configs.  E.g. koji
and covscan.  They typically want to provide an alternative to this (so
why not to allow them to easily package their configs).  I expect something
like `koji-configs.rpm` which does `Provides: mock-configs`.

Conan-Kudo: Koji generates its configs on the fly, though? There are no configs on-disk for Koji.

praiskup: But they could.  Even if that was just set of few templates useful for further
koji config generator.

praiskup: TBH, I came to this topic because of covscan project;  but I could speak
at least for copr.  We keep doing manual updates of mock-core-config files all
the time, as the buildroots evolve ...  We could provide `copr-mock-configs`
for copr purposes (instead of mock-core-configs) and easily tag that into
infra-tags - without being worry that we will break anyone else.

praiskup: That said, does it make sense that someone might want to not install mock-core-configs,
and this is just compromise to allow them to do that?

Conan-Kudo: What would be different from `copr-mock-configs` from `mock-core-configs`? Generally the updates you've been making help _everyone_ using mock...

praiskup: Ideally no difference;  but practically we need sometimes to ship some change
to copr earlier.

But you are nitpicking on deployment specific problem.  Does this PR make any
sense in general?  Does it make any sense to force installation of mock-core-configs
on koji builders?

Conan-Kudo: I think there's no scenario where we would not want `mock-core-configs` installed with `mock`. The separation of the two was mainly to allow shipping updated configs without cutting a new mock release.

praiskup: What do you want `mock-core-configs` for on koji builder?

Conan-Kudo: Does it matter if it's there or not for Koji builders?

praiskup: It shouldn't.  Somewhat semantically it does, because koji could use
mock-core-configs by mistake.  It also goes against the minimization
efforts all over the place.

But I'm using this example to convince you
that mock is independent software and it _may_ have sense to have
mock installed without _default_ configuration.

Is there any other practical problem with this PR?  I mean, nothing
changes for anyone after merging this, except for those who really
want to override the configs.  Or?

Conan-Kudo: I'd add a `Suggests: mock-core-configs` to that, but yeah, otherwise it's fine.

praiskup: Frankly, covscan is pretty important customer of mock - so even if
we only helped those people this patch makes sense.  There's nothing
worse than seeing anyone re-building package you maintain with
downstream patches. :-/

Conan-Kudo: Who is `covscan`?

praiskup: s/covscan/csmock/ -> tooling for automatic static analysis (using mock):
https://github.com/kdudka/csmock

praiskup: > I'd add a Suggests: mock-core-configs to that, but yeah, otherwise it's fine.

I missed this message, ... why, when we have the Requires: there?

Conan-Kudo: You are depending on a virtual name, and having the Suggests means that `mock-core-configs` is preferred, but can be substituted by anything else, too.

praiskup: Ok, I just have to test that resolver accepts this (suggesting one package, but
installing other, conflicting package).

praiskup: I tested this.  When I did `dnf install mock test-configs`, dnf did not complain about
the Suggests at all.

Issue #546, opened 2020-03-25
move 'site-defaults' to %doc


praiskup:
.. and perhaps keep it as %config file in mock-core-configs, with content
like:

  ```
  # Global (non-user specific) mock configuration.
  # 
  # Please consult configuration examples from
  # /usr/share/doc/mock*/defaults.txt
  # and consider using `$HOME/.config/mock.cfg` file instead if possible.
  ```

I'm not sure.  People are continuously using site-defaults.cfg for
things that should be done in `$HOME/.config/mock.cfg`.  The fact they are
using site-defaults.cfg is often because it is marked %config in spec file
-- and that it is "pre-configured" -- so it is easier for initial tweaks.  But
then, when we update the file (happens all the time, for documentation
purposes) they need to merge `.rpmnew` files ...  and perhaps they never
do that ... so they don't know the new options, etc.

Conan-Kudo:
@praiskup maybe it's better to move it to `/usr/share/mock` and make `/etc/mock/site-defaults.cfg` empty?

praiskup:
Since the site-defaults.cfg isn't anythin else than documentation (and we
don't plan to add there uncommented configs), why `/usr/share/mock` and
not `/usr/share/doc/mock`?

amessina:
site-defaults.cfg has been useful for someone like me, running a private Koji instance, where not all of mock's config options are easy to pass into koji. As long as there can be a drop in where /etc/site-defaults.cfg overrides the one /usr/share, that would help.

praiskup:
Ok, probably understand why we discuss `/usr/share/mock` directory;  but
that is not yet used implemented in mock.  We don't have default
configuration in config file, but in `util.py` file.  That said, I think
the proposal matches your expectation, can you take a look at #555?

PR #555, opened 2020-04-01
mock: move site-defaults.cfg from /etc to %doc

praiskup:
The file was not pre-configured by default (every option there was
commented-out - we specify defaults in util.py) so it always existed
for documentation purposes anyway.

We keep the full site-defaults.cfg file in mock.rpm as %doc, but now we
can move the site-defaults.cfg file to mock-core-configs.rpm so it is
shipped with other %config files.

Fixes: #546


mock/mock.spec
```diff
@@ -177,6 +177,9 @@ cp -a docs/mock.cheat %{buildroot}%{_datadir}/cheat/mock
 install -d %{buildroot}/var/lib/mock
 install -d %{buildroot}/var/cache/mock
 
+mkdir -p %buildroot/%_pkgdocdir
```
kdudka, line 180: I would write `%{buildroot}` and `%{_pkgdocdir}`, to make the specfile look consistent.

The slash before `%_pkgdocdir` and `%{_sysconfdir}` is redudundant?

praiskup: Fixed, the additional slash was ugly, good catch.  Thanks for the review!

I dislike using the `%{macro}` variant when it is not necessary, same as
with shell variables (the `${prefix}name` case).  But changed, FWIW.

mock-core-configs/mock-core-configs.spec
```diff
@@ -95,6 +95,10 @@ elif [ -d %{buildroot}%{_sysconfdir}/bash_completion.d ]; then
     echo %{_sysconfdir}/bash_completion.d/mock >> %{name}.cfgs
 fi
 
+# reference correct site-defaults.cfg from mock.rpm
+mockdocir=%{_pkgdocdir}
+mockdocir=${mockdocir//mock-core-configs/mock}
+sed -i "s~@MOCKDOCDIR@~$mockdocir~" %{buildroot}%{_sysconfdir}/mock/site-defaults.cfg
```
xsuchy, line 101: mockdocir? Did not you mean mockdocdir?

mock/mock.spec
```diff
@@ -190,7 +193,7 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 
 %files
 %defattr(0644, root, mock)
-%config(noreplace) %{_sysconfdir}/mock/site-defaults.cfg
```
xsuchy, line 193: We should still own this one, right? It will be just the small one.

praiskup: Correct, in `mock-core-configs`.

kdudka: Yes, the file should not be owned by `mock`.  It would unnecessarily trigger file conflicts when deploying custom configuration as an RPM package (replacement of `mock-core-configs`).

mock-core-configs/mock-core-configs.spec
praiskup: Good catch, fixed.

mock/mock.spec
xsuchy: Well, in the past, I put it in mock (and not mock-core-configs) because it contains default mock configuration (albeit commented) and it is the first config file read (after setting up the defaults in utils.py). The intention was that if some other packages want to provide other configs, you can provide it in another file. But not in site-defaults.cfg.

kdudka: The problem is that there are no other files that work like `site-defaults.cfg`.  Either you can duplicate the common settings in all mock profiles, or put it somewhere to `$HOME`, which is far from optimal when you deploy the configuration as an RPM package.  Am I missing anything?

praiskup: > Well, in the past, I put it in mock (and not mock-core-configs) because
> it contains default mock configuration (albeit commented) and it is the
> first config file read (after setting up the defaults in utils.py).

Except for (intentional) exceptions, everybody will have the
/etc/mock/site-defaults.cfg files installed after this PR anyways - and
that file references the previous variant anyways below %doc.

But meh, this is certainly a negative side of this PR because some people
do `rpm -qc mock`, they notice the configuration files - go & edit (they
don't have to know there's mock-core-configs).  OTOH, positive thing is
that `rpm -qd mock` now reveals more ...

That said, would it make any sense to experiment with %ghost
/etc/mock/site-defaults.cfg to keep /etc/mock/site-deaults.cfg listed
as configuration file and heal your argument?

> The intention was that if some other packages want to provide other configs,
> you can provide it in another file

This is what was discussed in this PR;  that mock should put the default
configuration below /usr/share/mock, and override it by /etc/mock (similar
thing that is done by systemd configuration).

Issue #545, opened 2020-03-25
epel-rpm-macros is needed to build mock.spec on RHEL7 + EPEL7


praiskup:
I'm not sure I want to fix this, perhaps it is enough to re-define
'%python3_pkgversion 36' macro in mock.spec if it isn't defined, yet.

Conan-Kudo:
Why wouldn't using the `epel-7-x86_64` chroot not already bring in those macros?

praiskup:
When you are on RHEL7, and enable EPEL7 repo, you don's have those macros --
so you can not build mock.

PR #556, opened 2020-04-01
mock: re-define %python3_pkgversion on el7

praiskup:
The default RHEL 7 %python3_pkgversion contains value '3', which isn't
sufficient for mock.  We actually need `epel-rpm-macros` to be installed
so the macro %python3_pkgversion is re-defined to 36.

This patch fixes builds on EL7 _host_ machines.  The presence of
`epel-rpm-macros` isn't guaranteed there by default even when EPEL 7
repo is enabled (epel-rpm-macros is "only" installed into EPEL7 minimal
buildroot).

Fixes: #545

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -3,6 +3,9 @@
 
 %global __python %{__python3}
 %global python_sitelib %{python3_sitelib}
+%if 0%{?rhel} == 7
+%global python3_pkgversion 36
+%endif
 
 Summary: Builds packages inside chroots
 Name: mock


mock/mock.spec
```diff
@@ -3,6 +3,9 @@
 
 %global __python %{__python3}
 %global python_sitelib %{python3_sitelib}
+%if 0%{?rhel} == 7
```
xsuchy, line 6: and if python3_pkgversion equals to 3?
So if EPEL bumps up python3_pkgversion we do not need to change it here.

praiskup: You can not build it on RHEL 7 host, with EPEL 7 repo enabled, unless you also
install epel-rpm-macros ... and this problem is pretty sneaky to realize what is
going on ....

nkadel:
The same problem exists for RHEL 6 and CentOS 6, which are admittedly obsolete, and it helps keep CentOS 8 consistent. I In my RPM .spec files, I use:

%if 0%{?rhel}
BuildRequires:  epel-rpm-macros
%endif

That wouldn't be acceptable for building pure RHEL or CentOS RPM's but works well for my add-ons.

Similar problems exist for Amazon Linux, which is using python3_pkgversion of "37" in their python 3.7 updates.

praiskup:
`BuildRequires: epel-rpm-macros` doesn't help here.  It is too late -- we
need to have it in minimal buildroot, otherwise the buildrequires aren't
calculated properly.

> Similar problems exist for Amazon Linux, which is using python3_pkgversion of "37" in their python 3.7 updates.

We don't support build of mock on el6 and Amazon Linux;  el6 is out of scope now
and the other isn't needed (patches would be probably welcome, if you dared).

nkadel:
Ahh. I was unclear.  epel-rpm-macros hels building python3 based RPMS *for* CentOS 6, using mock on other systems. I've been using. I'd call those builds *by* mock, which is apparently supported based on mock-core-configs. It's a problem more easily addressed by tuning those configurations thatn editing mock directly, I think.

praiskup:
> It's a problem more easily addressed by tuning those configurations
> thatn editing mock directly, I think.

Ah, building mock **in** mock isn't a problem here; `epel-rpm-macros` is
indeed in minimal buildroot for epel (brought by `@buildsys-build` group)
or we explicitly install `epel-rpm-macros` (depends on concrete chroot).
So we didn't have to fix the mock configuration here.

The problem I tried to fix is building mock on **EL7 host**, directly.
Well, while I think about it - adding epel-rpm-macros would probably help
because we would enforce the people to install the package manually.
But since the package brings hacks like `python3_pkgversion` override, it
is IMO better if the package isn't installed on host.

nkadel:
Would you like my localized building setup, at https://github.com/nkadel/mockrepo/ ? I test out mock and mock-core-configs changes locally with that setup.

praiskup:
> Would you like my localized building setup, at https://github.com/nkadel/mockrepo/ ? 

Do you mean the `diff upstream/mock.spec nkadel/mock.spec`?  Perhaps, if
that makes things better (or it helps you...), we need to decide what is
OK to have upstream and what not ...  but I'm OK to review patches.

The Makefile machinery, that is probably not needed here.
I've seen the `eptl-rpm-config` dep, which is not IMO needed.
I've seen some python2 bits ... fight with python2 doesn't make sense nowadays.

Otherwise, go ahead, can't wait for the PR!

PR #561, opened 2020-04-14
bootstrap: fix --rootdir option

praiskup:
We don't want to
1. evaluate rootdir = basedir + 'root' on two places, and
2. mixup bootstrap with normal chroot.

Fixes: #560

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1008,6 +1008,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['version'] = version
     config_opts['basedir'] = '/var/lib/mock'  # root name is automatically added to this
     config_opts['resultdir'] = '{{basedir}}/{{root}}/result'
+    config_opts['rootdir'] = '{{basedir}}/{{root}}/root'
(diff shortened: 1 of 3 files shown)


praiskup:
This needs test case.

mock/py/mockbuild/util.py
```diff
@@ -1431,8 +1432,7 @@ def set_config_opts_per_cmdline(config_opts, options, args):
 
     check_config(config_opts)
     # can't cleanup unless resultdir is separate from the root dir
-    rootdir = os.path.join(config_opts['basedir'], config_opts['root'])
-    if is_in_dir(config_opts['resultdir'] % config_opts, rootdir):
+    if is_in_dir(config_opts['resultdir'] % config_opts, config_opts['rootdir']):
```
praiskup, line 1435: This is wrong.  The previous `rootdir` variable isn`t `config_opts['rootdir']` but one level up.
As a result, the RPMs from result dir are removed by default (weird thing that the testsuite
did not catch this... so testsuite needs fix here as well).

praiskup:
Merged as 29230f889083b76a5bed601a10f19b26a7470c11

PR #566, opened 2020-04-24
Fix 565

praiskup:
I'm not sure.  Don't we want to rather stop producing the list of installed
packages, as it is already printed out by dnf/yum?

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -240,15 +240,13 @@ class Commands(object):
         '''report the installed packages in the chroot to the root log'''
         self.buildroot.root_log.info("Installed packages:")
         self.buildroot.nuke_rpm_db()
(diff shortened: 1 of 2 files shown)


pmatilai:
Doh, didn't notice you already submitted a fix for this. Feel free to close #568 if you prefer doing it this way.

praiskup:
Your fix seems slightly better ... do you think we need the output on mock's
output, though?

pmatilai:
In the existing form, I don't know. What I'd find much more useful is the plain output of 'rpm -qa | sort' with the build dependencies installed in a separate file, so when "this was building fine just yesterday" happens, you could simply diff the files. As it is, the data is in the logs alright but getting it into a processable format requires quite a bit of sed-awk-wrestling.

praiskup:
Merged #568.

Issue #565, opened 2020-04-24
mock uses outside rpm for some queries in bootstrap mode


pmatilai:
### Short description of the problem

Mock uses outside rpm for some queries in bootstrap mode, which causes problems if the rpm on the outside doesn't understand the database format of the inside chroot. 

### Output of `rpm -q mock`

mock-2.2-1.fc31.noarch

### Steps to reproduce issue

```
[root@lumikko ~]# rpm -q rpm
rpm-4.15.1-1.fc31.x86_64
[root@lumikko ~]# cat /etc/mock/fedora-sqlite-x86_64.cfg
config_opts['target_arch'] = 'x86_64'
config_opts['legal_host_arches'] = ('x86_64',)

include('templates/fedora-rawhide.tpl')

config_opts['macros']['%_db_backend'] = 'sqlite'
config_opts['root'] = 'fedora-sqlite-{{ target_arch }}'
[root@lumikko ~]# rm -rf /var/cache/mock/fedora-sqlite-x86_64/root_cache/
[root@lumikko ~]#

$ mock --use-bootstrap-image -r fedora-sqlite-x86_64 <some_srpm>

INFO backend.py:241:  Installed packages:
DEBUG util.py:793:  child environment: None
DEBUG util.py:719:  Executing command: /bin/rpm --root /var/lib/mock/fedora-sqlite-x86_64/root -qa with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'} and shell True
DEBUG util.py:621:  error: cannot open Packages index using db5 - Permission denied (13)
DEBUG util.py:621:  error: cannot open Packages database in /var/lib/mock/fedora-sqlite-x86_64/root/var/lib/rpm
```

Any trivial src.rpm will be sufficient to trigger the above "Installed packages:" error after first root population.

### Any additional notes

I spotted two places where the outside rpm is used for queries
1. _show_installed_packages() in backend.py (ie the above case)
2. all_chroot_packages() in buildroot.py during dynamic buildrequires loop, but for that I have no reproducer

[...]

PR #568, opened 2020-04-24
bootstrap: avoid using host rpm _show_installed_packages()

pmatilai:
Similar to commit 1a6e49605f14137cd6e68459dbe3397addfc5588,
_show_installed_packages() trips up if host rpm cannot read the
database format used inside the chroot.

Using buildroot.all_chroot_packages() where this is already dealt with
fixes the issue and reduces code duplication while at it.

Fixes: #565


pmatilai:
Thanks!

PR #577, opened 2020-05-15
expand braced variables

dwmarshall:
If a baseurl has "${releasever}", dnf understands that, but mock does
not.

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -353,8 +353,10 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
                 expand[key] = self.config['dnf_vars'][key]
 
         for key, value in expand.items():
-            key = "$" + key
-            string = string.replace(key, value)
+            # replace both $key and ${key}
+            # dnf allows braced variables
+            string = string.replace('$' + key, value)
+            string = string.replace('${' + key + '}', value)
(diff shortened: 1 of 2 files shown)


mock/py/mockbuild/package_manager.py
```diff
@@ -353,8 +353,10 @@ def expand_url_vars(self, string):
                 expand[key] = self.config['dnf_vars'][key]
 
         for key, value in expand.items():
-            key = "$" + key
-            string = string.replace(key, value)
+            pattern = r"\$(?:" + re.escape(key)
+            pattern += r"|\{" + re.escape(key) + r"\})"
```
praiskup, line 357: Can not we simply
```python
string = string.replace("$" + key, value)
string = string.replace("${}".format(key), value)
```
?

I don't think it is correct to compile patterns in loop.  The re pattern is a bit
harder to read, e.g. `'\\$(?:releasever|\\{releasever\\})'`.

praiskup:
Thank you for the pull-request!

mock/py/mockbuild/package_manager.py
dwmarshall: Agreed, I think the second line should be

```
string = string.replace("${" + key + "}", value)
```

which is actually just about what I'm doing in my workaround hack.

praiskup:
Thank you!

PR #579, opened 2020-05-19
Fix nspawn user

praiskup:


praiskup:
Meh, till 4.0 we need to rewrite the uid change/escalation mechanism ..
This current variant is ... overcomplicated considering how trivial thing
we do :-/ (we need to have one place were we (re-)set uid/gid/user, no
matter if bootstrap/nspawn/simple chroot is used).

praiskup:
... I'm veeery close to burn out here.  Still not correct.  Ideas?

Issue #564, opened 2020-04-22
mock --postinstall does not work in bootstrap mode


javierpena:
### Short description of the problem
When building a package with --postinstall. In that case, mock fails to find the built package. For example:

$  mock --rebuild -r epel-8-x86_64 --postinstall python-distroinfo-0.3.2-2.el8.src.rpm

Using [1] as test source RPM. I get the following log message:

Can not load RPM file: /var/lib/mock/epel-8-x86_64/result/python3-distroinfo-0.3.2-2.el8.noarch.rpm.
Could not open: /var/lib/mock/epel-8-x86_64/result/python3-distroinfo-0.3.2-2.el8.noarch.rpm
WARNING: Failed install built packages

As a workaround, I can use --no-bootstrap-chroot, but the default configuration is broken.

[1] - https://kojipkgs.fedoraproject.org//packages/python-distroinfo/0.3.2/2.el8/src/python-distroinfo-0.3.2-2.el8.src.rpm

### Output of `rpm -q mock`

mock-2.2-1.el8.noarch (from epel-testing)

### Steps to reproduce issue

1. wget https://kojipkgs.fedoraproject.org//packages/python-distroinfo/0.3.2/2.el8/src/python-distroinfo-0.3.2-2.el8.src.rpm
2. mock --rebuild -r epel-8-x86_64 --postinstall python-distroinfo-0.3.2-2.el8.src.rpm

Do not forget to mention full commandline with the mock command you executed.

### Any additional notes

<details>
  <summary>
    <p>Output of `mock --debug-config`</p>
  </summary>
INFO: mock.py version 2.2 starting (python version = 3.6.8)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
config_opts['bootstrap_image'] = 'centos:8'
config_opts['build_log_fmt_str'] = '%(message)s'
config_opts['cache_alterations'] = False
config_opts['chroot_name'] = 'default'
config_opts['chroot_setup_cmd'] = ('install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 '
 'gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build '
 'info patch util-linux findutils grep epel-release epel-rpm-macros '
 'fedpkg-minimal')
config_opts['cleanup_on_failure'] = False
[...]

PR #580, opened 2020-05-21
Bindmount resultdir to bootstrap chroot

praiskup:
This is needed to allow package_manager to --postinstall the built
results to target chroot.

Fixes: #564, #580


praiskup:
... doesn't work for mockchain. :-(

Issue #570, opened 2020-04-26
%prep failure proceeds to %generate_buildrequires: dnf builddep: error: the following arguments are required: package


hroncok:
### Short description of the problem

When there is a failure in `%prep`, mock proceeds to process `%generate_buildrequires` but there is no SRPM, leading to `dnf builddep: error: the following arguments are required: package`.

### Output of `rpm -q mock`

mock-2.2-1.fc32.noarch

### Steps to reproduce issue

This spec:

```spec
Name:           reproducer-gbr
Version:        0
Release:        0
Summary:        ...
License:        MIT
BuildArch:      noarch

%description
...

%prep
exit 1

%generate_buildrequires
echo rpm

%files

```

```
$ rpmbuild -bs reproducer-gbr.spec
$ mock -r fedora-rawhide-x86_64 ../SRPMS/reproducer-gbr-0-0.src.rpm
```

#### Actual output:

```
...
Start: rpmbuild reproducer-gbr-0-0.src.rpm
Building target platforms: x86_64
Building for target x86_64
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.Ru023t
+ umask 022
+ cd /builddir/build/BUILD
+ exit 1


RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.Ru023t (%prep)
    Bad exit status from /var/tmp/rpm-tmp.Ru023t (%prep)
INFO: Dynamic buildrequires detected
INFO: Going to install missing buildrequires
No matches found for the following disable plugin patterns: local, spacewalk
usage: dnf builddep [-c [config file]] [-q] [-v] [--version] [--installroot [path]] [--nodocs] [--noplugins]
                    [--enableplugin [plugin]] [--disableplugin [plugin]] [--releasever RELEASEVER]
                    [--setopt SETOPTS] [--skip-broken] [-h] [--allowerasing] [-b | --nobest] [-C]
                    [-R [minutes]] [-d [debug level]] [--debugsolver] [--showduplicates] [-e ERRORLEVEL]
                    [--obsoletes] [--rpmverbosity [debug level name]] [-y] [--assumeno] [--enablerepo [repo]]
                    [--disablerepo [repo] | --repo [repo]] [--enable | --disable] [-x [package]]
                    [--disableexcludes [repo]] [--repofrompath [repo,path]] [--noautoremove] [--nogpgcheck]
[...]

PR #581, opened 2020-05-21
Better error for dynamic_buildrequires %prep failure

praiskup:
There were two more issues in dynamic_buildrequires handler.  First that
we set success=True anytime `rpmbuild -br` succeeded.  This would exit
the build requires loop too early if `-br` eventually returned exit
status 0.

The second is that we tried to install the nosrc.rpm file in finally
block, which was done everytime the `rpmbuild -br` command failed (not
only when exit status was 0 or 11).

Fixes: #570, #581


praiskup:
https://github.com/rpm-software-management/mock/pull/581/files?diff=split&w=1

PR #585, opened 2020-05-22
Revert "config: Change all openSUSE configs to use the download redir…

Conan-Kudo:
…ector"

This change made the situation worse, and now mock can't produce openSUSE
environments more often than before. The metalink at least gave DNF the option
of fallbacks, so we're switching back.

This reverts commit a7d588c22872d71a2f95666915cec94685e0234b.


praiskup:
Thank you, I'm giving it a week or so for thinking, and then probably merge.

praiskup:
Replaced by #607, thank you for the PR anyway!

PR #586, opened 2020-05-24
modularity: add support for disabling modules

jpoliv:
Add support for disabling dnf modules.

CentOS 8 use case:
disable the go-toolset module so that newer golang rpms can be installed from local repos.

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -154,6 +154,7 @@
 # and modules in bootstrap chroot.
 #config_opts['bootstrap_chroot_additional_packages'] = []
 #config_opts['bootstrap_module_enable'] = []
+#config_opts['bootstrap_module_disable'] = []
 #config_opts['bootstrap_module_install'] = []
 
 # if you want mock to automatically run createrepo on the rpms in your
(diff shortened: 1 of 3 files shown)


xsuchy:
Isn't it better to have it *before* the module-enable. This way you can disable default stream and then enable some different stream.

jpoliv:
Ind

> Isn't it better to have it _before_ the module-enable. This way you can disable default stream and then enable some different stream.

Indeed. Changed the execution order on the new commit.

praiskup:
Hi @jpoliv , thank you for the PR!

We dicussed the module enable/disable feature before in copr PR, and came
to conclusion that something like module_toggle option would make it good
enough, because we might need a bit more complicated enable/disable
sequence.

There's stub https://pagure.io/copr/copr/c/ddb706f92838dd167658be549761c71e00b2652e
in copr, we didn't finish it because there was no support in mock to
disable.

That said, I'd much rather see something _similar_ in mock --
configuration option which would allow us to enable/disable set of modules
in whatever order user prefers.

xsuchy:
@praiskup but these toggles will not help you with default streams, right?

praiskup:
Yeah, default streams that enable modules is a new topic for me so I need to think about it.

But we definitely enable the modules on the wrong place.  We have to process this before
every dnf operation, not during init phase ... otherwise the caches get affected, and we
never enable other modules again.

xsuchy:
I am going to close this one in favor or merged https://github.com/rpm-software-management/mock/pull/598 which provides more flexibility.

@jpoliv than you this PR, without this initial change the PR 598 would not happen.

jpoliv:
@xsuchy and @praiskup : Thanks for the #598 changeset (and mock-2.4). Much better.

Issue #588, opened 2020-05-28
custom CA certificates are not persistent


ktdreyer:
### Short description of the problem

Mock installs the host's custom CA certificates in such a way that they are almost always wiped out later. This means that DNF/Yum cannot verify HTTPS URLs signed by a custom CA, etc.

### Output of `rpm -q mock`

mock-2.2-1.el8.noarch

### Steps to reproduce issue

1. Start with a new CentOS 8 system and install mock
2. Add a new PEM-formatted CA file to `/etc/pki/ca-trust/source/anchors/my-custom-ca.crt`
3. Trust this CA system wide: `update-ca-trust extract`
4. Confirm that this new CA is available in the system bundle:
   `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`
5. Initialize a chroot, like epel-8-x86_64:

```
mock -r epel-8-x86_64 init
```
### Actual results:

My custom CA is not present in the main epel-8-x86_64 or epel-8-x86_64-bootstrap chroot locations.

```
grep -l My-Custom-CA /var/lib/mock/epel-8-x86_64*/root/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt
```

### Expected results:
My custom CA is present in both the epel-8-x86_64 and epel-8-x86_64-bootstrap chroot locations.

### Additional information:

Stepping through `package_manager.py`, Mock is calling `copy_certs()`, and we are succesfully calling `copy_tree()` like so:
```
copy_tree("/etc/pki/ca-trust/extracted", "/var/lib/mock/epel-8-x86_64/root/etc/pki/ca-trust/extracted")
```
The problem seems to be that something later overwrites `/var/lib/mock/epel-8-x86_64/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` in the chroot.

Perhaps one of the packages in CentOS 8 overwrites `ca-bundle.trust.crt` after mock modifies it. For example, this could happen if any of the packages in the chroot execute the `update-ca-trust extract` command.

[...]

Issue #397, opened 2019-11-06
RFE: Need to be able to inject file into chroot before chroot_setup_cmd runs


opoplawski:
### Short description of the problem

I need to make use of a local SSL filtering proxy.  So I need install our proxy CA cert into the chroot before the chroot_setup_cmd of install @buildsys-build runs.  I tried using:
```
config_opts['plugin_conf']['bind_mount_opts']['dirs'].append(('/etc/pki/ca-trust/extracted/pem', '/etc/pki/ca-trust/extracted/pem' ))
```
But that appears to run after chroot_setup_cmd.

### Output of `rpm -q mock`
mock-1.4.21-1.el7.noarch

opoplawski:
Thanks!

PR #589, opened 2020-05-28
mock: copy source CA certificates

ktdreyer:
Prior to this change, we would only copy the `extracted` SSL CA certificates into the chroot. If anything ran `update-ca-trust extract` inside the chroot, this would delete our custom SSL certificates from the `extracted` directory.

Copy the entire parent directory so that `sources` and `extracted` are both present in the chroot. With this change, `update-ca-trust extract` does not wipe out the CA certificates from the chroot.

Fixes: #588

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -301,7 +301,7 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
 
     @traceLog()
     def copy_certs(self):
-        cert_path = "/etc/pki/ca-trust/extracted"
+        cert_path = "/etc/pki/ca-trust"
         pki_dir = self.buildroot.make_chroot_path(cert_path)
         try:
             copy_tree(cert_path, pki_dir)


praiskup:
Thank you for the patch!

> If anything ran update-ca-trust extract inside the chroot

Can this happen by accident?  I mean, is it likely?  That would have to be
run by root, which means it would have to be done by some RPM scriptlet
installed by dnf...  Can you dump the story to the commit message?

Anyways, I'm +1, I don't think it can cause any harm, except
that it copies a bit more data into the chroot.

ktdreyer:
I found the package that does this. It's the main `ca-certificates` package, and it's done this from RHEL 6.5 onwards.

I've updated the commit message with your suggestion.

xsuchy:
For the record, the `extracted` was there just because of https://github.com/rpm-software-management/mock/issues/397, I think it is safe to copy the  `/etc/pki/ca-trust`. Especially when we copy whole tree. Which includes extracted as well.

praiskup:
> With this change, "update-ca-trust" does not wipe out the CA certificates from the host.

typo "from the chroot"

ktdreyer:
Whoops, thanks for catching that typo. I've amended this PR description and the Git commit log.

praiskup:
Thank you.

PR #594, opened 2020-06-22
recreate lvm_root snapshot after dnf update

dturecek:
Fixes: RHBZ#1175346

### DIFF
--- a/mock/integration-tests/20-lvm-plugin.tst
+++ b/mock/integration-tests/20-lvm-plugin.tst
@@ -43,6 +43,27 @@ runcmd "$MOCKCMD --rebuild ${TESTDIR}/test-C-1.1-0.src.rpm --resultdir $TMPDIR" 
 
 runcmd "$MOCKCMD --shell '/bin/true'" || die "mock shell failed"
 
+mkdir $TMPDIR/repodir || die "can't create repodir inside tmpdir"
+curl https://kojipkgs.fedoraproject.org//packages/python-copr-common/0.6/1.fc32/noarch/python3-copr-common-0.6-1.fc32.noarch.rpm \
+    --output $TMPDIR/repodir/python3-copr-common-0.6-1.fc32.noarch.rpm || die "failed to download old rpm package"
+
+runcmd "createrepo_c $TMPDIR/repodir" || die "failed to run createrepo"
+
+cat >> "$local_config" <<EOF
+config_opts['root'] = 'testrepo'
+config_opts['chroot_setup_cmd'] = 'install python3-copr-common'
+
+config_opts['yum.conf'] += """
+
+[testrepo]
+name=testrepo
+baseurl=file://$TMPDIR/repodir
+"""
+EOF
+
+runcmd "$MOCKCMD --rebuild ${TESTDIR}/test-C-1.1-0.src.rpm --resultdir $TMPDIR" \
+    | grep "created updated snapshot" || die "creating updated snapshot failed"
+
 runcmd "$MOCKCMD --scrub=all" || die "mock scrub failed"
 
 # repeated run should succeed as well, rhbz#1805179
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -215,7 +215,12 @@ class Buildroot(object):
                     and self.config['clean']):
                 update_state = "{0} update".format(self.pkg_manager.name)
                 self.state.start(update_state)
+                packages_before = self.all_chroot_packages()
                 self.pkg_manager.update()
+                packages_after = self.all_chroot_packages()
+                if packages_before != packages_after:
+                    self.plugins.call_hooks('postupdate')
+                    self.bootstrap_buildroot.plugins.call_hooks('postupdate')
                 self.state.finish(update_state)
         else:
             self._fixup_build_user()
--- a/mock/py/mockbuild/plugins/lvm_root.py
+++ b/mock/py/mockbuild/plugins/lvm_root.py
@@ -381,6 +381,25 @@ class LvmPlugin(object):
         lvm_do(['lvremove', '-f', self.vg_name + '/' + lv_name])
         self.buildroot.root_log.info("deleted {name} snapshot".format(name=name))
 
+    def update_snapshot_name(self, name):
+        # If the previous snapshot was already updated, remove
+        # the _updated_<date> postfix from its name.
+        name = re.sub(r"_updated_[\-\_0-9]{16}$", "", name)
+
+        timestamp = time.localtime(time.time())
+        name = name + "_updated_{y}-{mo:02d}-{d:02d}_{h:02d}-{m:02d}".format(
+            y=timestamp.tm_year, mo=timestamp.tm_mon, d=timestamp.tm_mday,
+            h=timestamp.tm_hour, m=timestamp.tm_min
+        )
+
+        return name
+
+    def hook_postupdate(self):
+        name = self.update_snapshot_name(self.get_current_snapshot())
+        self.make_snapshot(name)
+        self.buildroot.root_log.info("created updated snapshot {name}".format(
+            name=self.remove_prefix(name)))
+
 
 def init(plugins, lvm_conf, buildroot):
     LvmPlugin(plugins, lvm_conf, buildroot)
--- a/mock/py/mockbuild/plugins/root_cache.py
+++ b/mock/py/mockbuild/plugins/root_cache.py
@@ -57,6 +57,7 @@ class RootCache(object):
         plugins.add_hook("postshell", self._rootCachePostShellHook)
         plugins.add_hook("postchroot", self._rootCachePostShellHook)
         plugins.add_hook("postyum", self._rootCachePostShellHook)
+        plugins.add_hook("postupdate", self._rootCachePostUpdateHook)
         self.exclude_dirs = self.root_cache_opts['exclude_dirs']
         self.exclude_tar_cmds = []
         for ex_dir in self.exclude_dirs:
@@ -185,7 +186,7 @@ class RootCache(object):
         self._rebuild_root_cache()
 
     @traceLog()
-    def _rebuild_root_cache(self):
+    def _rebuild_root_cache(self, after_update=False):
         try:
             self._rootCacheLock(shared=0)
             # nuke any rpmdb tmp files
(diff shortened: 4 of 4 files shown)


mock/py/mockbuild/buildroot.py
```diff
@@ -216,7 +216,11 @@ def _init(self, prebuild, do_log):
                     and self.config['clean']):
                 update_state = "{0} update".format(self.pkg_manager.name)
                 self.state.start(update_state)
-                self.pkg_manager.update()
+                output = self.pkg_manager.update()
+                if ("lvm_root" in self.plugins.plugins and
+                        "Upgraded:" in output):
+                    self.plugins.call_hooks('recreate_snapshot')
+                    self.bootstrap_buildroot.plugins.call_hooks('recreate_snapshot')
```
praiskup, line 223: Thanks for the patch.  There's `all_chroot_packages` method somewhere in mock code, using that would be much more reliable way to detect something changed than parsing the dnf output.

praiskup: Calling the hook should be plugin agnostic, simply "if something changed => call the hook".  Ie don't ask for `lvm_root` plugin here.

mock/py/mockbuild/plugins/lvm_root.py
```diff
@@ -382,6 +384,22 @@ def hook_remove_snapshot(self, name):
         lvm_do(['lvremove', '-f', self.vg_name + '/' + lv_name])
         self.buildroot.root_log.info("deleted {name} snapshot".format(name=name))
 
+    def hook_recreate_snapshot(self):
+        name = self.get_current_snapshot()
```
praiskup, line 388: Why "recreate" mechanism, instead of creating a completely new snapshot?

mock/py/mockbuild/buildroot.py
dturecek: This makes sense, thanks for the hint.

mock/py/mockbuild/plugins/lvm_root.py
dturecek: That's basically what I do. However, this is called only when there already is an existing snapshot of the same name (so I can't just call `make_snapshot`), which is removed and a new one is created. So it made sense to me to create a new hook for this.

praiskup: Another hook name doesn't sound entirely wrong to me, but I'm surprised that nothing
within the tho hook methods creating the snapshot from (re-)initialized chroot  can be
shared.
Also, I'd prefer a fix that would generically fix mock, not only lvm_chroot (root_cache
suffers from the same problem).

praiskup: Thinking once more about this; What if we invented `hook_postupdate` call,
and we called this hook always when some package was added during `dnf
udpate`?

Both lvm and root-cache (and maybe overlayfs) could leverage that hook
(snapshot in name doesn't match root-cache use case).

From LVM plugin, I don't know.  Do you think that we could have another
snapshot named like `update-<timestamp>` and set it as current, so users
can rollback to the `postinit` if they wanted?  (sounds better to me than just
replacing the `postinit` snapshot).

dturecek: I think it makes sense to create a new, updated snapshot. I've changed it per your suggestion. The root_cache, however, needs some more work.

mock/py/mockbuild/plugins/lvm_root.py
```diff
@@ -382,6 +382,21 @@ def hook_remove_snapshot(self, name):
...
+        if re.match(r".*_updated_[\-\_0-9]{16}$", name):
+            name = re.sub(r"_updated_[\-\_0-9]{16}$", "", name)
+
+        timestamp = time.localtime(time.time())
+        name = name + "_updated_{y}-{mo:02d}-{d:02d}_{h:02d}-{m:02d}".format(
+            y=timestamp.tm_year, mo=timestamp.tm_mon, d=timestamp.tm_mday,
+            h=timestamp.tm_hour, m=timestamp.tm_min
+        )
+
+        self.make_snapshot(name)
```
praiskup, line 396: We probably need to call set_current_snapshot?

mock/py/mockbuild/plugins/lvm_root.py
```diff
@@ -382,6 +382,21 @@ def hook_remove_snapshot(self, name):
         lvm_do(['lvremove', '-f', self.vg_name + '/' + lv_name])
         self.buildroot.root_log.info("deleted {name} snapshot".format(name=name))
 
+    def hook_postupdate(self):
+        name = self.get_current_snapshot()
+        if re.match(r".*_updated_[\-\_0-9]{16}$", name):
```
praiskup, line 387: I fail to see the reason for this condition, can you please explain it in comment?

mock/py/mockbuild/plugins/lvm_root.py
```diff
@@ -382,6 +382,21 @@ def hook_remove_snapshot(self, name):
...
+            name = re.sub(r"_updated_[\-\_0-9]{16}$", "", name)
+
+        timestamp = time.localtime(time.time())
+        name = name + "_updated_{y}-{mo:02d}-{d:02d}_{h:02d}-{m:02d}".format(
+            y=timestamp.tm_year, mo=timestamp.tm_mon, d=timestamp.tm_mday,
+            h=timestamp.tm_hour, m=timestamp.tm_min
+        )
```
praiskup, line 394: The name is pretty complicated.  I think it would be enough to have just the numeric `time.time()` output there, but no problem here .. just saying.

praiskup: Just checking, are the "updated" snapshot deleted with --scrub?

mock/py/mockbuild/plugins/lvm_root.py
praiskup: Thanks.

> The root_cache, however, needs some more work.

Isn't it just about calling `self._rebuild_root_cache()`?

dturecek: Well, `_rebuild_root_cache()` only rebuilds the cache if `self.config['cache_alterations']` is set. If that's what we want, then yes.

mock/py/mockbuild/plugins/lvm_root.py
dturecek: `set_current_snapshot` is called in `make_snapshot`.

mock/py/mockbuild/plugins/lvm_root.py
dturecek: You are right, the if is unnecessary here, the `re.sub` does the same thing without it.

mock/py/mockbuild/plugins/lvm_root.py
dturecek: Yes, they are.

mock/py/mockbuild/plugins/lvm_root.py
praiskup: Well, we can slightly alter `_rebuild_root_cache`, too (so it does something in other occasions, i.e. chroot is updated).

mock/py/mockbuild/plugins/lvm_root.py
praiskup: Ah, I was mistaken by `hook_postinit` which (mistakenly) re-calls the `set_current_snapshot` method.

praiskup: Can you please quickly review #596?

mock/py/mockbuild/plugins/lvm_root.py
praiskup: I still don't get what is the `sub` for :-) documentation is needed.

dturecek: Its for the situation when the current snapshot is already updated, so that there are no `postinit_updated_<date1>_updated_<date2>...` snapshots.

mock/py/mockbuild/plugins/lvm_root.py
dturecek: I've made a change to `_rebuild_root_cache` so now it should be rebuilt in the `postupdate` hook.

praiskup: I like the current variant.  Can you please wrap the logic related to snapshot rename to separate, easily testable function?  Something like `new_name = update_snapshot_name(old_name)` (pick whatever name and input/output format, the point is to make it testable)?  Then you could add several trivial unit-tests for the function.

Could you please also enhance the lvm integration test-case so we have this functionality tested?

praiskup:
Gently ping here :-)

dturecek:
I've added a work-in-progress test.

mock/integration-tests/20-lvm-plugin.tst
```diff
@@ -43,6 +43,27 @@ runcmd "$MOCKCMD --rebuild ${TESTDIR}/test-C-1.1-0.src.rpm --resultdir $TMPDIR"
 
 runcmd "$MOCKCMD --shell '/bin/true'" || die "mock shell failed"
 
+mkdir $TMPDIR/repodir || die "can't create repodir inside tmpdir"
+curl https://kojipkgs.fedoraproject.org//packages/python-copr-common/0.6/1.fc32/noarch/python3-copr-common-0.6-1.fc32.noarch.rpm \
+    --output $TMPDIR/repodir/python3-copr-common-0.6-1.fc32.noarch.rpm || die "failed to download old rpm package"
```
praiskup, line 48: The problem is that the `python3-copr-common` package is in the default fedora repository, and (especially in future) it is in higher version --- so this explicit variant likely won't be installed.

We need some trivial spec file, and build a custom RPM with artificial NVR here ... which will be later (before the image update) bumped once more.

praiskup: I created a `mock-test-bump-version` package, [v1 src.rpm](https://raw.githubusercontent.com/rpm-software-management/mock-test-data/master/mock-test-bump-version-1-0.src.rpm) and [v2 src.rpm](https://raw.githubusercontent.com/rpm-software-management/mock-test-data/master/mock-test-bump-version-2-0.src.rpm).

Feel free to use this in pattern:
```
wget those files
mock <version1> --target --resultdir <repo>
createrepo_c <repo>
generate custom config <custom.cfg> with that repo + added the package in the minimal buildroot
mock -r <custom.cfg> --init
mock <version2> --target --resultdir <repo>
createrepo_c <repo>
mock -r <custom.cfg> --rebuild something # check that the package is updated, and cache regenerated
```

praiskup:
Ping, can you please push the actual variant you probably have only locally?

praiskup:
I'm trying to finish this PR in separate one #637.  Any review is very welcome!

praiskup:
And thank you a lot for this PR.

PR #595, opened 2020-06-23
Use a different .rpmmacros for install/build time

praiskup:
This was unified before under an assumption that nobody installs any
RPMs from within the mock chroot (except for the mock bootstrap chroot
itself).

But apparently tools like lorax and dracut do this, and then they miss
the /sys and /proc directories (normally installed by filesystem.rpm).
So let's split the logic again.

Before 85d544291da we created the installation .rpmmacros inside
"basedir" which was breaking --bootstrap-chroot inside Fedora Toolbox.
That was because the "basedir" isn't recursively bind-mounted below the
bootstrap chroot and thus the bootstrapped RPM failed to find the file.
Newly we create it in "rootdir" instead.

Resolves: rhbz#1848201
Reverts: 85d544291da69dbc521369c8b9b87baa3f7ffc42

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -194,7 +194,6 @@ class Buildroot(object):
         self.pkg_manager.initialize()
         self._setup_resolver_config()
         if not self.chroot_was_initialized:
-            self._prepare_rpm_macros()
             self._setup_dbus_uuid()
             self._init_aux_files()
             if not util.USE_NSPAWN:
@@ -204,7 +203,6 @@ class Buildroot(object):
             self._make_build_user()
             self._setup_build_dirs()
         elif prebuild:
-            self._prepare_rpm_macros()
             if 'age_check' in self.config['plugin_conf']['root_cache_opts'] and \
                not self.config['plugin_conf']['root_cache_opts']['age_check']:
                 self._init_pkg_management()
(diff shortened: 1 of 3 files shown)


mock/py/mockbuild/buildroot.py
```diff
@@ -520,25 +518,44 @@ def chown_home_dir(self):
...
-        To let use this, mock needs to set $HOME to /builddir.  Note that the
-        `rpm --root` command reads this from $HOME on host - not from
-        <root>/$HOME.
+        Install the /builddir/.rpmmacros file used by /bin/rpmbuild at build
+        time.
         """
         macro_dir = self.make_chroot_path(self.homedir)
         util.mkdirIfAbsent(macro_dir)
         macrofile_out = os.path.join(macro_dir, ".rpmmacros")
+
```
praiskup, line 550: useless change

mock/py/mockbuild/buildroot.py
```diff
@@ -520,25 +518,44 @@ def chown_home_dir(self):
...
         with open(macrofile_out, 'w+') as rpmmacros:
 
             # user specific from rpm macro file definitions first
             if 'macrofile' in self.config:
                 with open(self.config['macrofile'], 'r') as macro_conf:
                     rpmmacros.write("%s\n\n" % macro_conf.read())
 
-            for key, value in list(self.config['macros'].items()):
+            for key, value in list(self.config["macros"].items()):
```
praiskup, line 558: useless change as well

PR #596, opened 2020-06-26
lvm: don't recall set_current_snapshot unnecessarily

praiskup:
Per discussion in PR#594, it is useless to re-call set_current_snapshot
after make_snapshot.


dturecek:
Looks good to me.

PR #598, opened 2020-06-29
Add `module_setup_commands` configuration option

praiskup:
It obsoletes `module_enable` and `module_install` configuration options
(still supported) but allows users to also configure "disable",
"remove", "list", and other commands.

It may or may not be useful in future, but we also execute the commands
from `module_setup_commands` as they are specified, so people can do
things like `disable foo:*` and then `enable foo:stream`.

Replaces: #586

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -153,8 +153,7 @@
 # These three are overrided in bootstrap by default as we usually do not want additional packages
 # and modules in bootstrap chroot.
 #config_opts['bootstrap_chroot_additional_packages'] = []
-#config_opts['bootstrap_module_enable'] = []
-#config_opts['bootstrap_module_install'] = []
+#config_opts['bootstrap_module_setup_commands'] = []
 
 # if you want mock to automatically run createrepo on the rpms in your
 # resultdir.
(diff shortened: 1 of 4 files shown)


mock/py/mockbuild/util.py
```diff
@@ -1663,6 +1665,11 @@ def load_config(config_path, name, uidManager, version, pkg_python_dir):
                     "please use config_opts['use_bootstrap'] instead")
         config_opts['use_bootstrap'] = config_opts['use_bootstrap_container']
 
+    # module_enable and module_install are obsoleted by module_setup_commands
+    for option in ["module_enable", "module_install"]:
+        if option in config_opts and config_opts[option]:
+            log.warning("%s option is obsoleted, use module_setup_commands", option)
```
praiskup, line 1671: After discussion with @xsuchy, we should keep this without warning for now.

mock/docs/site-defaults.cfg
```diff
@@ -419,9 +418,17 @@
...
-# config_opts['module_install'] = ['module1/profile', 'module2/profile']
+
+# List of module commands to be executed when initializing chroot, before
+# `chroot_setup_cmd`.  Each command is a pair like `(action, modules)`
+# where modules is a string with comma-separated list of modules.
+#
+#config_opts['module_setup_commands'] = [
+#  ('disable', 'foo_module:*'),
+#  ('enable', 'bar_module:2, baz_module:3'),
+#  ('install', 'bar_module:2'),
```
praiskup, line 429: @xsuchy's idea, we should make clear we can do some of those multiple times

praiskup:
Some of the core-config templates contain examples with module_enable, etc.  We should transform them.

mock/docs/site-defaults.cfg
```diff
@@ -419,9 +418,23 @@
...
+#
+## Artificial example: (a) Disable any potentially enabled postgresql module
+## stream, (b) enable _specific_ postgresql and ruby module streams,
+## (c) install the development nodejs profile and (d) disable it immediately.
+#config_opts['module_setup_commands'] = [
+#  ('disable', 'postgresql'),
+#  ('enable',  'postgresql:12, ruby:2.6'),
+#  ('install', 'nodejs:13/development'),
+#  ('disable', 'nodejs'),
+#]
```
praiskup, line 436: @dmach could you please quickly review that this is a reasonable mock chroot configuration design?

PR #603, opened 2020-07-14
introduce mock-filesystem subpackage

xsuchy:
Since introduction of `mock-configs` virtual provides, it can happen
that mock-core-configs is not actually installed and `mock` group will
not be created.
Let put the scriptlet into separate subpackage. And together with basic
filesystem.

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -1,6 +1,9 @@
 %bcond_with lint
 %bcond_without tests
 
+# mock group id allocate for Fedora
+%global mockgid 135
+
 %global __python %{__python3}
 %global python_sitelib %{python3_sitelib}
 %if 0%{?rhel} == 7
@@ -35,6 +38,7 @@ Conflicts: mock-core-configs < 32.6
 
 # Requires 'mock-core-configs', or replacement (GitHub PR#544).
 Requires: mock-configs
+Requires: %{name}-filesystem
 %if 0%{?fedora} || 0%{?rhel} >= 8
 # This is still preferred package providing 'mock-configs'
 Suggests: mock-core-configs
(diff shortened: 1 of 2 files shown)


praiskup:
> This subpackage should probably be part of mock instead of mock-core-configs.

Do you think it makes any practical difference?

praiskup:
One thing that comes to my mind is that someone may want to build mock.rpm, but doesn't care about our mock-core-configs.

Conan-Kudo:
Yes, this already is the case in OpenMandriva, where their buildsystem just needs the file structure so that they can put the configs in the right place.

xsuchy:
updated

mock/mock.spec
```diff
@@ -35,6 +38,7 @@ Conflicts: mock-core-configs < 32.6
 
 # Requires 'mock-core-configs', or replacement (GitHub PR#544).
 Requires: mock-configs
+Requires: %{name}-filesystem
```
praiskup, line 41: I think we should require `%name-fliesystem = %version-%release`.

praiskup:
Otherwise looks good, thank you.

mock/mock.spec
xsuchy: Why? I am not fan of this. If we ever need some functionality from mock-filesystem then we can put there `>= X.Y`. No need to enforce exact version when it is not needed.

praiskup: I thought it is per Fedora guidelines, but [it's just a suggestion, not must](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_requiring_base_package).  And the suggestion is only related to the base package....

If you **do not** overspecify the Requires, it can (in theory at least) cause some troubles on version mismatch in the future.  While there are no consequences when you do so.  Nevermind, I'm not going to require a fix here.

praiskup:
Ok, I'd like to release `mock-core-configs` first (for the F33 branched configs).  Then we
can merge this (after this PR, we should probably release both mock and mock-core-configs
together).

mock/mock.spec
praiskup: > Nevermind, I'm not going to require a fix here.

I still prefer it though... (when both packages come from the same spec).  But feel free to merge as is.

PR #604, opened 2020-07-17
config: adding config for epel-8-s390x

schlupov:
Adding missing config for s390x for epel 8 so we can use emulated s390x builders in copr.

### DIFF
--- a/mock-core-configs/etc/mock/epel-8-s390x.cfg
+++ b/mock-core-configs/etc/mock/epel-8-s390x.cfg
@@ -1,0 +1,5 @@
+include('templates/epel-8.tpl')
+
+config_opts['root'] = 'epel-8-s390x'
+config_opts['target_arch'] = 's390x'
+config_opts['legal_host_arches'] = ('s390x',)


mock-core-configs/etc/mock/epel-8-s390x.cfg
```diff
@@ -0,0 +1,5 @@
+include('templates/epel-8.tpl')
```
xsuchy, line 1: You are missing: `include('templates/centos-8.tpl')` at very top of this file.

xsuchy: Hmm, even with that line it does not work. Without the centos line I get:
```

TypeError: can only concatenate tuple (not "str") to tuple

ERROR: Error in configuration
```
With the line I get:
```
CentOS-8 - Base                                                                          400  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'BaseOS': Cannot prepare internal mirrorlist: No URLs in mirrorlist
ERROR: Command failed: 
 # /usr/bin/dnf --installroot /var/lib/mock/epel-8-s390x-bootstrap/root/ --releasever 8 --setopt=deltarpm=False --allowerasing --disableplugin=local --disableplugin=spacewalk --forcearch s390x install dnf dnf-plugins-core --setopt=tsflags=nocontexts
No matches found for the following disable plugin patterns: local, spacewalk
CentOS-8 - Base                                                                          400  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'BaseOS': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

Conan-Kudo:
There is no s390x port of CentOS, so this cannot be merged until that is fixed.

praiskup:
This was closed by GitHub while moved to the new `main` branch name.  @schlupov, this is blocked PR anyway, so once it is unblocked please submit a new PR.

PR #608, opened 2020-08-02
set user_agent in dnf.conf

### LINKED ISSUE #562: set user_agent in dnf.conf
Mock should set 'user_agent' in dnf.conf to identify as a build system.

See https://bugzilla.redhat.com/show_bug.cgi?id=1813424

xsuchy:
Fixes #562

Any comments? If there will be no objections, I will add it to all configs.

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-branched.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-branched.tpl
@@ -29,6 +29,7 @@ metadata_expire=0
 best=1
 module_platform_id=platform:f{{ releasever }}
 protected_packages=
+user_agent={{ user_agent }}
 
 # repos
 
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -485,3 +485,6 @@
 # Fedora or EL systems work with that.  But some destination chroots can have
 # different configuration, and copying the bundle helps.
 #config_opts['ssl_ca_bundle_path'] = None
+
+# user_agent string to identify HTTP request to repositories
+# config_opts['user_agent'] = "mock (Fedora {{ releasever }}; mock; {{ target_arch }})"
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1251,7 +1251,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['rpm_command'] = '/bin/rpm'
     config_opts['rpmbuild_command'] = '/usr/bin/rpmbuild'
     config_opts['dnf_disable_plugins'] = ['local', 'spacewalk']
-
+    config_opts['user_agent'] = "Mock ({{ root }}; {{ target_arch }})"
     config_opts['opstimeout'] = 0
 
     config_opts['stderr_line_prefix'] = ""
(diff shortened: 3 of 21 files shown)


mock-core-configs/etc/mock/templates/fedora-branched.tpl
```diff
@@ -29,6 +29,7 @@ metadata_expire=0
 best=1
 module_platform_id=platform:f{{ releasever }}
 protected_packages=
+user_agent=mock (Fedora {{ releasever }}; mock; {{ target_arch }})
```
praiskup, line 32: Don't you want to use the standard formatt `Mock/<Version> <comment>`?  `tito.props` describes `VERSION_TEMPLATE` that could help.

Also, I'd slightly prefer using something from `python-distro` instead of `target_arch`.  The target arch should be obvious from the downloaded package file name (path).

praiskup:
At least rawhide should have this as well.  I'm not sure about EPEL, etc?

mock-core-configs/etc/mock/templates/fedora-branched.tpl
xsuchy: That is a question. I even considered just "mock" as the version or arch will not be too much helpful.

xsuchy:
I created new config option and it is used in all config (but f30, which is EOLed)

mock-core-configs/etc/mock/templates/fedora-branched.tpl
praiskup: Ok, per RFC 7231, product version is [optional](https://tools.ietf.org/html/rfc7231#page-46).  Previously I looked at some wrong interpretation by Mozilla.

praiskup: Otherwise, I don't know.  For us (as mock maintainers) is interesting to hear
where people run mock tool (Fedora version and architercture, e.g.).

The `target_arch` seems to be redundant info.

So good enough format would be e.g.:
`Mock/2.4 (Fedora 32; x86_64)`

xsuchy:
updated

mock/py/mockbuild/util.py
```diff
@@ -1251,7 +1251,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['rpm_command'] = '/bin/rpm'
     config_opts['rpmbuild_command'] = '/usr/bin/rpmbuild'
     config_opts['dnf_disable_plugins'] = ['local', 'spacewalk']
-
+    config_opts['user_agent'] = "Mock ({{ root }}; {{ target_arch }})"
```
praiskup, line 1254: Strictly speaking, I think we can drop the `target_arch` part as well since it is part of `root`?

mock/docs/site-defaults.cfg
```diff
@@ -485,3 +485,6 @@
 # Fedora or EL systems work with that.  But some destination chroots can have
 # different configuration, and copying the bundle helps.
 #config_opts['ssl_ca_bundle_path'] = None
+
+# user_agent string to identify HTTP request to repositories
+# config_opts['user_agent'] = "mock (Fedora {{ releasever }}; mock; {{ target_arch }})"
```
praiskup, line 490: This needs to be synced with the real value.

praiskup:
Otherwise +1

mock/docs/site-defaults.cfg
praiskup: Ping @xsuchy at least about this.

praiskup: It is just a leftover comment, the code now mentiones `root`, which is `<distro>-<release>-<arch>` triplet.

Conan-Kudo:
@xsuchy Fix the comment in the site-defaults config file and then 👍

praiskup:
Closed by 47474cc5b186388a345cd1ab98e420fa4c7dbcfb

PR #611, opened 2020-08-04
Add Fedora 33 branched configs

praiskup:


xsuchy:
You want to add `Requires: distribution-gpg-keys >= 1.41` - which is going to be released soon and will containg F34 gpg key.

praiskup:
> You want to add `Requires: distribution-gpg-keys >= 1.41` - which is going to be released soon and will containg F34 gpg key.

Do you have infra issue for this?

PR #613, opened 2020-08-06
Add showrc plugin to record the output of rpm --showrc

jcpunk:
For deeper insight into what macros existed and how they were defined at build time, I added a plugin that will record the output of `rpm --showrc`.  This would be handy when trying to determine exactly what packaging macros were present in the buildroot.

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -231,6 +231,10 @@
 # config_opts['plugin_conf']['procenv_enable'] = False
 # config_opts['plugin_conf']['procenv_opts'] = {}
 #
+# config_opts['plugin_conf']['showrc'] = False
+# config_opts['plugin_conf']['showrc'] = {}
+#
+#
 # bind mount plugin is enabled by default but has no configured directories to
 # mount
 # config_opts['plugin_conf']['bind_mount_enable'] = True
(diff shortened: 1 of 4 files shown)


mock/docs/site-defaults.cfg
```diff
@@ -231,6 +231,10 @@
 # config_opts['plugin_conf']['procenv_enable'] = False
 # config_opts['plugin_conf']['procenv_opts'] = {}
 #
+# config_opts['plugin_conf']['showrc'] = True
```
xsuchy, line 234: Do we really want to have this enabled by default for everybody?

xsuchy:
Otherwise +1 from me.
Once merged (and if) please document it here https://github.com/rpm-software-management/mock/wiki#plugins

mock/docs/site-defaults.cfg
jcpunk: These logs would increase reproducibility and provide more insight into the build root.  I'm inclined to have it on by default for folks.

praiskup: I'm afraid there's unnecessary risk that we'll break something (production Copr/Koji etc.).  There were dozens of similar problems when we were executing rpm inside the chroot from host.  That said, I'd prefer to enable it by default a bit later.

jcpunk: I'll update the diff to disable by default, squash, and have the PR updated shortly.

praiskup:
Thank you!

praiskup:
Can you please take a look at the wiki page documentation?

jcpunk:
The wiki should now be up to date : https://github.com/rpm-software-management/mock/wiki/Plugin-Showrc (the published in version I set to 2.4.2 as my guess for the next mock release)

praiskup:
Thank you.  The next release will be `2.5-1` (version 2.5).  I did one additional cleanup update on that wiki;  can you check it is OK?

jcpunk:
looks perfect to me :)

PR #625, opened 2020-09-09
create /etc/localtime as symlink even with isolation=simple

xsuchy:


mock/py/mockbuild/buildroot.py
```diff
@@ -276,14 +276,23 @@ def all_chroot_packages(self):
         return set(out.splitlines())
 
     @traceLog()
-    def _copy_config(self, filename):
+    def _copy_config(self, filename, symlink=False):
```
Conan-Kudo, line 279: Is there a good reason this default behavior should be `False`?

Conan-Kudo (changes requested):

praiskup:
> create /etc/localtime as symlink even with isolation=simple

I'd appreciate a better description, what is the background story?  Nspawn creates this out of box?

mock/py/mockbuild/buildroot.py
xsuchy: I am trying to make minimal impact. This function is used to copy /etc/{hosts,resolv.conf} too. In case they are symlinks, I do not want to copy the symlink ('cos the destination will likely does not exists), but I rather want to copy just the content.

xsuchy:
> I'd appreciate a better description, what is the background story? Nspawn creates this out of box?

https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/BNTFZH6VS43Q7FLRIZYSBOTKDK6KMQZQ/

PR #629, opened 2020-09-11
Nicer --verbose output

praiskup:
Print the commands we run in format which is copy-paste'able to the
commandline.


xsuchy:
+100, that will ease the debugging a lot :)

PR #631, opened 2020-09-15
Fix _copy_config() for broken symlinks in dst=

praiskup:
shutil.copy*() ignores the broken symbolic link in dst=, and simply
attempts to create the symbolic link destination file [1] and that
operation is likely to fail in mock case.

We already tried to remove the destination file, but first it was a (a)
TOCTOU way and (b) also the used os.path.exists() returns False for
broken symlinks.  So newly, let's unlink the file first without asking,
and just ignore FileNotFoundError.

[1] https://bugs.python.org/issue34374

Fixes: rhbz#1878924


keszybz:
LGTM.

praiskup:
Thanks for taking a look!

praiskup:
Followup after @xsuchy's review: #632

PR #633, opened 2020-09-15
Fix --chain and --continue

praiskup:


praiskup:
I believe that we should actually abstract the "failover" with
`--continue` option somewhere, and reuse it with `--rebuild` too.
But that's not something I'm able to spend the time on now, so let's be at
least honest and admit that `--continue` and `--recurse` doesn't work
without `--chain`.

xsuchy (approved):

PR #636, opened 2020-09-17
Log mock's NVR

praiskup:
Fixes: rhbz#1879231


keszybz:
Oh, that was quick. Thanks, looks like this solves the issue nicely.

Issue #639, opened 2020-10-02
Error "TypeError: must be real number, not TemplatedDictionary" when `%2F` in base path


sonneveld:
### Error "TypeError: must be real number, not TemplatedDictionary" when `%2F` in base path

For context, we create custom mock environments for our jenkins builds and for some reason, one of our nodes started creating workspace directories with "%2F" in the name. It looks like mock is trying to do variable substitution for the values, which gets confused with `%2F` 

### Output of `rpm -q mock`

mock-2.4-1.el7.noarch

### Steps to reproduce issue

1. Create new mock config directory with a default.cfg, logging.ini and site-defaults.cfg
2. In site-defaults.cfg, set base dir and cache_topdir to a path with `%2F` in it
3. Run mock with `mock --configdir=$CONFIGDIR --shell`

site-defaults.cfg:

```
config_opts['basedir'] = '/data/jenkins/workspace/abc-group/abc%2Fabc/MR-123/abc/build/work/abc/mock/base'
config_opts['cache_topdir'] = '/data/jenkins/workspace/abc-group/abc%2Fabc/MR-123/abc/build/work/abc/mock/cache'
```

You'll get an error along the lines of

```
16:39:43  ERROR: must be real number, not TemplatedDictionary
16:39:43  Traceback (most recent call last):
16:39:43    File "/usr/libexec/mock/mock", line 1026, in <module>
16:39:43      exitStatus = main()
16:39:43    File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 93, in trace
16:39:43      result = func(*args, **kw)
16:39:43    File "/usr/libexec/mock/mock", line 655, in main
16:39:43      util.set_config_opts_per_cmdline(config_opts, options, args)
16:39:43    File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 93, in trace
16:39:43      result = func(*args, **kw)
16:39:43    File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 1437, in set_config_opts_per_cmdline
16:39:43      if is_in_dir(config_opts['resultdir'] % config_opts, basechrootdir):
16:39:43  TypeError: must be real number, not TemplatedDictionary
```

xsuchy:
Hmm, simple reproducer:

```
$ python3
Python 3.9.0rc2 (default, Sep 17 2020, 00:00:00) 
[GCC 10.2.1 20200826 (Red Hat 10.2.1-3)] on linux
[...]

PR #640, opened 2020-10-04
Allow %-sign in config_opts['resultdir']

praiskup:
Fixes: #639


xsuchy:
If the %(foo) is present in the string, then this emits a warning. But it does not fixes #639 as it does nothing about `%2F` string as Python will try to use it in format().

xsuchy:
I agree with the merge. Leave there for two releases. And then remove it together with %(foo) functionality.
Although, you should remove the 'fixes' from commit description.

praiskup:
ACK, but AIUI this patch should fix #639 -- as long as the %() syntax isn't used intentionaly.  So the expansion in #639 should even happen.

PR #643, opened 2020-10-13
mock: own the cheat directory

praiskup:
There's nothing like cheat-filesystem, yet - and we don't want to depend
on cheat itself.

Resolves: rhbz#1887905


xsuchy:
I believe that a better place will be mock-filesystem as suggested in https://docs.fedoraproject.org/en-US/packaging-guidelines/#_the_directory_is_owned_by_a_package_which_is_not_required_for_your_package_to_function

xsuchy (approved):

PR #644, opened 2020-10-14
moving stuff out of util.py

xsuchy:
Because util.py is too long. See individual commits.


xsuchy:
I am going to ignore the remaining @pep8speaks warning as those bare-excepts are 1) pylint-disabled 2) copied from old code. And I do not want to disable this warning globally.

praiskup:
+1 to any future movement like that, feel free to push directly -- I neither want to block you nor I want to spend the time on reviewing this..., don't see large **benefits**, only **risks**.  Just please run the test suite once you are finished with your refactorings.

Such changes are proven to be error-prone, especially in mock where we have very minimal unit-test coverage (adding more coverage, while you are on it, would have huge benefits, though!).  So honetly, my personal rule is never do any refactorings just to make the code prettier.

PR #648, opened 2020-10-17
Introduce external BuildRequires

xsuchy:
This will allow special kinds of buildrequires. E.g.
`      BuildRequires: external:pypi:foo
`    
will install module "foo" from PyPI using pip3.
RPM db will be silenced by fake packages created by create-fake-rpm.
    
This feature is by default disabled.

This is a draft - it still does not work as intended (e.g. pip install does not have network :( )

What do you think?

Edit: SRPM for testing http://miroslav.suchy.cz/tmp/evil-3.31.1.r1-1.fc33.src.rpm This package BuildRequires external:pypi:copr


mock/py/mockbuild/backend.py
```diff
@@ -240,6 +249,59 @@ def installSpecDeps(self, spec_file):
...
+        finally:
+            self.uid_manager.restorePrivs()
+        self.install_fake_rpm('pypi', deps)
+
+    @traceLog()
+    def install_fake_rpm(self, external_type, deps):
+        """ Create and install fake packages using create-fake-rpm """
+        list_of_packages = []
+        for dep in deps:
+            command = ['create-fake-rpm', '--print-output',
```
TomasTomecek, line 292: wow, this is really interesting and impressive

Conan-Kudo:
@xsuchy Why not make the fake package actually do the install in a `%post` scriptlet?

xsuchy:
@Conan-Kudo interesting idea.

My intention, is that the create-fake-rpm will be removed one day. And rpm or dnf will undestand this 'external:*' stanza. But rpm guys to refuse to do that now. And dnf guys has tons of other problems that I do not even mentioned it to them yet. 

I would like to let it live here in Mock and if people will like it, then we can start looking towards rpm/dnf and maybe extend it to runtime deps as well.

mock/py/mockbuild/backend.py
```diff
@@ -240,6 +225,61 @@ def installSpecDeps(self, spec_file):
...
+                       '--build',
+                       'external-{0}-{1}'.format(external_type, dep),
+                       'external:{0}:{1}'.format(external_type, dep)]
+            output = self.bootstrap_buildroot.doChroot(
+                command, shell=False,
+                returnOutput=True,
+                printOutput=True,
+                returnStderr=False, raiseExc=True)
+            package = output.split()[0]
+            list_of_packages.append(package)
```
praiskup, line 280: There's new Suggests, while we run it eventually in bootstrap ... I don't know, perhaps you eventually want to build the fake RPM directly in the target chroot.

mock/py/mockbuild/backend.py
```diff
@@ -298,6 +338,14 @@ def build(self, srpm, timeout, check=True, spec=None):
                 raise Error('DynamicBuildRequires are found but support is disabled.'
                             ' See "dynamic_buildrequires" in config_opts.')
 
+            external_deps = [i for i in requires if i.startswith('external:')]
+            if external_deps:
+                if self.config.get('external_buildrequires'):
+                    self.install_external_deps(external_deps)
```
praiskup, line 344: I think this if-in-if condition would better be reversed, aka do nothing if  `external_buildrequires` is False.

mock/py/mockbuild/backend.py
```diff
@@ -298,6 +338,14 @@ def build(self, srpm, timeout, check=True, spec=None):
...
+                else:
+                    raise Error('ExternalBuildRequires are found but support is disabled.'
+                                ' See "external_buildrequires" in config_opts.')
```
praiskup, line 347: Ah, but then this wouldn't be shown ...ok.  Scratch the previous comment.

mock/py/mockbuild/backend.py
```diff
@@ -778,7 +826,7 @@ def install_build_results(self, results):
                 results_bindmount.mount()
             pkgs = [pkg for pkg in results if not pkg.endswith("src.rpm")]
             try:
-                self.install(*pkgs)
+                self.buildroot.install(*pkgs)
```
praiskup, line 829: Is this change needed?

mock/py/mockbuild/backend.py
```diff
@@ -240,6 +225,61 @@ def installSpecDeps(self, spec_file):
...
+        """ deps is list of python modules. Without the prefix. """
+        self.buildroot.install_as_root('/usr/bin/pip3', 'python3-setuptools')
+        command = ['pip3', "install"] + deps
+        try:
+            self.uid_manager.becomeUser(0, 0)
+            self.buildroot.doChroot(
+                command, shell=False,
+                nspawn_args=self._get_nspawn_args(),
+                returnOutput=True, unshare_net=False,
+                returnStderr=True, raiseExc=True)
```
praiskup, line 259: It is indeed inconvenient that we have to implement this logic in mock :-(

xsuchy:
rebased. But `pip3 install` still fails even when installed into bootstrap which has network enabled. And the run of `create-fake-rpm` fails and I have no idea why. :(

praiskup:
On by box it doesn't even try to execute the commands.

<details><summary>Traceback</summary>
<p>
```
ERROR: 'Commands' object has no attribute 'bootsstrap_buildroot'
  File "/usr/libexec/mock/mock", line 1059, in <module>
    exitStatus = main()
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 825, in main
    result = run_command(options, args, config_opts, commands, buildroot, state)
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 938, in run_command
    mockbuild.rebuild.do_rebuild(config_opts, commands, buildroot, options, srpms)
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.9/site-packages/mockbuild/rebuild.py", line 83, in do_rebuild
    rebuild_generic(srpms, commands, buildroot, config_opts, cmd=build,
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.9/site-packages/mockbuild/rebuild.py", line 25, in rebuild_generic
    ret = cmd(item)
  File "/usr/lib/python3.9/site-packages/mockbuild/rebuild.py", line 64, in build
    commands.build(srpm, timeout=config_opts['rpmbuild_timeout'],
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.9/site-packages/mockbuild/backend.py", line 359, in build
    self.install_external_deps(external_deps)
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.9/site-packages/mockbuild/backend.py", line 243, in install_external_deps
    self.install_external_deps_pypi(pypi_deps)
  File "/usr/lib/python3.9/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.9/site-packages/mockbuild/backend.py", line 253, in install_external_deps_pypi
    self.bootsstrap_buildroot.root_log.info("Executing %s", cmd_pretty(command))
AttributeError: 'Commands' object has no attribute 'bootsstrap_buildroot'
```
</p></details>

xsuchy:
Dammn last moment typo. Fixed. Now it ends with traceback because of create-fake-rpm returns empty output. While it prints one line.

mock/py/mockbuild/backend.py
xsuchy: Not necessary, but I find it more intuitive. Rather Buildroot.install() than Command.install().

xsuchy:
Updated and ready for review. I moved all the code about external deps to a separate file. Later we can spin-off a separate project from it (as suggested by @praiskup). For now, I would rather keep it here.
Beside pypi I added crates, but as I am not a rust person, I did not test it (it works only with binary packages and I am not sure which one is binary). If all this is ok, I would ask community with help with other languages (Go, rubygem...) and do that as separate PR.

mock/py/mockbuild/backend.py
praiskup: But it's not entirely equivalent.  The `Command.install` prints additional log output.  If we really want to change the output I'd prefer to have this in separate commit that is saying that we don't want those log entries.

mock/py/mockbuild/buildroot.py
```diff
@@ -866,3 +866,27 @@ def delete(self):
...
+        self.root_log.info("installing package(s): %s", " ".join(rpms))
+        output = self.pkg_manager.install(*rpms, returnOutput=1)
+        self.root_log.info(output)
+
+    @traceLog()
+    def remove(self, *rpms):
+        """Call package manager to remove the input rpms from the chroot"""
+        self.root_log.info("removing package(s): %s", " ".join(rpms))
+        output = self.pkg_manager.remove(*rpms, returnOutput=1)
+        self.root_log.info(output)
```
praiskup, line 892: Here is a bit harder to review if there is no unnecessary change, becaues the two methods were moved.  Can you keep them on the old place?

mock/integration-tests/test-F.spec
```diff
@@ -0,0 +1,29 @@
+Name:           test-F
+Version:        0
+Release:        1%{?dist}
+Summary:        Test of external builddeps
+
+License:        WTFPL
+
+BuildRequires:  external:pypi:copr
+BuildRequires:  external:pypi:bokeh
```
praiskup, line 9: Can you add also some crate BuildRequires?

mock/py/mockbuild/buildroot.py
```diff
@@ -866,3 +866,27 @@ def delete(self):
     @property
     def uses_bootstrap_image(self):
         return self.is_bootstrap and self.use_bootstrap_image
+
+    @traceLog()
+    def install_as_root(self, *deps):
+        """Figure out deps from srpm. Call package manager to install them"""
```
praiskup, line 872: I don't get this comment :-/ it seems like become-root wrapper around self.install() only?

mock/py/mockbuild/backend.py
praiskup: Ok, this eventually broke `make --install`, as I was right initially - the Commands.install is used in mock.py.

mock/py/mockbuild/external.py
```diff
@@ -0,0 +1,96 @@
...
+        if crate_deps:
+            self.install_external_deps_crate(crate_deps)
+
+        if deps:
+            raise ExternalDepsError("Unknown external dependencies: {}".format(', '.join(deps)))
+
+    @traceLog()
+    def install_external_deps_pypi(self, deps):
+        """ deps is list of python modules. Without the prefix. """
+        self.bootstrap_buildroot.install_as_root('/usr/bin/pip3', 'python3-setuptools')
```
praiskup, line 50: When bootstrap chroot is disabled, we need to stop building much sooner then here.

PR #652, opened 2020-10-22
do not TB if rpmbuild fails with exit code 11

xsuchy:
additionally, make the logging to build.log and root.log a bit better.

During the review, note that I added raiseExc=False which may not be directly visible from the diff.


praiskup:
Looks good, thanks.

PR #653, opened 2020-10-29
Don't swallow preexec_fn exceptions

praiskup:
If any exception is raised from preexec_fn, subprocess module just
swallows the exception, and prints non-explaining error:

    subprocess.SubprocessError: Exception occurred in preexec_fn.

Newly we temporarily catch the exception, and dump it to logs before we
re-raise it.

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -617,20 +617,24 @@ class ChildPreExec(object):
         getLog().debug("child environment: %s", env)
 
     def __call__(self, *args, **kargs):
-        if not self.shell:
-            os.setsid()
-        os.umask(0o02)
-        condUnshareNet(self.unshare_net)
-        condPersonality(self.personality)
-        condEnvironment(self.env)
-        # Even if nspawn is allowed to be used, it won't be used unless there
-        # is a chrootPath set
-        if not USE_NSPAWN or not self.chrootPath:
-            condChroot(self.chrootPath)
-            condDropPrivs(self.uid, self.gid)
-            condChdir(self.cwd)
-        condUnshareIPC(self.unshare_ipc)
-        reset_sigpipe()
+        try:
+            if not self.shell:
+                os.setsid()
+            os.umask(0o02)
+            condUnshareNet(self.unshare_net)
+            condPersonality(self.personality)
+            condEnvironment(self.env)
+            # Even if nspawn is allowed to be used, it won't be used unless there
+            # is a chrootPath set
+            if not USE_NSPAWN or not self.chrootPath:
+                condChroot(self.chrootPath)
+                condDropPrivs(self.uid, self.gid)
+                condChdir(self.cwd)
+            condUnshareIPC(self.unshare_ipc)
+            reset_sigpipe()
+        except:  # pylint: disable=bare-except
+            getLog().exception("Exception raised in ChildPreExec")
+            raise
 
 
 def setup_operations_timeout(config_opts):


praiskup:
The __call__ method is a mess IMO, we mostly rely on the fact that no exception is raised - otherwise the mock state ends up in an inconsistent state (but we should fail horribly in such situation anyway).

xsuchy (approved):

praiskup:
I was told that this PR doesn't help with the diagnosis, so I need to double check.

praiskup:
No no, this indeed can not work.  Preexec_fn is executed in sub-process, and as such everything we log-out or print to stdout/err would have to be processed by logOutput (which we don't do in case of any exception in preexec_fn).

praiskup:
The traceback is somewhat confusing, and broken ....  But if the numbers are correct, the prbolem was in:
`frame = inspect.getouterframes(inspect.currentframe())[1][0]`
I don't know, is this expected to always work? @torsava can you please take a quick look
if that is familiar to you (or anyone else from python team)?

praiskup:
Meh, sorry @torsava, the comment about the traceback was related to pull request #655.

hroncok:
> can you please take a quick look if that is familiar to you (or anyone else from python team)?

Not to me, sorry.

encukou:
Not familiar to me either :(

mock/py/mockbuild/util.py
```diff
@@ -617,20 +617,24 @@ def __init__(self, personality, chrootPath, cwd, uid, gid, env=None,
...
+            condEnvironment(self.env)
+            # Even if nspawn is allowed to be used, it won't be used unless there
+            # is a chrootPath set
+            if not USE_NSPAWN or not self.chrootPath:
+                condChroot(self.chrootPath)
+                condDropPrivs(self.uid, self.gid)
+                condChdir(self.cwd)
+            condUnshareIPC(self.unshare_ipc)
+            reset_sigpipe()
+        except:  # pylint: disable=bare-except
```
hroncok, line 635:
```suggestion
        except BaseException:
```

vstinner:
I'm not aware of this issue. preexec_fn is fragile. handling preexec_fn exceptions in subprocess is fragile. So i'm not surprised that you get such issue. I guess that Python failed to format the exception when trying to send it to the parent process. preexec_fn is run in the child process. if it fails, Python serializes the traceback in a pipe which is read by the parent process. Here it is seems like formatting the exception failed, but there is no way for Python to report the new exception while serialiazing the exception.

To debug your issue, I suggest to manually handle exceptions in your preexec_fn. Something like:

```
import traceback

LOG = open("/tmp/log.txt", "w", encoding="utf-8", errors="replace")

def real_func(): ...

def wrapper():
  try:
    real_func()
  except:
    print("real_func() failed!", file=LOG)
    traceback.print_exc(file=LOG)
    print("^^ traceback", file=LOG)
    LOG.flush()
    raise
```

Pass `wrapper` rather than real_func() to preexec_fn parameter of subprocess.Popen.

If you manage to understand why Python failed to serialize the exception **name** in subprocess, I'm curious to know more about it ;-)

praiskup:
Thanks for taking a look!

And sorry, I did not C&P it correctly, here is the full trace:

```
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 822, in __call__
    log.debug("run: cmd = %s\n", cmd)
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 520, in condUnshareIPC
    logger.info("Child dead, killing orphans")
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 461, in unshare
    # https://github.com/rpm-software-management/mock/issues/113
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 22, in __init__
    # works around problem where reconfiguring the logging module means loggers
  File "/usr/lib64/python3.6/inspect.py", line 1478, in getouterframes
    frameinfo = (frame,) + getframeinfo(frame, context)
  File "/usr/lib64/python3.6/inspect.py", line 1452, in getframeinfo
    lines, lnum = findsource(frame)
  File "/usr/lib64/python3.6/inspect.py", line 828, in findsource
    if pat.match(lines[lnum]): break
IndexError: list index out of range
```

> Also, more context might help. Especially:
>
>    What are the symptoms you talk about?

I mean, some traceback is triggered by trying to access logging structures
inside preexec_fn, and the broken traceback (seemingly?) -- e.g. it shows one
comment instead of real code, ...

>    How to reproduce the problem?

Get EL8 s390x machine, and try to build the latest rhpkg package with mock.  And
you have to apply the fix @vstiner suggested in the other PR to get this broken
traceback.

praiskup:
Line 22 looks like:
https://github.com/rpm-software-management/mock/blob/e84a337a61961660b5dfa37d50b7c5ade1fb7c62/mock/py/mockbuild/trace_decorator.py#L22

praiskup:
Meh, I'm going to close this PR, since the fix is entirely wrong - and keep only #655.

mock/py/mockbuild/util.py
praiskup: Thanks, should I note this as prefered way to catch **any** exception?
FWIW, tihs ends up with a slightly different PyLint error:
`broad-except: Catching too general exception BaseException`

hroncok: Yes, this is the preferred way. However since pylint seem to dislike it anyway, I have no idea what's the preferred way for pylint.

vstinner:
The issue seems to be in trace_decorator which doesn't catch exceptions.

inspect.findsource() raises an IndexError, likely because the .py file changed compared to the run code.

praiskup:
> inspect.findsource() raises an IndexError, likely because the .py file changed compared to the run code.

What kind of error in mock code can cause that?

vstinner:
Maybe you run a "old" .pyc  file whereas the ".py" file was updated. I don't know.

trace_decorator should attempt to catch such error.

praiskup:
I'm 99% sure I removed the pyc file, but I'll retry if I'll have a chance. @tkopecek is the s390 box still usable for testing?

PR #655, opened 2020-11-04
Hotfix preexec_fn problem on RHEL 8 s390x

praiskup:
This symptoms look like a combination of several problems.

It was pretty complicated to debug (we can not easily see tracebacks
from subprocess's if we set stdout/err to subprocess.PIPE and handling
that in logOutput() too late);  with the help of traceback.print_exc(),
I catched the error in ChildPreExec.__call__(), and it looked like:

```
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 821, in __call__
    log = getLog()
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 520, in condUnshareIPC
    logger.info("Child dead, killing orphans")
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 461, in unshare
    # https://github.com/rpm-software-management/mock/issues/113
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 22, in __init__
    # works around problem where reconfiguring the logging module means loggers
  File "/usr/lib64/python3.6/inspect.py", line 1478, in getouterframes
    frameinfo = (frame,) + getframeinfo(frame, context)
  File "/usr/lib64/python3.6/inspect.py", line 1452, in getframeinfo
    lines, lnum = findsource(frame)
  File "/usr/lib64/python3.6/inspect.py", line 828, in findsource
    if pat.match(lines[lnum]): break
```

Which looks like a Python bug?  But doing getLog() inside the preexec_fn
wouldn't probably work anyways (separate process, not sharing output
file descriptors).

Relates: #653

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -268,7 +268,7 @@ def getAddtlReqs(hdr, conf):
 
 @traceLog()
 def unshare(flags):
-    getLog().debug("Unsharing. Flags: %s", flags)
+    #getLog().debug("Unsharing. Flags: %s", flags)
     try:
         res = _libc.unshare(flags)
         if res:


encukou:
The traceback looks incomplete – it's missing the exception type/message at the end.

hroncok:
Also, more context might help. Especially:

 - What are the symptoms you talk about?
 - How to reproduce the problem?

vstinner:
See discussion at https://github.com/rpm-software-management/mock/pull/653 about the incomplete traceback (missing exception).

praiskup:
Thanks for taking a look!

And sorry, I did not C&P it correctly, here is the full trace:

```
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 822, in __call__
    log.debug("run: cmd = %s\n", cmd)
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 520, in condUnshareIPC
    logger.info("Child dead, killing orphans")
  File "/usr/lib/python3.6/site-packages/mockbuild/util.py", line 461, in unshare
    # https://github.com/rpm-software-management/mock/issues/113
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 22, in __init__
    # works around problem where reconfiguring the logging module means loggers
  File "/usr/lib64/python3.6/inspect.py", line 1478, in getouterframes
    frameinfo = (frame,) + getframeinfo(frame, context)
  File "/usr/lib64/python3.6/inspect.py", line 1452, in getframeinfo
    lines, lnum = findsource(frame)
  File "/usr/lib64/python3.6/inspect.py", line 828, in findsource
    if pat.match(lines[lnum]): break
IndexError: list index out of range
```

> Also, more context might help. Especially:
>
>    What are the symptoms you talk about?

I mean, some traceback is triggered by trying to access logging structures
inside preexec_fn, and the broken traceback (seemingly?) -- e.g. it shows one
comment instead of real code, ...

>    How to reproduce the problem?

Get EL8 s390x machine, and try to build the latest rhpkg package with mock.  And
you have to apply the fix @vstiner suggested in the other PR to get this broken
traceback.

praiskup:
The guilty line 22 triggering the traceback is here:
  https://github.com/rpm-software-management/mock/blob/e84a337a61961660b5dfa37d50b7c5ade1fb7c62/mock/py/mockbuild/trace_decorator.py#L18-L29

praiskup:
I think that @tkopecek could give access to the affected s390 box also to anyone from the Python team (if anyone is interested).

praiskup:
I mereged this 776a74686dea49a29836c6371237bcde3e867816 to fix the
tracebacks on EL8 s390x.  I'm going to keep this open for a while, it would
be nice to see (a) what is going on there in the inspect module (provides a
broken traceback output) and (b) why is the traceback actually generated
at all because this only happens on one platform.  But to debug those, we
probably need more minimal reproducer so we are sure this is not a problem
in Mock.

praiskup:
This was automatically closed by GitHub while we moved to the new `main` branch name.  The issue is still actual, but at least Mock doesn't trigger the Python bug as far as I can tell, so this is low priority for this project.

PR #658, opened 2020-11-11
Add missing CRB repository

jkonecny12:
Without this a plenty of build dependencies are missing. At least Anaconda can't be build.

When building Anaconda I'm getting https://github.com/rhinstaller/anaconda/pull/2868#issuecomment-725525020 most of these dependencies are part of the CRB and from my understanding the CRB repository should be part of build-root which is exactly what mock is doing.

However, even with these I'm missing:
- gtk3-devel-docs
- libtimezonemap-devel

Wonder where these are...

@mmathesius could you please verify validity of this PR?


xsuchy (approved):

mmathesius:
LGTM.

praiskup:
Thank you

PR #659, opened 2020-11-12
ELN should use for build Everything repository

jkonecny12:
ELN BaseOS, AppStream and CRB repositories are for runtime. However, as in RHEL there are even packages which are just in the buildroot. These packages are not part of the repositories above but only Everything repository.

So to reflect how the next RHEL will be build we should use only the Everything repository which should contain all the packages not just ones which will be shipped.

Hopefully, this should be the ultimate solution for our ELN missing packages problem. @sgallagher @mmathesius could you please do a review for me here? Thanks!

The everything repository have the missing packages required for Anaconda build.

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-eln.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-eln.tpl
@@ -31,6 +31,9 @@ module_platform_id=platform:eln
 protected_packages=
 
 # repos
+# The Everything repository have to be enabled to get all of the packages because
+# some of them are not present anywhere else. Also the AppStream repository
+# have to be enabled because it contain modules. Everything else is not required.
 [local]
 name=local
 baseurl=https://kojipkgs.fedoraproject.org/repos/eln-build/latest/$basearch/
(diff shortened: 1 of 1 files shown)


sgallagher (approved):
LGTM

praiskup:
Are modules provided in Everything?

mock-core-configs/etc/mock/templates/fedora-eln.tpl
```diff
@@ -60,83 +60,13 @@ gpgcheck=1
...
 gpgkey=file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary
 skip_if_unavailable=False
-
-
-
-[eln-appstream]
-name=Fedora - ELN AppStream - Developmental packages for the next Enterprise Linux release
-baseurl=https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/AppStream/$basearch/os/
-#metalink=https://mirrors.fedoraproject.org/metalink?repo=eln&arch=$basearch
-enabled=1
```
praiskup, line 80: I'd propose to keep them disabled, if this is really about to be disabled.

sgallagher:
Oh, good point. I think you are right: modules only show up in AppStream.

jkonecny12:
Hmm, so I guess the best option would be to have all `BaseOS`, `AppStream`, `CRB` and `Everything` together?

sgallagher:
> Hmm, so I guess the best option would be to have all `BaseOS`, `AppStream`, `CRB` and `Everything` together?

Probably we should have at least `AppStream` and `Everything` enabled, with `BaseOS` and `CRB` present but disabled in the configuration.

jkonecny12:
> > Hmm, so I guess the best option would be to have all `BaseOS`, `AppStream`, `CRB` and `Everything` together?
> 
> Probably we should have at least `AppStream` and `Everything` enabled, with `BaseOS` and `CRB` present but disabled in the configuration.

OK, will change the PR.

TomasTomecek:
I'm wondering, are we able to enable/disable certain repositories defined in a mock config in COPR for our builds?

jkonecny12:
@TomasTomecek exactly. These mock configuration files are used in the COPR so these are the packages used by your Packit service.

praiskup:
> I'm wondering, are we able to enable/disable certain repositories defined in a mock config in COPR for our builds?

No, there's still an RFE for that.  But the mock-core-configs defaults
should be reasonable by default so people shouldn't need to explicitly
disable particular repositories at least (and adding additional
repositories is implemented in Copr, even though you need to specify
baseurl).

@jkonecny12 thank you for the update.  I don't know, e.g. check how rhel-8
chroot is configured -- there's baseos+appstream+crb -- and the fact that
the buildroot is not complete is just a design decision (IMO one of the
worst, we should have the "unsupported" complementary repo on top of CRB).
I'm worried here that fedora-ELN is getting too far from real enterprise
linux, but if you say that appstream+everything is the correct way here,
I'll probably merge.  Has anyone tested the mock-core-configs from this
PR built by Travis?

jkonecny12:
I've updated the PR. Enabling just `Everything` and `AppStream`. Tested locally on Anaconda and the build went fine.

sgallagher:
> I'm worried here that fedora-ELN is getting too far from real enterprise
> linux, but if you say that appstream+everything is the correct way here,
> I'll probably merge.

This is actually pretty close, as "AppStream+Everything" is basically identical to the buildroot being used internally. So I think this is the right approach.

jkonecny12:
Just tested the travis built by reinstalling the package on my system and everything went well for Anaconda build.

jkonecny12:
Ping, could we please merge this? It's blocking our projects from testing builds on ELN.

praiskup:
Thank you for the PR, I can't promise more than merge this week, though.

jkonecny12:
Seems that this finally enabled Anaconda build on COPR ELN. 

https://github.com/rhinstaller/anaconda/pull/2868

Thanks a lot everyone.

PR #665, opened 2020-11-26
suggests create-fake-rpm

xsuchy:
this complements
https://github.com/rpm-software-management/mock/pull/648


praiskup:
I don't think create-fake-rpm is actually needed on host.

xsuchy:
Ahh, right.

PR #677, opened 2020-12-11
file_util: Improve an error message

tbaederr:
Print the path we've tried to call rmtree() on as well. Otherwise it's
hard to know where the faulty symbolic link even is.


praiskup:
Thank you!

Issue #678, opened 2020-12-14
regression: --isolation=nspawn doesn't in 2.7


cgwalters:
See https://pagure.io/releng/failed-composes/issue/2126

With `mock-2.7-1.fc33.noarch`, if I do `mock --new-chroot --shell` I am clearly not running in nspawn. Downgrading to `2.6` fixes it - I again see e.g. `container=systemd-nspawn` in the environment, and it shows up in `machinectl` etc.

cgwalters:
I tried to debug this but AFAICS at least as far as the parsing `config.py` `USE_NSPAWN` is `True`...but it somehow gets lost later?

AdamWill:
copy/pasted from the releng issue:

So my first suspect here is https://github.com/rpm-software-management/mock/commit/877b332610173f849bde8fbccd98abb2dda4345b , which moves around some of the relevant bits here. In particular there's some weirdness related to a var called `USE_NSPAWN`. Before that commit, it was only ever declared in `mock/py/mockbuild/util.py`; a few other things imported it from there. That commit moves *some* of the logic related to it into `mock/py/mockbuild/config.py`, but the result is kinda odd. It's still declared in `util.py` , from where `config.py` imports it and sets it as a global and does stuff to it. Other things still import it from `util`, not from `config`.

I haven't entirely traced out the consequences, but I'd bet a medium sized amount of money that's the problem.

AdamWill:
yeah, so I'm pretty sure that when `config` imports `USE_NSPAWN` from `util` and runs through various logic to set it, it's only setting `config.USE_NSPAWN`. Nothing it does changes `util.USE_NSPAWN`. So far as `util` itself and anything else that imports `USE_NSPAWN` from util is concerned, `USE_NSPAWN` is now just always `False` (as it's set `False` at the start of `util.py` and never changed).

AdamWill:
https://github.com/rpm-software-management/mock/pull/679 should fix this. I did check my work a bit: I hacked up everything in `util` that depends on the value of `USE_NSPAWN` to log it. Before that PR, even when running with `--new-chroot`, they all logged it as false. With the changes from my PR, they all log it as true. If I run `mock -r fedora-rawhide-x86_64 --new-chroot --shell` I can see a `systemd-nspawn` command running.

PR #679, opened 2020-12-14
Fix use of nspawn (#678)

AdamWill:
877b332 broke this by splitting the code that actually decides
what `USE_NSPAWN` should be into the `config` module...where it
now only set `config.USE_NSPAWN`, but everything that actually
changes behaviour based on the value reads `util.USE_NSPAWN`,
which was now always False.

This feels a bit ugly, but should at least solve the problem in
a safe way, by having the parsing code in `config` update the
value of the actual variable in `util`.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -23,7 +23,7 @@ from . import text
 from .file_util import is_in_dir
 from .trace_decorator import getLog, traceLog
 from .uid import getresuid
-from .util import USE_NSPAWN, setup_operations_timeout
+from .util import set_use_nspawn, setup_operations_timeout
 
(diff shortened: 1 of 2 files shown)


AdamWill:
BTW, it concerns me that the test suite apparently didn't catch this.

cgwalters (approved):
Only skimmed but seems sane to me

praiskup:
Thank, good catch and the fix looks OK.  Merged as 1dfda6358b882a01d31451957b47e19f94217acf.

> BTW, it concerns me that the test suite apparently didn't catch this.

Nah, the testsuite takes several hours, and no hit here.  Apparently mock "just worked" but
with different kind of isolation.  I added a simple test case in f48e5f91d560092069a817646cfe9a1a1b2abb1c.
Thank you again.

PR #681, opened 2020-12-18
rpkg_preprocessor: Add a force_enable option

tstellar:
This option tells rpkg_preprocessor to ignore rpkg.conf and always
preprocess the spec file.  This is useful for testing mass package
changes where it's not practical to add an rpkg.conf to every package.


praiskup:
@clime do you want to take a look at review?

clime:
It looks OK to me.

praiskup:
k, thank you!

PR #686, opened 2021-01-11
Fix bootstrapping of newer Fedora on EL7

praiskup:
We forgot to add Fedora 33 host override when adding other Fedora 33
configs.  This is something which would definitely repeat in future, so
I'm rather using a script that will do this hack for us automatically in
the future.

While I'm on it I'm removing rhel-8 override without replacement because
in the meantime the RPM in EL8 got the libzstd support.

Resolves: rhbz#1914750

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -53,7 +53,6 @@ Config files which allow you to create chroots for:
 
 
 %build
-cd etc/host-overrides
 HOST=none
 %if 0%{?fedora}
 HOST="fedora-%{fedora}"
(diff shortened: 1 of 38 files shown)


mock-core-configs/mock-core-configs.spec
```diff
@@ -62,13 +62,20 @@ HOST="fedora-%{fedora}"
...
+# host overrides
+case $HOST in
+  rhel-7)
+    # RPM on EL7 doesn't link against libzstd, and newer Fedora is compressed
+    # using ZSTD.  We need to enable bootstrap image here to be able to
+    # initialize the Fedora bootstrap chroot.
+    for config in etc/fedora-*-*.cfg; do
+        version=$(echo "$config" | cut -d- -f2)
+        if test $version = rawhide || test $version -ge 31; then
+            echo "config_opts['use_bootstrap_image'] = True" >> "$config"
```
xsuchy, line 74: Why not outputting it to host-overrides?

praiskup: Because we did not install host overrides in the end, we actually only catenated the contents to the default configs.

PR #688, opened 2021-01-25
configs: add RHEL 6 x86_64 configuration

praiskup:
I've seen people complaining how hard it is to build against ELS rhel6.  While I don't
think we should encourage anyone (therefor EOL subdir), some circles will
work with EL6 for some time and this might help them.


xsuchy:
??? There is `eol/epel-6-x86_64.cfg`, which should work.

praiskup:
>  ??? There is eol/epel-6-x86_64.cfg, which should work.

This is the RHEL one using the subscription.

xsuchy (approved):

PR #695, opened 2021-01-30
Rename centos-stream centos-8-stream

opoplawski:
We're going to have centos-9-stream, etc..


praiskup:
I think this is likely, but do you have an official statement this is going to happen?

Conan-Kudo:
@praiskup @carlwgeorge has confirmed that is what is happening.

praiskup:
Ok, thanks.  @opoplawski can you please change the name to `centos-stream-8-<arch>`?

carlwgeorge:
> I think this is likely, but do you have an official statement this is going to happen?

https://centos.org/distro-faq/#q6-will-there-be-separateparallelsimultaneous-streams-for-8-9-10-etc

carlwgeorge:
> Ok, thanks. @opoplawski can you please change the name to `centos-stream-8-<arch>`?

Yes please, the distro is CentOS Stream, and the version is 8.  I know the distgit branches are currently named c8s, but that was some regex technical debt and shouldn't be used to draw any conclusions.  `centos-stream-8-<arch>` is the best name to use for mock build roots.

opoplawski:
Renamed.

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
```diff
@@ -57,63 +57,63 @@ gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Officia
 
 [Stream-base-debuginfo]
 name=CentOS-Stream - Debuginfo
-baseurl=http://debuginfo.centos.org/8-stream/$basearch/
+baseurl=http://debuginfo.centos.org/stream-8/$basearch/
```
praiskup, line 60: These repositories give me error 404.  Should they stay in the `8-stream` format?

carlwgeorge: We haven't started publishing debuginfo packages for CS8 yet.  An 8-stream directory exists, but it's empty.  That will likely be the correct location once we do start publishing debuginfo packages.

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
```diff
@@ -4,7 +4,7 @@ config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'registry.centos.org/centos:8'
-config_opts['dnf_vars'] = { 'stream': '8-stream',
+config_opts['dnf_vars'] = { 'stream': 'stream-8',
```
carlwgeorge, line 7: This variable should stay `8-stream`, as it's used by the mirrorlist service.

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
carlwgeorge: In fact all the baseurls should stick with `8-stream` for now.

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
arrfab: yes, that's what the Stream came with and as mirrorlist will pick path appearing on mirror network, let's stick with 8-stream (as that's also how it's done on Stream itself for mirrorlist .. let's not confuse more :)

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
praiskup: Yep, that's what I meant.

opoplawski:
Ah, well now we know where I got the idea to name things "8-stream" in the first place.  Hopefully things are correct now.

carlwgeorge (approved):

Conan-Kudo (approved):

praiskup:
thank you

PR #703, opened 2021-02-13
new core-config: support almalinux

prymar56:
Signed-off-by: Mark Pryor <pryorm09@gmail.com>These are the 2 files, *tpl and *cfg, to support almalinux in latest mock.

### DIFF
--- a/mock-core-configs/etc/mock/alma-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/alma-8-x86_64.cfg
@@ -1,24 +1,8 @@
 include('/etc/mock/templates/alma-8.tpl')
-include('/etc/mock/templates/epel-8.tpl.rpmnew')
+include('/etc/mock/templates/epel-8.tpl')
 
-config_opts['root'] = 'alma-8-x86_64'
 config_opts['target_arch'] = 'x86_64'
 config_opts['legal_host_arches'] = ('x86_64',)
 config_opts['hostname'] = 'mock'
 config_opts['environment']['HOSTNAME'] = 'mock'
-
-config_opts['files']['etc/dnf/modules.d/virt.module'] = """[virt]
-name=virt
-profiles=
-state=disabled
-stream=
-"""
-
-config_opts['releasever'] = '8'
-config_opts['package_manager'] = 'dnf'
-
-config_opts['use_bootstrap'] = False
-config_opts['use_bootstrap_image'] = False
-
-config_opts['chroot_setup_cmd'] += ' rpm-build nano dnf dnf-plugins-core glibc-langpack-en '
 
--- a/mock-core-configs/etc/mock/alma-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/alma-8-x86_64.cfg
@@ -1,0 +1,24 @@
+include('/etc/mock/templates/alma-8.tpl')
+include('/etc/mock/templates/epel-8.tpl.rpmnew')
+
+config_opts['root'] = 'alma-8-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
+config_opts['hostname'] = 'mock'
+config_opts['environment']['HOSTNAME'] = 'mock'
+
+config_opts['files']['etc/dnf/modules.d/virt.module'] = """[virt]
+name=virt
+profiles=
+state=disabled
+stream=
+"""
+
+config_opts['releasever'] = '8'
+config_opts['package_manager'] = 'dnf'
+
+config_opts['use_bootstrap'] = False
+config_opts['use_bootstrap_image'] = False
+
+config_opts['chroot_setup_cmd'] += ' rpm-build nano dnf dnf-plugins-core glibc-langpack-en '
+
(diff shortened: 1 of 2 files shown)


mock-core-configs/etc/mock/alma-8-x86_64.cfg
```diff
@@ -0,0 +1,24 @@
+include('/etc/mock/templates/alma-8.tpl')
+include('/etc/mock/templates/epel-8.tpl.rpmnew')
```
Conan-Kudo, line 2: This file doesn't exist.

mock-core-configs/etc/mock/alma-8-x86_64.cfg
```diff
@@ -0,0 +1,24 @@
...
+
+config_opts['root'] = 'alma-8-x86_64'
```
Conan-Kudo, line 4: This can be put into the template file and structured to change correctly per-architecture.

mock-core-configs/etc/mock/alma-8-x86_64.cfg
```diff
@@ -0,0 +1,24 @@
...
+config_opts['legal_host_arches'] = ('x86_64',)
+config_opts['hostname'] = 'mock'
+config_opts['environment']['HOSTNAME'] = 'mock'
+
+config_opts['files']['etc/dnf/modules.d/virt.module'] = """[virt]
+name=virt
+profiles=
+state=disabled
+stream=
+"""
```
Conan-Kudo, line 15: This needs to be dropped.

mock-core-configs/etc/mock/alma-8-x86_64.cfg
```diff
@@ -0,0 +1,24 @@
...
+
+config_opts['releasever'] = '8'
+config_opts['package_manager'] = 'dnf'
+
+config_opts['use_bootstrap'] = False
+config_opts['use_bootstrap_image'] = False
+
+config_opts['chroot_setup_cmd'] += ' rpm-build nano dnf dnf-plugins-core glibc-langpack-en '
```
Conan-Kudo, line 23: This is redundant with what you have in the template file and should be dropped.

mock-core-configs/etc/mock/templates/alma-8.tpl
```diff
@@ -0,0 +1,119 @@
...
+enabled=1
+gpgcheck=0
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux
+
+[extras]
+name=AlmaLinux 8.3 - Extras
+baseurl=https://repo.almalinux.org/almalinux/8/extras/$basearch/os/
+enabled=1
+gpgcheck=0
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux
```
Conan-Kudo, line 61: Several issues here:

* This references AlmaLinux 8.3 but doesn't set a fixed path, should it be "AlmaLinux 8"?
* GPG check is turned off
* Path to the GPG key file is invalid, the key needs to be added to [xsuchy/distribution-gpg-keys](https://github.com/xsuchy/distribution-gpg-keys) like all the others have

Conan-Kudo (changes requested):
Thanks for making this submission!
 
 Here's some initial feedback on how to get this into better shape to be included.

xsuchy:
I just added Alma to distribution-gpg-keys in https://github.com/xsuchy/distribution-gpg-keys/commit/492dbff707a8c08461bb1cd6d066e2f9860278b1
Please use the path `/usr/share/distribution-gpg-keys/RPM-GPG-KEY-AlmaLinux`
I will release new version of distribution-gpg-keys together with new mock-core-config as soon as this PR is merged - and I agree with all @Conan-Kudo comments.

mock-core-configs/etc/mock/alma-8-x86_64.cfg
```diff
@@ -0,0 +1,8 @@
+include('/etc/mock/templates/alma-8.tpl')
+include('/etc/mock/templates/epel-8.tpl')
+
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
+config_opts['hostname'] = 'mock'
+config_opts['environment']['HOSTNAME'] = 'mock'
```
xsuchy, line 7: Why these two lines about `hostname`? I guess this should go away unless you have some strong reason.

Conan-Kudo:
Generally, it might make sense to copy the `centos-8-*` configs and make the Alma Linux ones based on that.

prymar56:
while testing almalinux mock configs on fc33, I noticed that without enabling bootstrap, the rpmdb in /var/lib/rpm is sqlite.
This will break rpm when using the alma chroot for tests. Solution is to convert it from fc33.

Run this post-instll from fc33:
sudo rpmdb --rebuilddb --define "_db_backend bdb" --root /var/lib/mock/alma-8-x86_64/root

Conan-Kudo:
Yeah, Mock's default behavior is to use bootstrap mode, so this shouldn't be an issue in practice.

praiskup:
> Generally, it might make sense to copy the centos-8-* configs and make the Alma Linux ones based on that.

This makes sense to me as well.

xsuchy:
@prymar56 ping, any progress on fixing issues mentioned in comments?

Conan-Kudo:
@xsuchy @praiskup Due to lack of response by the submitter and the new release of Alma 8 stable, I've made #716 to replace this.

xsuchy:
Closing this one in favour of #716

Issue #702, opened 2021-02-11
The ELN repo is signed by old F34 keys


praiskup:
So updated distribution-gpg-keys disallow us to build against ELN builds, because rawhide == F35 now.
Reported here: https://pagure.io/releng/issue/10007

But perhaps we could accept both Fedora <LATEST> and Fedora <LATEST -1> keys here?  To always
give the fedora releng people a chance to update the chroots?

praiskup:
Per releng issue 10007 it seems that it's never going to be an instant step to re-sign ELN.  So we should
have the same approach as in rawhide configs -- use the latest two gpg keys.

praiskup:
The ELN compose takes longer than expected (more than 10 days now) so we should probably
prioritize this issue.

PR #704, opened 2021-02-22
configs: use Fedora N-1 gpg keys for ELN

praiskup:
Related discussions:
https://pagure.io/releng/issue/10007
https://github.com/fedora-eln/eln/issues/28
https://bugzilla.redhat.com/show_bug.cgi?id=1931095

Fixes: #702

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-eln.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-eln.tpl
@@ -1,4 +1,5 @@
 config_opts['releasever'] = 'eln'
+config_opts['eln_rawhide_releasever'] = '35'
 
 config_opts['root'] = 'fedora-eln-{{ target_arch }}'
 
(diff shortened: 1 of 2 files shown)


mock-core-configs/etc/mock/templates/fedora-eln.tpl
```diff
@@ -49,7 +57,7 @@ enabled=1
 repo_gpgcheck=0
 type=rpm
 gpgcheck=1
-gpgkey=file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary
+gpgkey={{ rawhide_gpg_keys() }}
```
xsuchy, line 60: It seems to my too codish for config. While this works and it is better as the definition is on one place only, I would prefer:
```
gpgkey=file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-{{ eln_rawhide_releasever|int }}-primary file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-{{ eln_rawhide_releasever|int -1}}-primary 
```

praiskup: The code is taken from the rawhide template, and I really like to have this defined in one place.  And I hate long lines ...  but will not fight too much here :-) you tell..

I actually thought that I could cut the code out into a separate template used both by rawhide and eln templates.
But I gave up as we can not use releasever here, and I'd have to touch the rawhide template.

xsuchy: > will not fight too much here

me neither :)

praiskup:
Ok, lemme release it.

PR #716, opened 2021-04-02
configs: Add AlmaLinux 8 configs

Conan-Kudo:
For now, only x86_64 is provided, but when more architectures become
available, it should be very easy to add them.

Obsoletes #703


xsuchy (approved):

xsuchy:
Thank you.

PR #722, opened 2021-04-16
mock-core-configs: add configs for the CentOS Hyperscale SIG

davide125:
Add mock configs for building packages against the Hyperscale SIG repos in CentOS Stream. The GPG key was previously added in https://github.com/xsuchy/distribution-gpg-keys/pull/44

### DIFF
--- a/mock-core-configs/etc/mock/centos-stream-hyperscale-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-hyperscale-8-x86_64.cfg
@@ -1,0 +1,17 @@
+include('templates/centos-stream-8.tpl')
+
+config_opts['root'] = 'centos-stream-hyperscale-8-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
+
+config_opts['chroot_setup_cmd'] += " centos-release-hyperscale"
+
+config_opts['dnf.conf'] += """
+
+[Stream-hyperscale]
+name=CentOS-Stream - Hyperscale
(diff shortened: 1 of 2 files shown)


Conan-Kudo (changes requested):
This should probably be called `centos-stream-8-hyperscale` instead of `hyperscale-8`. Otherwise there's no context for this config.

Conan-Kudo:
@davide125 Alternatively, we might want to have `centos-sig-mock-configs` package for all the SIGs there. Typically we've avoided all these layered things in `mock-core-configs`.

davide125:
@Conan-Kudo yeah that works for me, I mostly just want to have these in a place easy to access. Would `mock-centos-sig-configs` still be a part of mock like `mock-core-configs` is, or would that be a wholly separate thing?

Conan-Kudo:
@davide125 It would be a totally separate project, and it would layer on top of `mock-core-configs`. I did this with `mock-mageia-configs` for Mageia: https://svnweb.mageia.org/packages/cauldron/mock-mageia-configs/current/

davide125:
Ok, I've created https://pagure.io/centos-sig-hyperscale/mock-centos-sig-configs for this, let's continue there. Thanks!

xsuchy:
@Conan-Kudo @davide125 It will be nice if you can put in your spec file:

Enhances: mock-core-configs

This is reverse dependency to `Suggests`. See https://docs.fedoraproject.org/en-US/packaging-guidelines/WeakDependencies/

davide125:
thanks @xsuchy, I've added that in mock-centos-sig-configs-0.1-2: https://src.fedoraproject.org/rpms/mock-centos-sig-configs/c/2fd0ee62dffc058c486a873b5a161f773ab6bc10?branch=rawhide

PR #723, opened 2021-04-24
config: Add openSUSE Tumbleweed s390x config

Conan-Kudo:
openSUSE Tumbleweed has a IBM Z (s390x) port that mostly works now,
so we should have the ability to build for it.


praiskup:
Thank you.

PR #724, opened 2021-04-24
config: Move openSUSE Leap 15.1 to EOL and add openSUSE Leap 15.3

Conan-Kudo:
This PR moves openSUSE Leap 15.1 to the EOL path and adds openSUSE Leap 15.3.

New with openSUSE Leap 15.3 is official DNF-based containers, which makes environment bootstrap in Mock *much* faster and easier.


praiskup:
Thank you!

PR #725, opened 2021-04-25
Add Oracle Linux configuration

Conan-Kudo:
This PR adds configuration for building with Oracle Linux.

@Djelibeybi I was unable to find a `-latest` equivalent for the UEK repos to add to this. Is there a possibility of having one of those so that I can add optional UEK repo configuration into here? That would make it tremendously easier for people building kernel modules against UEK.

### DIFF
--- a/mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
+++ b/mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
@@ -1,0 +1,66 @@
+# This list is taken from 'epel-7-x86_64' @buildsys-build group, minus the
+# 'epel-*' specific stuff.
+config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
+
+config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
+config_opts['releasever'] = '7'
+config_opts['bootstrap_image'] = 'docker.io/library/oraclelinux:7'
+config_opts['package_manager'] = 'yum'
+
+config_opts['yum_install_command'] += " --disablerepo=ol7_software_collections"
+
+config_opts['yum.conf'] = """
+[main]
+keepcache=1
+debuglevel=2
+reposdir=/dev/null
+logfile=/var/log/yum.log
+retries=20
+obsoletes=1
+gpgcheck=0
+assumeyes=1
+syslog_ident=mock
+syslog_device=
+metadata_expire=0
+mdpolicy=group:primary
+best=1
+protected_packages=
+user_agent={{ user_agent }}
+
+# repos
+
+[ol7_latest]
+name=Oracle Linux $releasever Latest ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/latest/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=1
+skip_if_unavailable=False
+
+[ol7_optional_latest]
+name=Oracle Linux $releasever Optional Latest ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/optional/latest/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=1
+skip_if_unavailable=False
+
+[ol7_software_collections]
+name=Software Collection Library packages for Oracle Linux 7 ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/SoftwareCollections/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=1
+includepkgs=devtoolset*
+skip_if_unavailable=False
+
+{% for uekver in range(3,7) %}
+[ol7_UEKR{{ uekver }}]
+name=Latest Unbreakable Enterprise Kernel Release {{ uekver }} for Oracle Linux $releasever ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/UEKR{{ uekver }}/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=0
+{% endfor %}
+
+"""
--- a/mock-core-configs/etc/mock/templates/oraclelinux-8.tpl
+++ b/mock-core-configs/etc/mock/templates/oraclelinux-8.tpl
@@ -1,0 +1,67 @@
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['dist'] = 'el8'  # only useful for --resultdir variable subst
+config_opts['releasever'] = '8'
+config_opts['package_manager'] = 'dnf'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+config_opts['bootstrap_image'] = 'docker.io/library/oraclelinux:8'
+
+
+config_opts['dnf.conf'] = """
+[main]
+keepcache=1
+debuglevel=2
+reposdir=/dev/null
+logfile=/var/log/yum.log
+retries=20
+obsoletes=1
+gpgcheck=0
+assumeyes=1
+syslog_ident=mock
+syslog_device=
+metadata_expire=0
+mdpolicy=group:primary
+best=1
+protected_packages=
+module_platform_id=platform:el8
+user_agent={{ user_agent }}
+
+# repos
+
+[ol8_baseos_latest]
+name=Oracle Linux 8 BaseOS Latest ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL8/baseos/latest/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol8
+gpgcheck=1
+enabled=1
+
+[ol8_appstream]
+name=Oracle Linux 8 Application Stream ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL8/appstream/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol8
+gpgcheck=1
+enabled=1
+
+[ol8_codeready_builder]
+name=Oracle Linux 8 CodeReady Builder ($basearch) - Unsupported
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL8/codeready/builder/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol8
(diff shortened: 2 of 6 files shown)


Djelibeybi:
@Conan-Kudo there is no `-latest` equivalent for the UEK repos because each UEK has its own life span and they overlap extensively (UEK3 has only _just_ left production support, for example).  This means folks building kernel modules will need to mock for each UEK release, which would be required anyway.

Also, this is awesome. I have a branch on a fork somewhere for this myself but never got time to test it so I'm very appreciative of your efforts. :)

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
```diff
@@ -0,0 +1,56 @@
+# This list is taken from 'epel-7-x86_64' @buildsys-build group, minus the
+# 'epel-*' specific stuff.
+config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
+
+config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
+config_opts['releasever'] = '7'
+config_opts['bootstrap_image'] = 'docker.io/library/oraclelinux:7'
+config_opts['package_manager'] = 'yum'
+
+config_opts['yum_install_command'] += "{% if target_arch in ['x86_64', 'ppc64le', 'aarch64'] %} --disablerepo=ol7_software_collections{% endif %}"
```
Djelibeybi, line 10: I'm not sure what the purpose of this is? We have an `ol7_software_collections` repo for both `x86_64` and `aarch64` and we don't have any `ppc64le` repos.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
```diff
@@ -0,0 +1,56 @@
...
+skip_if_unavailable=False
+
+[ol7_software_collections]
+name=Software Collection Library packages for Oracle Linux 7 ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/SoftwareCollections/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=1
+includepkgs=devtoolset*
+skip_if_unavailable=False
```
Djelibeybi, line 54: This should probably be `skip_if_unavailable=True`

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
```diff
@@ -0,0 +1,56 @@
...
+enabled=1
+skip_if_unavailable=False
+
+[ol7_software_collections]
+name=Software Collection Library packages for Oracle Linux 7 ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/SoftwareCollections/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=1
+includepkgs=devtoolset*
```
Djelibeybi, line 53: What's the intention of limiting the SCL to just the `devtoolset`?

Djelibeybi (commented):
I would probably include all of the UEK repos (with the UEK6 repo enabled by default) and the `ol${releasever}_developer_EPEL` repos in each template as well.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: This mimics what is done in CentOS and makes it easy to layer for EPEL builds (matching Fedora EPEL policy as devtoolset is the _only_ SCL exposed for EPEL 7 builds).

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: This is for preventing the SCL repo to influence the bootstrap phase (when not using a bootstrap image) when setting up the root for setting up the build environment.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: Well, we want it to fail if packages aren't fetchable, I think?

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Djelibeybi: You can probably just drop `ppc64le` completely and leave `x86_64` and `aarch64` based on the intended purpose.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Djelibeybi: If `devtoolset*` is only for EPEL layering, I'm more inclined to suggest this entire repo be added to a future `oraclelinux7-epel.tpl` but I also can't come up with a scenario in which `ol7_latest` is available and `ol7_software_collection` is not, so 🤷🏻

Conan-Kudo:
> @Conan-Kudo there is no `-latest` equivalent for the UEK repos because each UEK has its own life span and they overlap extensively (UEK3 has only _just_ left production support, for example). This means folks building kernel modules will need to mock for each UEK release, which would be required anyway.
> 

Why not have all the UEK packages available in a single UEK repo? That way it'd be easy for all of them and we don't have to update this thing every time you add a new UEK series?

> Also, this is awesome. I have a branch on a fork somewhere for this myself but never got time to test it so I'm very appreciative of your efforts. :)

I'd had my own private one for doing builds of things based on UEK (for reasons), and I figured that it should be upstreamed.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Djelibeybi: In which case, would it not make sense to remove this repo from the base and include it in a future `oraclelinux7-epel.tpl` instead?

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: We can always change this later if we want to come up with a fancier abstraction. I just wanted to mimic the CentOS one closely for fungibility.

Djelibeybi:
> Why not have all the UEK packages available in a single UEK repo? 

Because we made the decision never to force a UEK upgrade on a customer. Having seperate repos, means customers can stay on UEK4 or UEK5 and continue to receive security, bug and even feature updates without having to upgrade to a new major kernel version. This also helps with ISV/third-party integration who tend to lag on releasing kernel modules for new UEK releases.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Djelibeybi: Fair enough.

Djelibeybi:
> (for reasons)

My personal email is visible on my profile as is my real name which becomes my professional email if you put a dot between the first and last name and add at oracle dot com at the end. I mention this in case you're interested in letting me know what the reasons are in case there's any insight I can provide on how to make your life easier. :)

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: Fixed.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: I think if I do this, I'd want to change `centos-7` at the same time.

Conan-Kudo:
> > Why not have all the UEK packages available in a single UEK repo?
> 
> Because we made the decision never to force a UEK upgrade on a customer. Having seperate repos, means customers can stay on UEK4 or UEK5 and continue to receive security, bug and even feature updates without having to upgrade to a new major kernel version. This also helps with ISV/third-party integration who tend to lag on releasing kernel modules for new UEK releases.

Another way to do this is to do `-uek<version>` as flavors (e.g. `kernel-uek6`), but I guess we're stuck with this now already...

Djelibeybi:
> Another way to do this is to do `-uek<version>` as flavors (e.g. `kernel-uek6`)

True, but that makes the interventionj required to upgrade from one major UEK version to another a little bit more complicated. Right now, the manual intervention is just to enable a later UEK repo and a `yum update` will grab the latest version from that repo.

mock-core-configs/etc/mock/templates/oraclelinux-8.tpl
```diff
@@ -0,0 +1,64 @@
...
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol8
+gpgcheck=1
+enabled=0
+
+[ol8_UEKR6]
+name=Latest Unbreakable Enterprise Kernel Release 6 for Oracle Linux $releasever ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL8/UEKR6/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol8
+gpgcheck=1
+enabled=0
```
Djelibeybi, line 62: The ol8_UEKR6 repo only exists for `x86_64`. On `aarch64`, the UEKR6 is in `ol8_baseos_latest` as we do not ship an RHCK for that architecture.

Conan-Kudo: Fixed.

Conan-Kudo:
@Djelibeybi So I've added all the available UEK repos disabled by default (`mock` needs to be used with `--enablerepo=<repoid>` to use it). For OL7, there's now a template macro to expand out all the UEK versions (which makes it easy to expand...). Since there's only one for OL8, I haven't done that yet...

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
```diff
@@ -0,0 +1,66 @@
...
+skip_if_unavailable=False
+
+{% for uekver in range(3,7) -%}
+[ol7_UEKR{{ uekver }}]
+name=Latest Unbreakable Enterprise Kernel Release {{ uekver }} for Oracle Linux $releasever ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/UEKR6/$basearch/
```
Djelibeybi, line 60: The `baseurl` needs the `uekver` as well.

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
```diff
@@ -0,0 +1,66 @@
...
+[ol7_software_collections]
+name=Software Collection Library packages for Oracle Linux 7 ($basearch)
+baseurl=https://yum.oracle.com/repo/OracleLinux/OL7/SoftwareCollections/$basearch/
+gpgkey=file:///usr/share/distribution-gpg-keys/oraclelinux/RPM-GPG-KEY-oracle-ol7
+gpgcheck=1
+enabled=1
+includepkgs=devtoolset*
+skip_if_unavailable=False
+
+{% for uekver in range(3,7) -%}
```
Djelibeybi, line 57: There is no UEK7... yet. :) It's in alpha, so I'm half-tempted to tell you to leave this, but it'll break if someone tries it now, so perhaps not.

Conan-Kudo: `range()` is non-inclusive: https://jinja.palletsprojects.com/en/2.10.x/templates/#range

mock-core-configs/etc/mock/templates/oraclelinux-7.tpl
Conan-Kudo: Oops, fixed!

praiskup:
Thank you for the PR, we don't have the capacity to maintain these so we expect that it will be community-mainained.

Conan-Kudo:
@praiskup Aren't most of these community maintained as it is? 😉

PR #729, opened 2021-05-20
sync centos-stream-8 with centos-stream-repos

### LINKED ISSUE #708: Use mirror lists in CentOS stream
Use mirrorlists in CentOS Stream config, not baseurl, e.g.
`mirrorlist=http://mirrorlist.centos.org/?release=$stream&arch=$basearch&repo=BaseOS&infra=$infra`
instead of
`baseurl=http://mirror.centos.org/centos/8-stream/BaseOS/$basearch/os/`

The main repo is sometimes overloaded or inconsistent, causing failures like:
https://bugzilla.redhat.com/show_bug.cgi?id=1937953

xsuchy:
Replace baseurl with mirrorlist - resolves #708
sync name and repoid
add several repos which are present in centos-stream-repos - they are disabled by default so no impact

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-8.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-8.tpl
@@ -26,19 +26,28 @@ protected_packages=
 module_platform_id=platform:el8
 user_agent={{ user_agent }}
 
-[Stream-BaseOS]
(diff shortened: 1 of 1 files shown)


praiskup:
Looks ok, thank you

PR #730, opened 2021-05-20
Define _platform_multiplier macro

### LINKED ISSUE #697: Let --forcearch builds define %_mock_forcearch
Context: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/IO2ZFXQ4ZSPT6DKG65OCPMZD35EF5WXS/

I think it would be useful if builds done with  `--forcearch` exposed the information to rpmbuild by defining a macro, e.g. `%_mock_forcearch`  (or any other name). So in that case, maintainers could do things like:

```
# The tests should fail the build if they run for 10+ minutes (or 100+ minutes on emulated arhcitecture)
%make_build tests TIMEOUT=600%{?_mock_forcearch:0}
```

xsuchy:
This allows packagers to count with some delay on emulated architectures. Useful in %check section.

Resolves #697

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -258,6 +258,8 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['macros'] = {
         '%_topdir': '%s/build' % config_opts['chroothome'],
         '%_rpmfilename': '%%{NAME}-%%{VERSION}-%%{RELEASE}.%%{ARCH}.rpm',
+        # This is actually set in check_arch_combination()
+        # '%_platform_multiplier': 1,
     }
     config_opts['hostname'] = None
     config_opts['module_enable'] = []
@@ -310,6 +312,13 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
 
     return config_opts
 
+
+def multiply_platform_multiplier(config_opts):
+    """ Define '%_platform_multiplier' macro based on forcearch.
+        But respect possible overrides in config.
+    """
+    if '%_platform_multiplier' not in config_opts["macros"]:
+        config_opts["macros"]["%_platform_multiplier"] = 10 if config_opts["forcearch"] else 1
 
 @traceLog()
 def set_config_opts_per_cmdline(config_opts, options, args):
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -456,6 +456,11 @@
 
 ## Use this to force foreing architecture (requires qemu-user-static)
 # config_opts['forcearch'] = None
+#
+## Emulating architecture results in slower builds. Expose it to packagers
+## This is set to 1 normally. And to 10 when forcearch is in play.
+# config_opts['macros']['%_platform_multiplier'] = 1
+#
 # If you change chrootgid, you must also change "mock" to the correct group
 # name in this line of the mock PAM config:
 #   auth  sufficient pam_succeed_if.so user ingroup mock use_uid quiet
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -540,6 +540,7 @@ def check_arch_combination(target_arch, config_opts):
                  target_arch, host_arch)
         config_opts['forcearch'] = target_arch
 
+    config.multiply_platform_multiplier(config_opts)
     if config_opts['forcearch']:
         binary = '/usr/bin/qemu-x86_64-static'
         if not os.path.exists(binary):


mock/py/mockbuild/config.py
```diff
@@ -311,6 +312,12 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     return config_opts
 
 
+def multiply_platform_multiplier(config_opts):
+    if config_opts['forcearch']:
+        if (not config_opts['macros']['%_platform_multiplier']) or \
+           (str(config_opts['macros']['%_platform_multiplier']) == '1'):
+            config_opts['macros']['%_platform_multiplier'] = 10
```
praiskup, line 319: Can this be just?
`config_opts['macros']['%_platform_multiplier'] = 10 if config_opts['forcearch'] else 1`

xsuchy: Not if we want to allow admin to override (fine tune) it in config. Consider having `config_opts['macros']['%_platform_multiplier'] = 5` in config.

mock/py/mockbuild/config.py
```diff
@@ -258,6 +258,7 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['macros'] = {
         '%_topdir': '%s/build' % config_opts['chroothome'],
         '%_rpmfilename': '%%{NAME}-%%{VERSION}-%%{RELEASE}.%%{ARCH}.rpm',
+        '%_platform_multiplier': 1,
```
praiskup, line 261: Don't we want to use something with `%mock_` prefix?  I'd personally think this is somet default variable used by rpmbuild.

mock/py/mockbuild/config.py
```diff
@@ -356,6 +363,8 @@ def set_config_opts_per_cmdline(config_opts, options, args):
         config_opts['rpmbuild_arch'] = config_opts['target_arch']
     if options.forcearch:
         config_opts['forcearch'] = options.forcearch
+    # forcearch can be set from other places too
+    multiply_platform_multiplier(config_opts)
```
praiskup, line 367: Can we only call this in `check_arch_combination`?

mock/py/mockbuild/config.py
xsuchy: It can be. I wanted to raise it for rpm.org or Fedora Packaging Guidelines. As I can imagine having this macro defined on system-level and e.g. slow armhpf can have this set to 2 by default.

mock/py/mockbuild/config.py
praiskup: Ok, I'd suggest something like this (I'm reading over again, and it looks
too magic to me):
```python
def multiply_platform_multiplier(config_opts):
    if not config_opts['forcearch']:
        # forcearch disabled, no need to increase the multiplier
        return

    if config_opts['macros']['%_platform_multiplier'] is not None:
        # multiplier set by end-user explicitly
        return

    # increase to 10 for forcearch
    config_opts['macros']['%_platform_multiplier'] = 10
```

And since I believe we can call this only once, I don't think we necessarily have to have this as
a separate method from the `check_arch_combination` (but this is not a strong opinion).

praiskup: No, wrong.  This is better:
```python
if config_opts['macros']['%_platform_multiplier'] is not None:
    # multiplier set by end-user explicitly
    return
config_opts['macros']['%_platform_multiplier'] = 10 if config_opts['forcearch'] else 1
```

It is OK because we don't have to call this multiple times.

mock/py/mockbuild/config.py
xsuchy: I can, but then the multiplier and forcearch option will be "out of sync" for most of `set_config_opts_per_cmdline()` and for several lines in `mock.py`.

mock/py/mockbuild/config.py
xsuchy: The default is 1 (not None).

praiskup: The logic is is really complicated;  e.g. if user decides to set the value explicitly to `1`, we reset it to `10`.

xsuchy: > I don't think we necessarily have to have this as a separate method from the check_arch_combination

check_arch_combination is already quite long. I want to have it separate.

xsuchy: I updated the PR and added comments to this section. Is it better or you still want to rephrase it?

mock/py/mockbuild/config.py
praiskup: Hmm, ok ... I won't object too much honestly :-) it just looks too complicated to me, while we could just live with something like
https://github.com/praiskup/mock/commit/4f638bedadef476dd2bc66933cb6274ae4e2e439

xsuchy:
> while we could just live with something like praiskup@4f638be

This is too complicated for me :))

xsuchy:
I wanted to change the default to 0 or '' which would simplify the code, but... the question is how we suggest using it?

```
%global timeout 60*%{?_platform_multiplier:%{_platform_multiplier}}%{!_platform_multiplier:1}
```
This sounds too complicated to me. :( And it will not even work if we make the default '' or 0 :(

/me scratching on the head. :(

praiskup:
Hmpf, not an ideal state, unless redhat-rpm-macros defines the default: `rpm --eval 'sleep 60*%{?_platform_multiplier}%{!?_platform_multiplier:1}'`

praiskup:
Alternatively, we can document the use as:
```
%{!?_platform_multiplier:%global _platform_multiplier 1}
timeout $(( 60*%_platform_multiplier )) the-long-running-task
```

praiskup:
+1

Issue #693, opened 2021-01-29
[RFE] `mock shell` should change directory to `%{_builddir}`


voxik:
Currently, whenever I do `mock shell --unpriv`, I endup in root directory and this is typically followed byt `cd /builddir/build/BUILD/builddir_of_the_package`. Would you consider to do this step automatically? I don't think that being in the root directory is good default.

praiskup:
I'm not sure why login shell isn't used, I have to check later.  IMO we should switch to $HOME automatically, but this doesn't happen.

I don't think it is wise to go anywhere else than `/builddir` by default (non-standard behavior, something contrary to what I personally expect - so that would be about preferences).  The `builddir_of_the_package` is variable, hard to guess correctly ... I'd say using shell completion could be enough?  Perhaps we could have some option for initial `cd path` command?  I don't think it is worth it...?

hroncok:
Maybe have a different shell command or option(s) to do this:

 - su mockbuild
 - cd $HOME

    $ mock -r fedora-rawhide-x86_64 --shell --user

Or

    $ mock -r fedora-rawhide-x86_64 --usershell

voxik:
> IMO we should switch to $HOME automatically

Sorry for confusion, this actually happens. But that is not useful either.

> I don't think it is wise to go anywhere else than `/builddir` by default (non-standard behavior, something contrary to what I personally expect - so that would be about preferences).

You are right, that it would be non-standard, but probably more convenient. There is basically nothing useful what can one do in `$HOME` in mock.

> The builddir_of_the_package is variable, hard to guess correctly

The `%{_builddir}` is `$HOME/rpmbuild/BUILD` by default, but if there was chance to even step into the expanded sources directory, that could be nice as well.

> Perhaps we could have some option for initial `cd path` command?

That starts to be more complicated then the explicit `cd`. But if the default behavior could be configured, that could be helpful.

jwakely:
What would be useful is if `mock --rebuild` (and `fedpkg mockbuild`) printed a message when they fail, telling you the directory to change to.

```
Build failed, see /build/rpmbuild/BUILD/foo-1.2.3 in the chroot filesystem.
Alternatively, /var/lib/mock/$chroot/build/rpmbuild/BUILD/foo-1.2.3 in the local filesystem.
```

Then at least you can just select that with the mouse and paste it into the shell (either mock shell or the host shell).

xsuchy:
[...]

PR #732, opened 2021-05-20
honor --cwd for --shell

xsuchy:
Additionally change to `chroothome` when --shell.

Relates: #693


praiskup:
Thanks, this looks like it could work but I will run the tests first.

praiskup:
Ok, looks like no breakage happens, thanks!

PR #733, opened 2021-05-23
allow mock --install mock --install external:*

xsuchy:
one commit which moves code to separate function and then second commit which calls it from `install()`


praiskup:
Thank you!

PR #736, opened 2021-06-02
compress_logs: compress also after repo failures

praiskup:
Before, errors like
    Errors during downloading metadata for repository 'local'
Ended up with non-compressed logs in resultdir.

We can not easily move the 'postbuild' hook call as others already
depend on that - so I'm proposing a new hook for log processing.

Original report:
https://pagure.io/copr/copr/issue/1648


praiskup:
Needs fix, this doesn't compress logs when there's no exception.

praiskup:
Meh, not awesome, not 100% complete - but simple change enough.  PTAL

Issue #731, opened 2021-05-20
Leap 15.3 repo failures?


praiskup:
```
...
warning: /var/lib/mock/opensuse-leap-15.3-x86_64-bootstrap/root/var/cache/dnf/opensuse-leap-oss-6b694521cb63b0e5/packages/bash-4.4-17.83.x86_64.rpm: Header V3 RSA/SHA256 Signature, key ID 39db7c82: NOKEY
openSUSE Leap 15.3 - x86_64 - OSS                                                                                                                                                                           965 kB/s | 988  B     00:00    
GPG key at file:///usr/share/distribution-gpg-keys/opensuse/RPM-GPG-KEY-openSUSE (0x3DBDC284) is already installed
warning: /var/lib/mock/opensuse-leap-15.3-x86_64-bootstrap/root/var/cache/dnf/opensuse-leap-oss-6b694521cb63b0e5/packages/dnf-4.5.2-bp153.1.14.noarch.rpm: Header V3 RSA/SHA256 Signature, key ID 65176565: NOKEY
The GPG keys listed for the "openSUSE Leap 15.3 - x86_64 - OSS" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: bash-4.4-17.83.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/opensuse/RPM-GPG-KEY-openSUSE
Public key for coreutils-8.32-1.2.x86_64.rpm is not installed. Failing package is: coreutils-8.32-1.2.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/opensuse/RPM-GPG-KEY-openSUSE
Public key for cracklib-2.9.7-11.3.1.x86_64.rpm is not installed. Failing package is: cracklib-2.9.7-11.3.1.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/opensuse/RPM-GPG-KEY-openSUSE
Public key for cracklib-dict-full-2.8.12-1.22.x86_64.rpm is not installed. Failing package is: cracklib-dict-full-2.8.12-1.22.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/opensuse/RPM-GPG-KEY-openSUSE
Public key for dbus-1-1.12.2-8.3.1.x86_64.rpm is not installed. Failing package is: dbus-1-1.12.2-8.3.1.x86_64
[...]

PR #737, opened 2021-06-05
config: Add GPG keys and RPM repositories for openSUSE Leap 15.3

Conan-Kudo:
openSUSE Leap 15.3 has multiple GPG keys and requires multiple
repositories to construct its updates.

This requires a version of distribution-gpg-keys that includes
the extra GPG keys.

Reference: https://bugzilla.opensuse.org/1186593

Requires: https://github.com/xsuchy/distribution-gpg-keys/pull/45
Requires: https://github.com/xsuchy/distribution-gpg-keys/pull/46

Fixes: #731

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -18,7 +18,7 @@ BuildArch:  noarch
 Provides: mock-configs
 
 # distribution-gpg-keys contains GPG keys used by mock configs
-Requires:   distribution-gpg-keys >= 1.48
+Requires:   distribution-gpg-keys >= 1.54
 # specify minimal compatible version of mock
 Requires:   mock >= 2.5
 Requires:   mock-filesystem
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
cc: @lkocman @carlwgeorge

mock-core-configs/mock-core-configs.spec
```diff
@@ -18,7 +18,7 @@ BuildArch:  noarch
 Provides: mock-configs
 
 # distribution-gpg-keys contains GPG keys used by mock configs
-Requires:   distribution-gpg-keys >= 1.48
```
praiskup, line 21: I don't see this update, yet.  It means that the testing package wouldn't be installable.  Otherwise LGTM

Conan-Kudo: Yeah, I raised it to the next version because there are two PRs pending for it:

* https://github.com/xsuchy/distribution-gpg-keys/pull/45
* https://github.com/xsuchy/distribution-gpg-keys/pull/46

PR #741, opened 2021-06-10
Add epel-next-8 configs

carlwgeorge:
EPEL Next is a new repository that allows packagers to build against CentOS Stream instead of RHEL.  It is not an entire rebuild of EPEL and is intended to be used in conjunction with regular EPEL.  See the [wiki page](https://fedoraproject.org/wiki/EPEL_Next) for more details.


Conan-Kudo (approved):

praiskup:
thank you

PR #747, opened 2021-07-02
Disable versionlock DNF plugin by default

ignatenkobrain:
The system-wide versionlock from DNF does not make any sense to use for
mock builds.

Signed-off-by: Igor Raits <igor.raits@gmail.com>


Conan-Kudo (approved):

xsuchy:
True, but it does not make any harm either. Right?

ignatenkobrain:
It actually does if you happen to versionlock packages that you need in buildroot :)

ignatenkobrain:
(And versions do not match)

praiskup:
thank you

Issue #719, opened 2021-04-12
add CentOS Stream 9


TomasTomecek:
This is a request to add CentOS Stream 9 as a chroot to mock. Sadly, it seems this can't be fulfilled yet since the composes are not public. I just wanted to open this issue so people can subscribe and we can post more info here.

sandrobonazzola:
Composes are public: https://composes.stream.centos.org/test/latest-CentOS-Stream/

Conan-Kudo:
They are not GPG signed yet.

TomasTomecek:
Is it a requirement to have them signed in order to include a config for Stream 9 in mock? I understand this could be improved in future once they become signed. (just checked and they are still not signed)

xsuchy:
gpgcheck can be turned on later. The question is: is the URL final, will it be able to get the traffic?

Conan-Kudo:
I believe that @bstinson is going to have it producing signed content from a stable (e.g. mirrorable) URL soon.

Conan-Kudo:
I believe that @bstinsonmhk is going to have it producing signed content from a stable (e.g. mirrorable) URL soon.

TomasTomecek:
Looks fixed to me :)

https://github.com/rpm-software-management/mock/releases/tag/mock-core-configs-34.5-1

Thanks, Pavel!

PR #751, opened 2021-07-08
config: add CentOS Stream 9 "preview" configs

praiskup:
The composes are not signed, and at the time of writing this patch
mirrors do not exist.  Neither centos:stream9 bootstrap image exists.

Relates: #719

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-9.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-9.tpl
@@ -1,0 +1,49 @@
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['dist'] = 'el9'  # only useful for --resultdir variable subst
+config_opts['releasever'] = '9'
+config_opts['package_manager'] = 'dnf'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+
+# TODO: flip to 'stream9' tag once available
+config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream9-development'
+
+config_opts['dnf.conf'] = """
+[main]
+keepcache=1
+debuglevel=2
+reposdir=/dev/null
+logfile=/var/log/yum.log
+retries=20
+obsoletes=1
+gpgcheck=1
+assumeyes=1
+syslog_ident=mock
+syslog_device=
+mdpolicy=group:primary
+best=1
+protected_packages=
+module_platform_id=platform:el9
+user_agent={{ user_agent }}
+
+[baseos-pre-release]
+name=CentOS Stream $releasever - BaseOS (pre-release)
+baseurl=https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/BaseOS/$basearch/os/
+failovermethod=priority
+#gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+gpgcheck=0
+skip_if_unavailable=False
+
+[appstream-pre-release]
+name=CentOS Stream $releasever - AppStream (pre-release)
+baseurl=https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/AppStream/$basearch/os/
+enabled=1
+#gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+gpgcheck=0
+
+[crb-pre-release]
+name=CentOS Stream $releasever - CRB (pre-release)
+baseurl=https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/CRB/$basearch/os/
+enabled=1
+#gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+gpgcheck=0
+"""
--- a/mock-core-configs/etc/mock/centos-stream-9-aarch64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-9-aarch64.cfg
@@ -1,0 +1,5 @@
+include('templates/centos-stream-9.tpl')
+
+config_opts['root'] = 'centos-stream-9-aarch64'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/centos-stream-9-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-9-ppc64le.cfg
@@ -1,0 +1,5 @@
+include('templates/centos-stream-9.tpl')
+
+config_opts['root'] = 'centos-stream-9-ppc64le'
+config_opts['target_arch'] = 'ppc64le'
+config_opts['legal_host_arches'] = ('ppc64le',)
--- a/mock-core-configs/etc/mock/centos-stream-9-s390x.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-9-s390x.cfg
@@ -1,0 +1,5 @@
+include('templates/centos-stream-9.tpl')
+
+config_opts['root'] = 'centos-stream-9-s390x'
+config_opts['target_arch'] = 's390x'
+config_opts['legal_host_arches'] = ('s390x',)
--- a/mock-core-configs/etc/mock/centos-stream-9-x86_64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream-9-x86_64.cfg
@@ -1,0 +1,5 @@
+include('templates/centos-stream-9.tpl')
+
+config_opts['root'] = 'centos-stream-9-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)


praiskup:
Do we really have to block these "preview" configs?

Conan-Kudo:
Yes, @bstinsonmhk requested we not add configs until packages and repository metadata are signed. That's why I haven't added them yet either.

praiskup:
Do you have ETA?  Or are we afraid that the baseurl will not handle the traffic?

Conan-Kudo:
I was told we'll have the final URLs with signed packages and repos next month, but I'll ask again in the Stream office hours next week.

praiskup:
OK, FTR, same as ELN (not yet mirrored) - having this patch is better than nothing IMO.  And I'd like to wrap new mock-core-configs soon (I was expecting this week).

Conan-Kudo:
ELN is different in that all the content is signed, and they have made progress on making it mirrorable.

praiskup:
Well, I still think that we can turn gpgcheck=1 later, and use mirrors once they are available.  People could at least do the first experiments (per #719 discussion desired).

Conan-Kudo:
If we can wait a week, I can talk to @bstinsonmhk and @carlwgeorge about this and we can figure out what to do next then.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['dist'] = 'el9'  # only useful for --resultdir variable subst
+config_opts['releasever'] = '9'
+config_opts['package_manager'] = 'dnf'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream9'
```
carlwgeorge, line 6: For now this can be set to `quay.io/centos/centos:stream9-development`.  Later on the `stream9` tag will be correct, but if you want this to work in the near term you'll need to use a tag that exists.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
...
+
+config_opts['dnf_vars'] = { 'stream': '9-stream',
+                            'contentdir': 'centos',
+                          }
```
carlwgeorge, line 10: These variables are specific to the way 8-stream is mirrored.  We haven't yet settled on a mirror strategy for 9-stream.  I would recommend removing them until we know that they are necessary.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
...
+retries=20
+obsoletes=1
+gpgcheck=0
+assumeyes=1
+syslog_ident=mock
+syslog_device=
+mdpolicy=group:primary
+best=1
+protected_packages=
+module_platform_id=platform:el8
```
carlwgeorge, line 27: This should be `platform:el9`.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
...
+user_agent={{ user_agent }}
+
+[baseos]
+name=CentOS Stream $releasever - BaseOS
+baseurl=https://composes.stream.centos.org/test/latest-CentOS-Stream/compose/BaseOS/$basearch/os/
```
carlwgeorge, line 32: **If** we move forward with this before the content is mirrored, we shouldn't use the `test` composes, we should use the `production` composes, e.g. `https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/BaseOS/$basearch/os/`.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
...
+failovermethod=priority
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+skip_if_unavailable=False
+
+[appstream]
+name=CentOS Stream $releasever - AppStream
+baseurl=https://composes.stream.centos.org/test/latest-CentOS-Stream/compose/AppStream/$basearch/os/
```
carlwgeorge, line 39: `https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/AppStream/$basearch/os/`

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
...
+enabled=1
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+
+[cr]
```
carlwgeorge, line 43: This repoid should be `crb`, not `cr`.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,48 @@
...
+name=CentOS Stream $releasever - CRB
+baseurl=https://composes.stream.centos.org/test/latest-CentOS-Stream/compose/CRB/$basearch/os/
```
carlwgeorge, line 45: `https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/CRB/$basearch/os/`

carlwgeorge (changes requested):

praiskup:
Thank you for the review!

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
praiskup: Good point, thanks.  @carlwgeorge   Can you guess the time when mirrors will be working?  If that is more than a week or two, we'd rather release these now, and release again with the mirrorlist variant once  available.

carlwgeorge: I don't have an exact guess, but it is safe to say it will take longer than two weeks.

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -0,0 +1,45 @@
...
+best=1
+protected_packages=
+module_platform_id=platform:el9
+user_agent={{ user_agent }}
+
+[baseos]
+name=CentOS Stream $releasever - BaseOS
+baseurl=https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/BaseOS/$basearch/os/
+failovermethod=priority
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
```
carlwgeorge, line 31: Another thing I just now thought of, we should remove the `gpgkey` lines and add an explicit `gpg_check=0` line for this initial config.

carlwgeorge (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
praiskup: There's gpgcheck=0 set globally, which will need a fix later on once the content is signed.
Do you expect there will be a different key than the CentOS Official?

carlwgeorge: Nope, it should be the same key.  I'm just concerned if someone changes that global setting that this will error out without the explicit `gpgcheck=0`.  It's even more misleading if the gpgkey is already defined.

bstinsonmhk:
I'm going to walk back a little bit on my blocker here. I think it would be OK to release these pointing at the composes as long as we signal (maybe in the repoids?) that the content is pre-release/unsigned.

ktdreyer:
I was just looking for this today, so that my team and I can build Ceph's dependencies in https://copr.fedorainfracloud.org/coprs/ceph/el9/

We're fine with unsigned composes for now. @praiskup is it possible for us to tese this config within Copr before it's merged here?

praiskup:
You can clone the praiskup/mock fork, and in the correct branch hit `tito build --test --rpm ( --install )`.  (sorry, the CI pull-request builds somewhat stopped working, I am going to fix that soon).

ktdreyer:
Oh, I should have clarified - could you patch https://copr.fedorainfracloud.org/ to include this so we can test it out early?

praiskup:
@carlwgeorge wrote
>  I'm just concerned if someone changes that global setting that this will error out without the explicit gpgcheck=0. It's even more misleading if the gpgkey is already defined.

Ok, I specified gpgcheck=0 everywhere, and commented-out gpgkey statements.  Good point.

@bstinsonmhk wrote:
> I think it would be OK to release these pointing at the composes as long as we signal (maybe in the repoids?) that the content is pre-release/unsigned.

The repoids now look like `baseos-pre-release`, etc.  The name= tags are changed too, so dnf writes:
```
CentOS Stream 9 - BaseOS (pre-release)                                                                                                                                                                                                                          965 kB/s | 916 kB     00:00    
CentOS Stream 9 - AppStream (pre-release)                                                                                                                                                                                                                       5.0 MB/s | 5.3 MB     00:01    
CentOS Stream 9 - CRB (pre-release)   
```

@ktdreyer wrote:
> could you patch https://copr.fedorainfracloud.org/ to include this so we can test it out early?

Yes.

praiskup:
> could you patch https://copr.fedorainfracloud.org/ to include this so we can test it out early?

[x86_64 and aarch64 enabled](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/UQA5XDTSCNHYKDMQVGOUTCOGB6REKFDH/)

ktdreyer:
For reference, the CentOS Stream 9 config files are available in mock-core-configs-34.5.

PR #754, opened 2021-07-16
mock: don't set --cwd for --shell when we know it fails

praiskup:
This is specific to EL7 environment, where the systemd-nspawn
implementation still doesn't support --chdir option.

Resolves: rhbz#1976702

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -357,7 +357,12 @@ class Commands(object):
             gid = 0
         cwd = options.cwd
         if not cwd:
-            cwd = self.config['chroothome']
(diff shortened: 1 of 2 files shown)


xsuchy:
+1
But on the other hand I propose to stop building Mock for EL7. There is not feature planned ahead and current version will continue working there. And we can only build new mock-core-configs for EL7. This will allows us to do some cleaning in master. Including this commit.

praiskup:
I'm OK to discuss this, though I don't see benefits in removing EL7 atm :-( cleaning **that** code is as
trivial as keeping it.  The real problems of mock are the isolation abstraction, logging and stdout/err handling :-(

Issue #715, opened 2021-04-02
Doesn't work inside a CentOS 7 toolbox


cheese:
### Short description of the problem
Mock failed at bind mount when run inside a CentOS 7 toolbox. Maybe the version of util-linux is too low?

Output of `mock -r epel-7-x86_64 shell`

```
INFO: mock.py version 2.9 starting (python version = 3.6.8, NVR = mock-2.9-1.el7)...
Start(bootstrap): init plugins
INFO: selinux disabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux disabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start(bootstrap): chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: enabled package manager cache
Start(bootstrap): cleaning package manager metadata
Finish(bootstrap): cleaning package manager metadata
INFO: enabled HW Info plugin
ERROR: Command failed: 
 # /bin/mount -n --rbind -o nodev,noexec,nosuid,readonly,rprivate /proc /var/lib/mock/epel-7-x86_64-bootstrap/root/proc
```

Output of `sudo /bin/mount -n --rbind -o nodev,noexec,nosuid,readonly,rprivate /proc /var/lib/mock/epel-7-x86_64-bootstrap/root/proc`

```
mount: wrong fs type, bad option, bad superblock on /proc,
       missing codepage or helper program, or other error

       In some cases useful info is found in syslog - try
       dmesg | tail or so.
```

### Output of `rpm -q mock`
mock-2.9-1.el7.noarch

### Steps to reproduce issue

1. build a Toolbox-compatible image from `images/centos/7` of https://github.com/kronenpj/toolbox.git branch centos-7 as `localhost/centos-toolbox:7`
2. Create a toolbox: toolbox create --image localhost/centos-toolbox:7 centos-toolbox-7
3. toolbox enter centos-toolbox-7
4. sudo yum install -y epel-release
5. sudo yum install -y mock
6. mock -r epel-7-x86_64 shell

### Any additional notes

- mock works inside a Fedora 34 toolbox
- mock works inside a root podman privileged container

<details>
  <summary>
    <p>Output of `mock --debug-config`</p>
  </summary>
config_opts['bootstrap_image'] = 'registry.centos.org/centos:7'
config_opts['build_log_fmt_str'] = '%(message)s'
config_opts['cache_alterations'] = False
config_opts['chroot_name'] = 'default'
[...]

PR #759, opened 2021-07-24
mock: don't pass source to `mount -o remount`

cheese:
The mount program with `-o remount` behave differently when source is not passed
(rhbz#1985628). Though in a normal host the result is the same whether passing
source argument, in some special environment, for example in some Toolbox
container, passing source to the mount program results in a permission error. So
we just don't pass the source argumnent and the mount command will work in hosts
and containers

Closes: #715

### DIFF
--- a/mock/py/mockbuild/mounts.py
+++ b/mock/py/mockbuild/mounts.py
@@ -112,7 +112,7 @@ class BindMountPoint(MountPoint):
         # Userspace must implement this as separate system calls anyway.
         if self.options:
             options = ','.join(['remount', self.options, bind_option])
-            util.do(['/bin/mount', '-n', '-o', options, self.srcpath,
+            util.do(['/bin/mount', '-n', '-o', options, "--target",
                      self.bindpath])
         return True
 


praiskup:
Thank you for the PR!

Can you please elaborate on "why this helps" in the commit message? Namely, I'm interested to see why you got permission denied before, and not after applying this.

The additional thing mount should do now is that it reads fstab/mtab which itself could result in a permission error (and nobody takes care of the /etc/fstab in mock).  From this perspective I'd say that we really want to have the *source* specified.

cheese:
> Thank you for the PR!
> 
> Can you please elaborate on "why this helps" in the commit message? Namely, I'm interested to see why you got permission denied before, and not after applying this.

Added more verbose comment in the commit, showing the syscall level difference.

> 
> The additional thing mount should do now is that it reads fstab/mtab which itself could result in a permission error (and nobody takes care of the /etc/fstab in mock). From this perspective I'd say that we really want to have the _source_ specified.

`mount` actually reads current options from /proc/mounts and /proc/self/mountinfo, so reading won't be a problem. And `/etc/mtab` is a usually symlink to `/proc/self/mounts` in current systems.

praiskup:
Thank you for the update!

>> mount("/usr/local", "/var/lib/m", 0x56381a4532d0, MS_RDONLY|MS_REMOUNT|MS_BIND, NULL)
>
> This will fail with permission error in some Toolbox containers, since there are
> many options that cannot be reset inside the container.

What options do you have in mind?

> But if source is not passed, the final arguments passed to mount(2) are combined
> with the current mount options.
>
>>  mount -n -o remount,ro,bind --target /var/lib/m
>
> results in
>
>> mount("fuse-overlayfs", "/var/lib/m", 0x55d9a97f3310,
>>  MS_RDONLY|MS_NODEV|MS_REMOUNT|MS_NOATIME|MS_BIND  "user_id=0,group_id=0,default_per"...)

I am still missing how this could start suddenly working, can you help me to
understand that?

The mount calls seem to be identical (even though you haven't copy-pasted the
mount(2) call return values from strace).  The second (working) call actually
contains much more options than the first one.

Note that I still fail to reproduce this, so it is rather weird to me.  It seems like some
glitch in your toolbox configuration which blocks you to do some operation (which I can
personally do on my box).

I'm not entirely against this patch, but I'd like to fully understand what we are changing
and why.

cheese:
> The mount calls seem to be identical (even though you haven't copy-pasted the
> mount(2) call return values from strace). The second (working) call actually
> contains much more options than the first one.

Yes, that's the tricky point. The mount commands are near identical but the resulting syscalls are different.
The simpler mount syscall generated by the mount command with target specified will fail with permission error.

> 
> Note that I still fail to reproduce this, so it is rather weird to me. It seems like some
> glitch in your toolbox configuration which blocks you to do some operation (which I can
> personally do on my box).

I reproduced this issue in a clean VM:

1. Install a VM from Fedora-Silverblue-ostree-x86_64-34-1.2.iso
2. On the host run: toolbox enter
3. In toolbox run: sudo yum install mock -y
4. In toolbox run: mock --shell

praiskup:
Thank you for the update, and patience :-)

> The simpler mount syscall generated by the mount command with target specified will fail with permission error.

Isn't this actually a bug in a specific environment?  Do we have a reproducer without mock?
Could that be reported against toolbox/kernel/util-linux?
The problem is that I don't understand *why* this is happening, and why it doesn't happen here.

praiskup:
The answer could be "SELinux on the host is preventing to touch XYZ, which leads to EPERM errno".
I'm afraid that this patch could just fix the problem by luck, so it could possibly return back to us later.

cheese:
I ask[1] util-linux maintainer in Fedora ealier to change the behavior of mount command. The issue is reproducible without mock. He rejected and suggested me to use the `--options-source` and `--options-mode` options to fix this issue. But those options are not available in the mock command of EL7. I want to keep EL7 compitibility so I prefer the solution in this PR.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1985628

praiskup:
Ok, the reason why this fails in Toolbox is that `mock -r remount,bind,ro SRC DEST` doesn't specify the `nodev` option.
The `mock -r remount,bind,ro DEST` is more likely to succeed because the `nodev` is already on the old bind-mount (intherited from the main mountpoint).  Thank you for the report!

Issue #762, opened 2021-08-13
mock-core-configs for Fedora 35 branched ? 


sergiomb2:
### Short description of the problem

I think we miss mock-core-configs for Fedora 35 branched 

thank you

praiskup:
Merged fix here: 036ae2f08fe493654064cd57b0032cfab5337731

PR #763, opened 2021-08-14
config: add Fedora 35 configs

eclipseo:
Close: #762

### DIFF
--- a/.tito/packages/mock-core-configs
+++ b/.tito/packages/mock-core-configs
@@ -1,1 +1,1 @@
-34.6-1 mock-core-configs/
+35-1 mock-core-configs/
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -1,5 +1,5 @@
 Name:       mock-core-configs
-Version:    34.6
+Version:    35
 Release:    1%{?dist}
 Summary:    Mock core config files basic chroots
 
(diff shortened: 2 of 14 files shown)


sergiomb2:
Hi, as in https://github.com/rpm-software-management/mock/commit/3480b06be2ab86133ade765619ffd3b21d7e7d60 we and bump rawhide release to 36 in mock-core-configs/etc/mock/templates/fedora-rawhide.tpl

hroncok:
The CI failed only because it tries to build the package in Copr and Rawhide Copr chroots are now broken by the thing that this fixes.

hroncok:
This fixes the 35 roots for me, but rawhide/36 still fails.

sergiomb2:
for a quick fix, we need update
/etc/mock/templates/fedora-rawhide.tpl line 12 with
config_opts['releasever'] = '36', as reference https://github.com/rpm-software-management/mock/commit/3480b06be2ab86133ade765619ffd3b21d7e7d60#diff-6d85f049d29a9ebeda1e02d48fb369982d6410ba63c64306001e1f41f8a3c4dd

Conan-Kudo:
Please drop the tag-release commit in here, merging that won't bring the tag over into the main repository, and it blocks @praiskup from releasing mock-core-configs-35.

.tito/packages/mock-core-configs
```diff
@@ -1 +1 @@
-34.6-1 mock-core-configs/
+35-1 mock-core-configs/
```
xsuchy, line 1: This should not be part of this change. This is done automatically when you do `tito tag` which is part of release process.

mock-core-configs/mock-core-configs.spec
```diff
@@ -149,6 +149,9 @@ fi
 %ghost %config(noreplace,missingok) %{_sysconfdir}/mock/default.cfg
 
 %changelog
+* Sat Aug 14 2021 Robert-André Mauchin <zebob.m@gmail.com> 35-1
+- Add Fedora 35 configs
+
```
xsuchy, line 154: This as well. The changes in the spec file are done during release process.

praiskup:
Thank you for the PR, I'm taking the first commit (plus 35=>36 change in rawhide template).

eclipseo:
Sorry I was working, didn't have time to follow up as quickly as expected.

PR #766, opened 2021-08-17
Add eurolinux-8 x86_64 and i686 buildroots

AlexBaranowski:
Hi,

I want to add EuroLinux 8 mocks. EuroLinux build roots are entire repos with all build artefacts (including lots of normally removed `-devel` packages).

I tested the following changes by building mock-core-configs with tito (--test is required) and running two builds of htop. One for x86_64 and a second one for i686.


praiskup:
Thank you for the patch.  Has EuroLinux mirrors?

AlexBaranowski:
Hi,

Public mirrors are on our roadmap. Right now we use Bunny CDN.
When public mirrors will be available I will update mocks. I already scheduled update for this repo in our planner because there will be new repos (HA, RS) and new arch (aarch64).

PR #770, opened 2021-09-02
config: Validate GPG signatures for CentOS Stream 9

Conan-Kudo:
CentOS Stream 9 packages are now signed on each compose. However,
we're still in a pre-release state where CentOS Stream 9 is not yet
on the mirror network.


praiskup:
Thank you!

PR #773, opened 2021-09-15
Productionize CentOS Stream 9 build configuration

Conan-Kudo:
This pull request makes the necessary adjustments to productionize the CentOS Stream 9 configuration. CentOS Stream 9 is now on the mirror network, so we're using the new metalinks from MirrorManager. Additionally, I've fixed an issue where weak dependencies were being installed in the build environments and synchronized the chroot setup package set with the one in Koji.

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-9.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-9.tpl
@@ -1,4 +1,4 @@
-config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['chroot_setup_cmd'] = 'install tar redhat-rpm-config redhat-release which xz sed make bzip2 gzip coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep glibc-minimal-langpack'
 config_opts['dist'] = 'el9'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '9'
 config_opts['package_manager'] = 'dnf'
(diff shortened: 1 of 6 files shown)


mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -1,4 +1,4 @@
-config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['chroot_setup_cmd'] = 'install tar redhat-rpm-config redhat-release which xz zstd sed make bzip2 gzip coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
```
xsuchy, line 1: Why zstd? I do not see it in 9's group `build`.

xsuchy: zstd is the dependency of rpm-build and will be transitively loaded.
And I wonder why we do not have glibc-minimal-langpack here?

Conan-Kudo: We should, that's my mistake... Fixed!

mock-core-configs/etc/mock/templates/centos-stream-9.tpl
```diff
@@ -1,4 +1,4 @@
-config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ redhat-rpm-config redhat-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['chroot_setup_cmd'] = 'install tar redhat-rpm-config redhat-release which xz zstd sed make bzip2 gzip coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep glibc-minimal-langpack'
```
xsuchy, line 1: Still.... why zstd? It is only deps of rpm-build.

Conan-Kudo: Dropped.

xsuchy:
Thank you. I will do a release tomorrow.

PR #776, opened 2021-09-30
A new option --rebuild --build-requires

praiskup:
For debugging purposes, it might be useful to complement the
BuildRequires specified in the package.  Turns out that the similar
option chroot_additional_packages is inconvenient for this purpose.

Fixes: #775


xsuchy:
I would propose:
```
   --additional_packages
```
So people do not get idea, that this is standard way to specify build requires :)

And store the value in:
``
config_opts['command_line_addition_packages']
```

File: mock/py/mockbuild/buildroot.py [hunk 1]
```diff
@@ -103,7 +103,13 @@ def __init__(self, config, uid_manager, state, plugins, bootstrap_buildroot=None
         self.build_log = getLog("mockbuild.Root.build")
         self.logging_initialized = False
         self.chroot_was_initialized = False
-        self.preexisting_deps = []
+
+        build_requires = config["build_requires"] or []
+        if is_bootstrap:
+            self.preexisting_deps = []
+        else:
+            self.preexisting_deps = build_requires
```
xsuchy, line 111: Hmm, why?
If you wipe away this whole if-section and set config_opts['bootstrap_command_line_addition_packages'] in config.py to empty list then you get the freedom to admin to override this value in config. Which is impossible with this hardcoded style.

praiskup: I can't imagine a reason to put this into configuration files... therefore I intentionally made this so it can not be configured at all.

In all situations where I got one would use 'chroot_additional_packages' instead.

praiskup:
> I would propose:
>   --additional_packages

No objections... Ok.

> config_opts['command_line_addition_packages']

I didn't want to introduce another option to configs when there's no motivation (but I may be wrong here, correct me if so).

praiskup:
Updated, PTAL

PR #779, opened 2021-10-05
Update Oraclelinux 7/8 configs and add Oraclelinux EPEL 7/8 configs

daarchib:
There are currently no Oracle Linux EPEL configs.
Add oraclelinux-release to existing oraclelinux templates

### DIFF
--- a/mock-core-configs/etc/mock/oraclelinux-epel-7-aarch64.cfg
+++ b/mock-core-configs/etc/mock/oraclelinux-epel-7-aarch64.cfg
@@ -1,0 +1,6 @@
+include('templates/oraclelinux-7.tpl')
+include('templates/oraclelinux-epel-7.tpl')
+
+config_opts['root'] = 'oraclelinux-7-aarch64'
+config_opts['target_arch'] = 'aarch64'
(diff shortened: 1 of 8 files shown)


Conan-Kudo:
@daarchib Why doesn't this use the real EPEL?

daarchib:
> @daarchib Why doesn't this use the real EPEL?

When you say the real EPEL, are you referring to the existing EPEL templates ? (epel7.tpl and epel8.tpl)

The existing EPEL configs are using centos and fedora mirrors. These specific Oracle Linux templates point to the official Oracle yum repositories and are meant to be used to build respective OL7 and OL8  distro packages, including OL EPEL packages.

praiskup:
@Conan-Kudo as the original contributor of those configs, can you please continue with the review here?

Conan-Kudo:
@daarchib Please rename the configs so that it's clear that this is Oracle EPEL and not the EPEL people expect. That is, change it to `oraclelinux-epel` instead of `epel-oraclelinux`. They aren't the same and people might accidentally assume they are.

daarchib:
@Conan-Kudo I have rebased and changed the names of the files as requested.

Issue #790, opened 2021-11-02
epel-next-9 broken, missing epel-release package


praiskup:
Chroot installation fails with:
```
...
Extra Packages for Enterprise Linux 9 - Next - x86_64                                                                                                                                                       
148 kB/s |  36 kB     00:00    
No match for argument: epel-release
No match for argument: epel-rpm-macros
No match for argument: fedpkg-minimal
No match for argument: epel-next-release
```

@carlwgeorge seems like we should override `chroot_setup_cmd` in epel-next-9.tpl, not just
append `epel-next-release` to it?

carlwgeorge:
No, the initial packages will have the gpg key in epel-release and the epel-next repo config in epel-next-release, with epel-next-release requiring epel-release.  See [this build](https://koji.fedoraproject.org/koji/buildinfo?buildID=1845105).  Once epel9 is ready we'll add the epel repo config to epel-release by enabling [this bcond](https://src.fedoraproject.org/rpms/epel-release/blob/epel9-next/f/epel-release.spec#_1).  I'm expecting to be able to get those packages into the repo later this week.

praiskup:
I meant, shouldn't we do it "for now"?  Because it clearly makes `mock -r epel-next-9-x86_64` fail. And at the same time we try to make the configuration work [by a temporary hack](https://github.com/rpm-software-management/mock/blob/16db6b71c14dbf90145e4e32f5b4e9f098fe72ee/mock-core-configs/etc/mock/templates/epel-9.tpl#L1-L4) which makes no sense (if we still fail).

Also, I don't fully understand what we need `epel-release` and `fedpkg-minimal` for, don't you have more info about that? It was originally added in [RHBZ#1764445](https://bugzilla.redhat.com/1764445) but I'm not really sure what is it for. I also [asked](https://pagure.io/releng/issue/10369) the releng folks, as the Koji situation motivated us to add those originally in daa2de45c193300ba335145b1283316f84422e96.

Yes, `epel-rpm-macros` are needed (and till we have them, builds will fail I guess).

The last one needed - `epel-next-release` - what do we need it for?

> No, the initial packages will have the gpg key in epel-release

Mock uses keys from distribution-gpg-keys?

carlwgeorge:
You raise good points.  I set up epel9-next to replicate what we had with epel8-next, which was based on what was set up for epel8.  It does indeed seem that epel-release and epel-next-release are not needed.

[...]

PR #793, opened 2021-11-03
Reduce packages installed in epel chroots

carlwgeorge:
The repo definitions are provided by mock-core-configs and the gpg keys are provided by distribution-gpg-keys, so epel-release/epel-next-release are not needed.  fedpkg-minimal is needed in koji for obtaining sources from dist-git and the lookaside cache, but koji installs that via the srpm-build group.

https://pagure.io/releng/issue/10369

Related #790

### DIFF
--- a/mock-core-configs/etc/mock/templates/epel-8.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-8.tpl
@@ -1,4 +1,4 @@
-config_opts['chroot_setup_cmd'] += " epel-release epel-rpm-macros fedpkg-minimal"
+config_opts['chroot_setup_cmd'] += " epel-rpm-macros"
 
 config_opts['dnf.conf'] += """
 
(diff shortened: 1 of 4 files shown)


praiskup:
Thanks!  Looks fine, but if we are not in hurry, I'd wait for the fedpkg-minimal resolution before we merge.  I'm pretty confident we can drop that one too..

Conan-Kudo:
@carlwgeorge `fedpkg-minimal` is not required for Mock configs, only Koji configs, so please drop it along with the others.

praiskup:
> only Koji configs

Do you think it is actually needed in Koji configs elsewhere than in SRPM builds?  If so, I'd prefer to keep it here as well, so we don't diverge from Koji configs.

Conan-Kudo:
It cannot be used by Mock for anything, we don't have a way to invoke `fedpkg sources` in Mock.

praiskup:
Ok, thank you for the patch and for the review!

hroncok:
Can we please get this released on Fedora? It appears that mock-core-configs was not updated for a while and EPEL 9 roots cannot be initialized due to missing fedpkg-minimal.

xsuchy:
Pavel is going to do a release this week. There are still some changes we want to get into this release.

PR #797, opened 2021-11-14
Fix `bootstrap_image` in mock-core-configs templates

gotmax23:
- core-configs: Update almalinux-8.tpl bootstrap_image
- core-configs: Use fully qualified bootstrap_image name

### DIFF
--- a/mock-core-configs/etc/mock/templates/almalinux-8.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-8.tpl
@@ -3,7 +3,7 @@ config_opts['dist'] = 'el8.alma'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
-config_opts['bootstrap_image'] = 'docker.io/almalinux/almalinux:8'
+config_opts['bootstrap_image'] = 'docker.io/library/almalinux:8'
 
 
 config_opts['dnf.conf'] = """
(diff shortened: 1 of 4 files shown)


xsuchy:
I do not see anything relevant on https://wiki.almalinux.org/containers/docker-images.html
Can you provide a link to an announcement? Or an official source?

Conan-Kudo (approved):

Conan-Kudo:
> https://wiki.almalinux.org/containers/docker-images.html

Docker maps  `<name>` to `docker.io/library/<name>` automatically. This PR dereferences it.

mock-core-configs/etc/mock/templates/almalinux-8.tpl
```diff
@@ -3,7 +3,7 @@ config_opts['dist'] = 'el8.alma'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
-config_opts['bootstrap_image'] = 'docker.io/almalinux/almalinux:8'
+config_opts['bootstrap_image'] = 'docker.io/library/almalinux:8'
```
Conan-Kudo, line 6: I would suggest using `quay.io/almalinux/almalinux:8` instead. [It's an official image](https://quay.io/repository/almalinux/almalinux) and updated more frequently.

Conan-Kudo (changes requested):

gotmax23:
> I do not see anything relevant on https://wiki.almalinux.org/containers/docker-images.html Can you provide a link to an announcement? Or an official source?

Also, here[1] is the page that says that the previous image is deprecated.

[1]: https://hub.docker.com/r/almalinux/almalinux

mock-core-configs/etc/mock/templates/almalinux-8.tpl
gotmax23: > I would suggest using `quay.io/almalinux/almalinux:8` instead. [It's an official image](https://quay.io/repository/almalinux/almalinux) and updated more frequently.

I just did. We could also switch to `quay.io/almalinux/8-base` which is UBI-compatible. Then, we would be in line with `rhel8.tpl`. What do you think?

Conan-Kudo: That would probably work too.

Conan-Kudo (approved):

Conan-Kudo:
Thanks for the contributions!

PR #799, opened 2021-11-18
argparse: handle old-style commands *before* ignoring "--"

AdamWill:
If the attempt to eat "--" as the first unknown arg comes before
the handling of old-style commands, then it doesn't handle this:

mock -r something chroot -- commands

because the first unknown arg in that case is "chroot", not "--".
So we should do the old-style command handling first, *then*
check if the first unknown arg is "--" and eat it if so.

Fixes: https://bugzilla.redhat.com/2024620

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -397,13 +397,6 @@ def command_parse():
 
     (options, args) = parser.parse_known_args()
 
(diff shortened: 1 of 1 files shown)


praiskup:
This seems OK, thank you!

PR #801, opened 2021-11-20
Add option ssl-extra-certs

plaimbock:
As requested by @praiskup on the buildsys ML, here's my PR to add the ssl-extra-certs option. Reason: currently there are config options to add RHEL entitlement certificates and CA certificates in the chroot. What's missing is a similar option for regular (non-entitlement, non-CA) client/key certificates so mock can also access private (non-RHEL) repos from the chroot. On RHEL/Fedora hosts these certificates seem to live in /etc/pki/tls/certs and /etc/pki/tls/private. This patch allows you to specify both the client and key certificates with full path and the destination path in the chroot. Example:

config_opts['ssl_extra_certs'] = [
    '/etc/pki/tls/certs/client.crt', '/etc/pki/tls/certs/',
    '/etc/pki/tls/private/client_nopass.key.crt', '/etc/pki/tls/private/'
]

This PR adds the 3rd use case:
Works: Mock -> public repos like Fedora & CentOS
Works: Mock -> RHEL repos requiring an entitlement
Works: Mock + this PR -> public and private repos via cert & key in standard locations

Mailing list discussion: https://lists.fedoraproject.org/archives/list/buildsys@lists.fedoraproject.org/thread/5ZAACHTR7FVPYDIH4DWYYYGHNADNYOVE/

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -500,5 +500,16 @@
 # different configuration, and copying the bundle helps.
 #config_opts['ssl_ca_bundle_path'] = None
 
+# Copy host's SSL certificates into a specified location inside the chroot if
+# mock needs access to repositories which require client certificate
+# authentication. Specify the full path to the public certificate on the host
+# and the destination directory in the chroot. Do the same for the private key.
+# The private key should not be password-protected if you want mock to run
+# unattended.
+#config_opts['ssl_extra_certs'] = None
+# Example:
+#config_opts['ssl_extra_certs'] = ['/etc/pki/tls/certs/client.crt', '/etc/pki/tls/certs/',
+#                                  '/etc/pki/tls/private/client_nopass.key.crt', '/etc/pki/tls/private/']
+
 # user_agent string to identify HTTP request to repositories
 # config_opts['user_agent'] = "Mock ({{ root }}; {{ target_arch }})"
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -328,11 +328,29 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
                 # when mock is not run on Fedora or EL
                 self.buildroot.root_log.debug("ca bundle not found on host")
 
+    @traceLog()
+    def copy_extra_certs(self):
+        extra_certs = self.config['ssl_extra_certs']
+        if extra_certs:
+            self.buildroot.root_log.debug('copying extra certificates into chroot')
+            for cert_src, cert_dest in zip(extra_certs[::2], extra_certs[1::2]):
+                host_cert_src = os.path.realpath(cert_src)
+                chroot_cert_dest = self.buildroot.make_chroot_path(cert_dest)
+                chroot_cert_dir = os.path.dirname(chroot_cert_dest)
+                file_util.mkdirIfAbsent(chroot_cert_dir)
+                try:
+                    shutil.copy(host_cert_src, chroot_cert_dest)
+                except FileNotFoundError:
+                    # when mock is not run on Fedora or EL
+                    self.buildroot.root_log.debug("extra certificates not found on host")
+
+
     def initialize(self):
         self.copy_gpg_keys()
         self.copy_certs()
         if self.buildroot.is_bootstrap:
             self.copy_distribution_gpg_keys()
+            self.copy_extra_certs()
         self.initialize_config()
 
         try:
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -106,6 +106,8 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
 
     config_opts['ssl_ca_bundle_path'] = None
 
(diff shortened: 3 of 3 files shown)


mock/py/mockbuild/package_manager.py
```diff
@@ -328,11 +328,34 @@ def copy_certs(self):
...
+                    shutil.copy(host_cert_src, chroot_cert_dest)
+                except FileNotFoundError:
+                    # when mock is not run on Fedora or EL
+                    self.buildroot.root_log.debug("extra certificates not found on host")
+
+
     def initialize(self):
         self.copy_gpg_keys()
         self.copy_certs()
+        self.copy_extra_certs()
```
praiskup, line 355: Can this call be dropped?  The later call seems better, as that is only affecting the bootstrap chroot... and we don't really want to install the chroots to the target chroot?

praiskup:
The `./mock/docs/site-defaults.cfg` file needs a documentation update.

praiskup:
Thank you for the PR!

mock/py/mockbuild/package_manager.py
plaimbock: I tested it and yes it can be dropped. Done in https://github.com/rpm-software-management/mock/pull/801/commits/c69fda9051c491bf91c22403f1ba4e301e2c2fad

plaimbock:
> The `./mock/docs/site-defaults.cfg` file needs a documentation update.
I added documentation in https://github.com/rpm-software-management/mock/pull/801/commits/23abb30c177b0a13dd2dc6786ee8ab36a4c86264

mock/docs/site-defaults.cfg
```diff
@@ -498,7 +498,16 @@
...
+# config_opts['ssl_ca_bundle_path'] = None
+
+# Copy host's SSL certificates into a specified location inside the chroot if
+# mock needs access to repositories which require client certificate
+# authentication. Specify the full path to the public certificate on the host
+# and the destination directory in the chroot. Do the same for the private key.
+# The private key should not be password-protected if you want mock to run
+# unattended.
+# config_opts['ssl_extra_certs'] = ['/etc/pki/tls/certs/client.crt', '/etc/pki/tls/certs/',
+#                                   '/etc/pki/tls/private/client_nopass.key.crt', '/etc/pki/tls/private/']
```
xsuchy, line 510: The values in site=defaults.cfg are the defaults. But the default for ssl_extra_cert is None.
This example is fine if you clearly mark it as example **and** state the default.

mock/py/mockbuild/package_manager.py
```diff
@@ -328,11 +328,33 @@ def copy_certs(self):
...
+    def copy_extra_certs(self):
+        extra_certs = self.config['ssl_extra_certs']
+        if extra_certs:
+            self.buildroot.root_log.debug('copying extra certificates into chroot')
+            for cert_src, cert_dest in zip(extra_certs[::2], extra_certs[1::2]):
+                host_cert_src = os.path.realpath(cert_src)
+                chroot_cert_dest = self.buildroot.make_chroot_path(cert_dest)
+                chroot_cert_dir = os.path.dirname(chroot_cert_dest)
+                try:
+                    os.makedirs(chroot_cert_dir)
```
xsuchy, line 341: Can you use file_util.mkdirIfAbsent() ? Then you can remove that try/except block.

mock/docs/site-defaults.cfg
plaimbock: Thank you for pointing that out. Fixed.

mock/py/mockbuild/package_manager.py
plaimbock: Done.

mock/docs/site-defaults.cfg
```diff
@@ -498,7 +498,18 @@
 # the whole /etc/pki/ca-trust/extracted directory recursively by default, and
 # Fedora or EL systems work with that.  But some destination chroots can have
 # different configuration, and copying the bundle helps.
-#config_opts['ssl_ca_bundle_path'] = None
+# config_opts['ssl_ca_bundle_path'] = None
```
praiskup, line 501: Nit, this is an intentional style (no space).  It helps the reader to more easily tell what
is the documentation text (space after the hash) and what is a commented-out code.

mock/docs/site-defaults.cfg
praiskup: I know it is a matter of taste, but this config option could easily be defined as a list of pairs, or a dict with {"source": "dest"}

praiskup:
Otherwise looks good to me.  Can you also please squash the commits into a single one, and mention  "Merges: #801" on a separate line in the commit message?

mock/docs/site-defaults.cfg
plaimbock: Fixed.

mock/docs/site-defaults.cfg
plaimbock: Given my limited pyFoo I just followed def copy_certs(self) so no taste was involved, just the quickest path to a solution :-)

praiskup:
Merged, thank you!

PR #802, opened 2021-11-22
config: epel move: CentOS+EPEL => RHEL+EPEL

praiskup:
CentOS 8 support ends at the end of Y2021, and we need to move on
WRT our default EPEL configuration.

The official EPEL packages are built (in Fedora Koji) against RHEL
repositories + EPEL repositories, so we try to mimic this here to be as
close to Koji as possible.

The obvious consequence here is that, from now on, users will need a Red
Hat subscription to build against epel-8-* chroots and newer (but the
free developer subscription is enough):

  https://rpm-software-management.github.io/mock/Feature-rhelchroots

Another problem is that Mock's forcearch (emulated) compilation isn't
working ATM: https://bugzilla.redhat.com/show_bug.cgi?id=1912847

Relates: https://bugzilla.redhat.com/show_bug.cgi?id=1965480

### DIFF
--- a/mock-core-configs/etc/mock/epel-8-aarch64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-aarch64.cfg
@@ -1,6 +1,4 @@
-include('templates/centos-8.tpl')
+include('rhel-8-aarch64.cfg')
 include('templates/epel-8.tpl')
 
-config_opts['root'] = 'epel-8-aarch64'
-config_opts['target_arch'] = 'aarch64'
-config_opts['legal_host_arches'] = ('aarch64',)
+config_opts['root'] = "epel-8-{{ target_arch }}"
--- a/mock-core-configs/etc/mock/epel-8-ppc64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-ppc64.cfg
@@ -1,4 +1,4 @@
 include('rhel-8-ppc64.cfg')
 include('templates/epel-8.tpl')
 
-config_opts['root'] = "rhelepel-8-{{ target_arch }}"
+config_opts['root'] = "epel-8-{{ target_arch }}"
--- a/mock-core-configs/etc/mock/epel-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/epel-8-ppc64le.cfg
@@ -1,6 +1,4 @@
-include('templates/centos-8.tpl')
+include('rhel-8-ppc64le.cfg')
 include('templates/epel-8.tpl')
 
-config_opts['root'] = 'epel-8-ppc64le'
-config_opts['target_arch'] = 'ppc64le'
-config_opts['legal_host_arches'] = ('ppc64le',)
+config_opts['root'] = "epel-8-{{ target_arch }}"
--- a/mock-core-configs/etc/mock/epel-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-x86_64.cfg
@@ -1,6 +1,4 @@
-include('templates/centos-8.tpl')
+include('rhel-8-x86_64.cfg')
 include('templates/epel-8.tpl')
 
-config_opts['root'] = 'epel-8-x86_64'
-config_opts['target_arch'] = 'x86_64'
-config_opts['legal_host_arches'] = ('x86_64',)
+config_opts['root'] = "epel-8-{{ target_arch }}"
--- a/mock-core-configs/etc/mock/rhel-8-ppc64.cfg
+++ b/mock-core-configs/etc/mock/rhel-8-ppc64.cfg
@@ -1,0 +1,4 @@
+include('templates/rhel-8.tpl')
+
+config_opts['target_arch'] = 'ppc64'
+config_opts['legal_host_arches'] = ('ppc64',)
--- a/mock-core-configs/etc/mock/rhelepel-8-aarch64.cfg
+++ b/mock-core-configs/etc/mock/rhelepel-8-aarch64.cfg
@@ -1,4 +1,0 @@
-include('rhel-8-aarch64.cfg')
-include('templates/epel-8.tpl')
-
-config_opts['root'] = "rhelepel-8-{{ target_arch }}"
--- a/mock-core-configs/etc/mock/rhelepel-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/rhelepel-8-ppc64le.cfg
@@ -1,4 +1,0 @@
-include('rhel-8-ppc64le.cfg')
-include('templates/epel-8.tpl')
-
-config_opts['root'] = "rhelepel-8-{{ target_arch }}"
--- a/mock-core-configs/etc/mock/rhelepel-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/rhelepel-8-x86_64.cfg
@@ -1,4 +1,0 @@
-include('rhel-8-x86_64.cfg')
-include('templates/epel-8.tpl')
-
-config_opts['root'] = "rhelepel-8-{{ target_arch }}"


Conan-Kudo:
We should not require RHEL since subscription management configuration on the host is required to use it. That will not work for all downstream consumers of our configuration.

praiskup:
Well, I don't see a better **default** option, I'm afraid.  We had rhelepel-* configs
in mock-core-configs for quite some time and it was tested before every release.

If anyone needed a non-standard EPEL configuration, I think we can easily add other
configuration later (e.g. almalinux-epel, or eurolinux-epel, or navy-epel?  Or do you
think that some of those should be added right now?

Note that we need to move on in Fedora Copr, too.  We need to keep `epel-8*`
chroots there, and it would be really a pitty to have a different configuration for
EPEL there.

hroncok:
As a default option, this means `fedpkg mockbuild` on epel-8 branch no longer works out of the box, making contributing to EPEL 8 harder to set up. Using this as a default in Copr might make sense, because as the user, I don't need to care about the subscription. But using it as default in mock worries me.

praiskup:
Someone would have to decide what fork we are going to promote now as the
CentOS replacement, and set it to be the default...  Which of those
(alphabetical order):

```
almalinux-8
centos-stream-8
eurolinux-8
navy-8
oraclelinux-8
rocky-8
```

There are probably others to be added in the future, and I don't really want to decide
which is better, and claim that they are better than subscribed RHEL.  Eventually, getting
the subscription is pretty trivial IMO.

@hroncok wrote
> fedpkg mockbuild

This is good point.  There will be probably other tools with this...  Though
I think that it is good idea to let Fedora project vote on what the fedpkg utility will do
when checked out in the `epel8` branch, and mock-upstream we can eventually add any
`<foo>-epel-8` config.  So `fedpkg mockbuild` would just run `mock -r foo-epel-8-<arch>`?

hroncok:
That'll work for me at least.

praiskup:
For the record, I created [new issue against fedpkg](https://pagure.io/fedpkg/issue/457) to discuss this.

hroncok:
Summary of my views based on the recent discussion on the mailing lists, here, and in the Alma PR:

 - I like the idea of epel configs to be configurable-aliases to one of the foo-epel configs :tada: 
 - the default value might as well be RHEL, as the historical usage of centos was indeed tricky -- but this needs to be communicated very loudly :mega: 
 - the default mock config used for `fedpkg --release epel8 mockbuild` should IMHO be decided by EPSCo :judge: and not by `fedpkg` engineers who are not EPEL packagers at all (sorry if I got that claim wrong, it's based on my view as an outsider)

Conan-Kudo:
I've made the ask to EPSCo here: https://pagure.io/epel/issue/133

praiskup:
> I like the idea of epel configs to be configurable-aliases to one of the foo-epel configs 

We can talk about this in the future, but IMO it is complicated :-/  See how default.cfg is handled.
Not impossible, though!

> the default value might as well be RHEL, as the historical usage of centos was indeed tricky -- but this needs to be communicated very loudly mega

I think we could print a very concrete warning on stdout for `-r epel-8-x86_64` when there's no subscription,
that people can also use `<foo>-epel-*` configs.

> the default mock config used for fedpkg --release epel8 mockbuild should IMHO be decided by EPSCo judge and not by fedpkg engineers who are not EPEL packagers at all (sorry if I got that claim wrong, it's based on my view as an outsider)

Absolutely.  Though Mock/Copr upstream is something different than fedpkg.  If EPSCo decides to go
(convenience reasons over correctness) with alma+epel as default - it can be implemented in fedpkg
easily.

Conan-Kudo:
Mock is not the same as COPR though. If Copr wants to use RHEL+EPEL instead of Alma+EPEL, that's its choice. Copr doesn't have the same issues that plain Mock has.

praiskup:
The big challenge in Fedora Copr is to be as close to the user local environment as possible.
Therefore we don't ever change (except for short transition periods) the configuration shipped
in the `mock-core-configs`.  This has major benefits -- when mock-core-configs gets broken,
Copr is broken (and vice versa).  This makes the repo bugs forwarded to the right hands much
faster.

carlwgeorge:
> I think we could print a very concrete warning on stdout for `-r epel-8-x86_64` when there's no subscription,
that people can also use `<foo>-epel-*` configs.

I love this idea.  The existence of the alternative epel configs and this warning would make this basically a non-issue.

**user**: Build for this epel target.
**mock**: You don't have a subscription, here are docs on how to set that up, or use a different epel config that's probably close enough.
**user**: Ah fair enough.

Conan-Kudo:
I would rather we delete the EPEL config entirely if the alternative is that RHEL is the default. RHEL as the default is too unfriendly.

praiskup:
> I would rather we delete the EPEL config entirely if the alternative is that RHEL is the default. RHEL as the default is too unfriendly.

But still 100% honest, so I think Mock users deserve that.  Would the "warning" change your opinion here?

Conan-Kudo:
> > I would rather we delete the EPEL config entirely if the alternative is that RHEL is the default. RHEL as the default is too unfriendly.
> 
> But still 100% honest, so I think Mock users deserve that. 

I still don't understand what "honest" means here...

> Would the "warning" change your opinion here?

No, because warnings are not visible by people using Mock in automation. The problem with RHEL is that RHEL folks will not make RHEL easy to consume to support all the various Mock capabilities and use-cases, which is why we rely on community rebuilds to close that gap.

praiskup:
> I still don't understand what "honest" means here...

It's OK to **admit** that to be as close as reasonably possible to the official EPEL
build system, users have to have official RHEL content (with all the inconvenience).
It is not our fault in Mock upstream, and it is not unfair to admit that.  At all, quite the
opposite (better than trying to hide that something is actually changing behind
the scenes).

> No, because warnings are not visible by people using Mock in automation.

The fact they don't have a subscription will bring the warning to their visibility?
So I fail to understand how "no config" is better than "non working with clear
warning".

Conan-Kudo:
> > I still don't understand what "honest" means here...
> 
> It's OK to **admit** that to be as close as reasonably possible to the official EPEL build system, users have to have official RHEL content (with all the inconvenience). It is not our fault in Mock upstream, and it is not unfair to admit that. At all, quite the opposite (better than trying to hide that something is actually changing behind the scenes).
> 

Official RHEL content isn't an inconvenience for Koji or COPR in the same way it is for regular users. Both of those systems would have protected import mirrors that they would use instead of RHSM. That also removes all RHSM-related inhibitors.

RHEL vs a well-done RHEL clone (like AlmaLinux) provides no meaningful differences. And using a RHEL clone means that users can freely work with EPEL content without worrying about host-specific stuff like subscription-manager and RHEL entitlement management.

> > No, because warnings are not visible by people using Mock in automation.
> 
> The fact they don't have a subscription will bring the warning to their visibility? So I fail to understand how "no config" is better than "non working with clear warning".

No config immediately breaks everything and people are forced to adapt.

praiskup:
> Official RHEL content isn't an inconvenience for Koji or COPR in the same way it is for regular users. Both of those systems would have protected import mirrors that they would use instead of RHSM.

Copr will be using RHSM.  That's the plan, we want to be as close to the defaults in mock-core-configs default (see
the point somewhere above).  And yes, this is an immediate inconvenience for Fedora Copr admins at this moment.

Conan-Kudo:
> > Official RHEL content isn't an inconvenience for Koji or COPR in the same way it is for regular users. Both of those systems would have protected import mirrors that they would use instead of RHSM.
> 
> Copr will be using RHSM. That's the plan, we want to be as close to the defaults in mock-core-configs default (see the point somewhere above). And yes, this is an immediate inconvenience for Fedora Copr admins at this moment.

That will be a major problem given that you dynamically spawn workers and such. Even if they ever fix the inability to access content across all arches (which is unlikely given the consequences it'd have for system repositories), you'll have problems tracking the provisioning because you'd be consuming entitlements for each run.

praiskup:
We need to unsubscribe when the builder VM is terminated, yes.

kkofler:
I think there are only 2 reasonable defaults: Rocky Linux or AlmaLinux (just pick whatever of those works better in practice, it should not really matter). Anything that requires a subscription of any sort, even a free-as-in-beer one, should be a non-starter. The default EPEL config needs to just work and to be free of field-of-use restrictions, including technically enforced policy restrictions breaking use cases such as emulation. So requiring a RHEL developer subscription is not an acceptable solution. The discrepancy with the EPEL Koji and/or Copr configuration can easily be fixed by changing those, too.

Conan-Kudo:
Closing in favor of #814.

PR #803, opened 2021-11-22
config: Migrate EPEL 8 configuration to use AlmaLinux 8 as the base OS

Conan-Kudo:
CentOS Linux 8 support ends at the end of 2021, so we need to select
a new base OS for the default EPEL configuration.

AlmaLinux satisfies the requirements to replace CentOS Linux for the
EPEL configuration, namely that:

* Alma provides the exact same content as Red Hat Enterprise Linux 8
* Alma releases the same time as RHEL (Alma 8.5 was 2 days after RHEL 8.5)
* AlmaLinux has committed to supporting all RHEL architectures
* AlmaLinux is freely available with no entitlement encumbrances

This allows all users of Mock to (re)build EPEL packages to successfully
do so regardless of host Linux distribution Mock runs on.

Resolves: [rhbz#1965480](https://bugzilla.redhat.com/show_bug.cgi?id=1965480)

### DIFF
--- a/mock-core-configs/etc/mock/epel-8-aarch64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-aarch64.cfg
@@ -1,4 +1,4 @@
-include('templates/centos-8.tpl')
+include('templates/almalinux-8.tpl')
 include('templates/epel-8.tpl')
 
 config_opts['root'] = 'epel-8-aarch64'
--- a/mock-core-configs/etc/mock/epel-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/epel-8-ppc64le.cfg
@@ -1,4 +1,4 @@
-include('templates/centos-8.tpl')
+include('templates/almalinux-8.tpl')
 include('templates/epel-8.tpl')
 
 config_opts['root'] = 'epel-8-ppc64le'
--- a/mock-core-configs/etc/mock/epel-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-8-x86_64.cfg
@@ -1,4 +1,4 @@
-include('templates/centos-8.tpl')
+include('templates/almalinux-8.tpl')
 include('templates/epel-8.tpl')
 
 config_opts['root'] = 'epel-8-x86_64'
--- a/mock-core-configs/etc/mock/epelplayground-8-aarch64.cfg
+++ b/mock-core-configs/etc/mock/epelplayground-8-aarch64.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/centos-8.tpl')
+include('/etc/mock/templates/almalinux-8.tpl')
 include('/etc/mock/templates/epelplayground-8.tpl')
 
 config_opts['root'] = 'epelplayground-8-aarch64'
--- a/mock-core-configs/etc/mock/epelplayground-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/epelplayground-8-ppc64le.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/centos-8.tpl')
+include('/etc/mock/templates/almalinux-8.tpl')
 include('/etc/mock/templates/epelplayground-8.tpl')
 
 config_opts['root'] = 'epelplayground-8-ppc64le'
--- a/mock-core-configs/etc/mock/epelplayground-8-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epelplayground-8-x86_64.cfg
@@ -1,4 +1,4 @@
-include('/etc/mock/templates/centos-8.tpl')
+include('/etc/mock/templates/almalinux-8.tpl')
 include('/etc/mock/templates/epelplayground-8.tpl')
 
 config_opts['root'] = 'epelplayground-8-x86_64'


Conan-Kudo:
cc: @jaboutboul, @hroncok

Conan-Kudo:
@praiskup This is my alternative for #802

Conan-Kudo:
I've been talking to @jaboutboul, @andrewlukoshko, and @ezamriy about this for a few months now. They've told me that they're close to having AlmaLinux also available for ppc64le, which is probably the only remaining blocker for this change.

praiskup:
> AlmaLinux satisfies the requirements to replace CentOS Linux for the EPEL configuration

As discussed in the original #802, we have to ask why Alma (in this case), and not any other
project by "default" (but let's keep this in the other issue).

Can you please add those as `alma-epel-8-*` (or pick a preferred naming) in the meantime
please?  That would be 100% valid.

Conan-Kudo:
> As discussed in the original #802, we have to ask why Alma (in this case), and not any other
project by "default" (but let's keep this in the other issue).

I put the justification in the commit message.

> Can you please add those as `alma-epel-8-*` (or pick a preferred naming) in the meantime
please? That would be 100% valid.

I can add those as a separate commit, sure.

Conan-Kudo:
@praiskup I've added `almaepel-8-*` configs now, matching the scheme that `rhelepel-8-*` configs have.

Conan-Kudo:
`oraclelinux-epel-8-*` is not our EPEL, it is the OL EPEL project.

praiskup:
> I can add those as a separate commit, sure.

The additional config could go to `main` right away, FTR.  So separate PR would be better.

Switching `epel-*` config to AlmaLinux is something I personally vote against, at least for now
as it is too fresh.

Conan-Kudo:
> Switching `epel-*` config to AlmaLinux is something I personally vote against, at least for now
as it is too fresh.

What do you mean by "too fresh"?

carlwgeorge:
> `oraclelinux-epel-8-*` is not our EPEL, it is the OL EPEL project.

Gross, but not my point.  I was just suggesting putting a dash between the distro and epel.

Unrelated to this PR, but I don't think we should be encouraging Oracle's overloading of the EPEL name.  Their "EPEL" does not contain all EPEL packages, and also contains packages that they have not submitted to the actual EPEL.  I'll send a separate PR to propose fixing that.

Conan-Kudo:
> That's not good, but that wasn't my point. I was just suggesting putting a dash between the distro and epel.

I agree with you, but I'd rather fix that in one go once it's less confusing...

praiskup:
> What do you mean by "too fresh"?

Submitted 32 minutes ago.

Conan-Kudo:
Well, sure, my PR was submitted a half hour ago, but I've been using this for a while locally. I was mainly holding back until ppc64le was published by AlmaLinux so that all of them can be switched at once.

carlwgeorge:
I suggest we convert this pr to just adding `alma-epel-8-*` configs to give people something they can start using, and leave the decision/action on the plain `epel-8-*` configs separate.

Conan-Kudo:
> I suggest we convert this pr to just adding `alma-epel-8-*` configs to give people something they can start using, and leave the decision/action on the plain `epel-8-*` configs separate.

Sorry, I can't as long as #802 exists. RHEL is an awful default because the entitlement system is such a mess and a pain to deal with for packagers, so this is my counter for that.

mock-core-configs/etc/mock/almaepel-8-ppc64le.cfg
```diff
@@ -0,0 +1,6 @@
+include('templates/almalinux-8.tpl')
+include('templates/epel-8.tpl')
+
+config_opts['root'] = 'almaepel-8-ppc64le'
+config_opts['target_arch'] = 'ppc64le'
```
carlwgeorge, line 5: Alma doesn't have ppc64le.

Conan-Kudo: I know, that's why this is a draft.

Conan-Kudo:
I've split out `almaepel-8-*` configs into #804.

kwizart:
I don't see why mock-base-configs needs to be alma gnostic, both rocky and alma are reasonable replacements distro for CentOS Linux to me, but it can be decided in the back of users. So all flavor needs to be supported, namely:
- ALMA
- RHEL
- Rocky
- OL
I think this can be reasonably achieved by using "alternatives". Of course, when using Rocky Linux, it seems logical to select rocky as a default provider for EPEL content, and etc.

carlwgeorge:
I'm opposed to this because it will fuel conspiracy theories among rebuilds that are not selected as the default.  This is not speculation on my part; I've already seen it happening.  I don't think the mock developers want to be accused of having a secret alliance with Alma against another rebuild.  The best choice is the neutral one, to have `epel-8-*` use `rhel-8-*` which matches actual EPEL builds in koji.  Additional configs like `alma+epel-8-*` and `oraclelinux-epel-8-*` are already available and any rebuild can send a pull request to include their own as well.

Conan-Kudo:
> I'm opposed to this because it will fuel conspiracy theories among rebuilds that are not selected as the default. This is not speculation on my part; I've already seen it happening. I don't think the mock developers want to be accused of having a secret alliance with Alma against another rebuild. 

There's no "secret alliance" here. I listed the reasons why I went with Alma in the commit message.

> The best choice is the neutral one, to have `epel-8-*` use `rhel-8-*` which matches actual EPEL builds in koji. Additional configs like `alma+epel-8-*` and `oraclelinux-epel-8-*` are already available and any rebuild can send a pull request to include their own as well.

It's actually the worst choice because it's incredibly difficult to use RHEL for package builds. Here are a couple of examples of problems with it:

* Cross-architecture content access is blocked, as subscription manager will refuse to grant you access.
* `subscription-manager` is not available in most distributions that Mock is available on. For users on Debian/Ubuntu, Mageia, OpenMandriva, openSUSE, ROSA, and a number of RHEL rebuild distributions, forcing `subscription-manager` for the default contribution experience will add too much friction for people to leverage EPEL and contribute to it.

My goal is to keep Fedora EPEL as an easy to leverage and contribute project. Forcing RHEL and all the foibles of subscription-manager is a great way to make people's lives miserable.

praiskup:
I'm not sure it is 100% clear here, but we are mostly talking about **the
default epel-8-ARCH** configs, nobody will ever block anyone from using
`alma+epel-8-*` or `rocky+epel-8-*` (once they exist).  It should actually
help the overall situation in understanding what `epel-*` chroot is.

I also argued against CentOS+EPEL nuance a long time before I was in Copr
team.  It has bitten me several times before as the user in Copr and normal
packager's life...  Simply, epel-* configs have not been honest in what they
really are.  And we "only" have not been forced to use `centos+epel-*` string while
working with mock options, so far.

So at the end of the day, what we are discussing now is just that
people who are unable to get subscription-manager for any reason (and I
fail to understand the technical problems behind installing and using
that, or getting that packaged if we want to go extreme) will have to do
use one of:

    mock -r alma+epel-8-x86_64 ...
    mock -r rocky+epel-8-x86_64 ...

instead of just

    mock -r epel-8-x86_64 ...

Any change on users is not ideal, but changes sometimes happen.  The important
thing now to consider is understanding what EPEL really is, and what it isn't.  EPEL
was never officially built against CentOS, and likely never will be against Alma.
It always was, and likely will be RHEL based.  I don't think there will be a better
opportunity to do the switch than now - when we actually *have to do something*.
Moreover, the reason why we are discussing the change now is that we built
against the non-default base and **that** decision forces us towards this change.

People will also be more open to trying alternatives, as they will be pushed
to understand what they really build against.  Is `epel-8` is broken?  Try `alma+epel`.
Broken?  Try `rocky+epel`.  Broken?  Try others.

Conan-Kudo:
@praiskup Would you rather I take this to the EPEL Steering Committee so they can give the final verdict about this? If that'll help, then I'll ask them to give their opinion in their meeting on Wednesday.

gotmax23:
> I think this can be reasonably achieved by using "alternatives". Of course, when using Rocky Linux, it seems logical to select rocky as a default provider for EPEL content, and etc.

I think this is an elegant solution. We should create separate epel configs for each EL distro and then have `/etc/mock/epel-8-*.cfg` be symlinks. I think it makes sense for Almalinux to be the default for the reasons @Conan-Kudo mentioned. Is switching those files to a symlinks feasable? I know that rpm does not always behave properly when filetypes change.

Conan-Kudo:
file->file is certainly possible, but managing that with alternatives is going to be very ugly.

xsuchy:
Hmm I can imagine someting like:
```
epel-8-x86_64.cfg:
include('templates/{{default_rhel}}-8.tpl')                                                                                                                                                         
include('templates/epel-8.tpl')
```

With default:
`config_opts['default_rhel'] = 'rhel' # or almalinux or anything else`

And users can very easily override this in their custom config.

Mind that this syntax will not work now, because we do not expand string passed to include(), but that can be easily changed.

Conan-Kudo:
> Hmm I can imagine someting like:
> 
> ```
> epel-8-x86_64.cfg:
> include('templates/{{default_rhel}}-8.tpl')                                                                                                                                                         
> include('templates/epel-8.tpl')
> ```
> 
> With default: `config_opts['default_rhel'] = 'rhel' # or almalinux or anything else`
> 
> And users can very easily override this in their custom config.
> 
> Mind that this syntax will not work now, because we do not expand string passed to include(), but that can be easily changed.

If we did that, then the default would have to be something that *does not* require dealing with subscription-management, because it needs to work for people on all platforms Mock is available and used on.

praiskup:
This is not that kind of "we have to" problem;  there's a **correct** solution, and so we should prefer that.
People will just go with `alma+epel` if they **want to**.  See the difference, we are arguing about, 6 bytes.

There a major negative blocker for me here, and that's the inconsistency with Fedora Copr.  I don't want
to put a different meaning to `epel-8-x86_64` in Fedora Copr vs. Mock.  **Finally** we can synchronize
Koji (epel = rhel), Mock (the conflict here), and Copr (will be RHEL by the end of the year).

Conan-Kudo:
> This is not that kind of "we have to" problem; there's a **correct** solution, and so we should prefer that. People will just go with `alma+epel` if they **want to**. See the difference we are arguing about - 5 bytes.

It's only correct if it's not regressive. Or you can get someone to make it _not_ regressive. I do not believe that will happen anytime soon.

Conan-Kudo:
I've made the ask to EPSCo here: https://pagure.io/epel/issue/133

praiskup:
The situation in the CentOS ecosystem is changing, with negative effects too (but
not only).  So calling this a Mock regression is IMO not appropriate. :-(  Again, what
we talk about here is that we "admit" by `alma+` prefix what we really build against.
Not an unacceptable inconvenience (considering how big the overall change is) IMO.

hroncok:
I wonder if the best option would not be to deprecate the `epel` configs entirely and demand an explicit choice.

praiskup:
Yeah, I was also thinking about proposing that, but it is less correct (at least in my opinion) so I view it as a fallback solution.

carlwgeorge:
EPEL builds against RHEL.  It's past time to stop hiding that from mock users by using a RHEL clone in the EPEL mock config.  There is no regression because alternative mock configs exist for the people that don't want to set up a subscription.

Conan-Kudo:
> EPEL builds against RHEL. It's past time to stop hiding that from mock users by using a RHEL clone in the EPEL mock config. There is no regression because alternative mock configs exist for the people that don't want to set up a subscription.

Then stop RHEL from hiding its content behind subscription-manager and make it useful for cross-arch stuff. You and I both know that'll never happen, and we both know how much of a delta in practice these things are. The main problem with CentOS Linux in the past was how long it took between RHEL point release to CentOS Linux point release. This is not a real issue with AlmaLinux because they get these things out the door within 48 hours of a RHEL release.

fatherlinux:
Here's a crazy idea, what if we set this to Red Hat Universal Base Image which doesn't have the same subscription-manager challenges? I know there are dependency problems, but what if we could fix those?

Conan-Kudo:
> Here's a crazy idea, what if we set this to Red Hat Universal Base Image which doesn't have the same subscription-manager challenges? I know there are dependency problems, but what if we could fix those?

Then I'd be totally okay with that! If RHEL UBI provided the entire collection, then I'd be the first to say "let's pull the trigger!"

hroncok:
But it does not.

gotmax23:
> But it does not.

Exactly. Until UBI reaches parity with actual RHEL, I believe that using that as the default is the worst of both worlds. It will confuse users, because it seems to work fine at first... until they come across a missing dependency. Although, we could probably remedy that by including a strong warning.

kkofler:
I think there are only 2 reasonable defaults: Rocky Linux or AlmaLinux (just pick whatever of those works better in practice, it should not really matter). Anything that requires a subscription of any sort, even a free-as-in-beer one, should be a non-starter. The default EPEL config needs to just work and to be free of field-of-use restrictions, including technically enforced policy restrictions breaking use cases such as emulation. So requiring a RHEL developer subscription is not an acceptable solution. The discrepancy with the EPEL Koji and/or Copr configuration can easily be fixed by changing those, too.

@praiskup:
> As discussed in the original #802, we have to ask why Alma (in this case), and not any other
> project by "default" (but let's keep this in the other issue).

I do not care whether Alma or Rocky is the default. I guess most users will in fact not care. They are supposed to be binary-compatible. Just pick whatever works better (e.g.: gets updates faster, etc.). But any of those will work better than RHEL for users, and that is what should matter here.

@kwizart:
> I don't see why mock-base-configs needs to be alma gnostic, both rocky and alma are reasonable replacements distro for CentOS Linux to me, but it can't be decided in the back of users. So all flavor needs to be supported, namely:
> 
> * ALMA
> * RHEL
> * Rocky
> * OL

See above, I think any of the others will work better than RHEL for users. (That said, I would pick Alma or Rocky over OL because Oracle is known for doing their own thing, such as shipping custom kernels, shipping their own fork of EPEL as pointed out by others in this thread, etc.)

>   I think this can be reasonably achieved by using "alternatives". Of course, when using Rocky Linux, it seems logical to select rocky as a default provider for EPEL content, and etc.

`alternatives` is broken by design because it is a systemwide, global choice, i.e., the choice can only be made by root, and it will affect all builds running at the same time. @xsuchy's `default_rhel` idea, if it can also be overridden in a per-user config file or even on the command-line, sounds like a better approach to me. But that still does not address the most important question here, i.e., what should be the default. And I also share the concern that a configurable default would actually increase rather than decrease user confusion (especially if downstreams then change their default as you suggest).

@carlwgeorge:
> I'm opposed to this because it will fuel conspiracy theories among rebuilds that are not selected as the default. This is not speculation on my part; I've already seen it happening. I don't think the mock developers want to be accused of having a secret alliance with Alma against another rebuild. The best choice is the neutral one, to have `epel-8-*` use `rhel-8-*` which matches actual EPEL builds in koji. Additional configs like `alma+epel-8-*` and `oraclelinux-epel-8-*` are already available and any rebuild can send a pull request to include their own as well.

Please do not settle for a compromise that, for fear of offending a handful developers, actually leads to the worst possible user experience. Even if I were a developer of rebuild A, I would prefer to see mock default to rebuild B rather than to the original which I am (would be) rebuilding for a reason. If some developers really cannot grasp that, it is their loss. The user's perspective is what counts here.

@praiskup:
> There a major negative blocker for me here, and that's the inconsistency with Fedora Copr. I don't want to put a different meaning to `epel-8-x86_64` in Fedora Copr vs. Mock. **Finally** we can synchronize Koji (epel = rhel), Mock (the conflict here), and Copr (will be RHEL by the end of the year).

Why not just fix EPEL in Koji to build against something easily reproducible by the community? The Koji config is really the odd one out and the one that should be fixed. There are several things that were hindered or outright prevented by EPEL building against RHEL rather than (at the time) CentOS, e.g., no support for CentOS 7 i?86 (32-bit), and (repeatedly) packages that cannot be built due to missing subpackages in RHEL. It also required special support in Koji to build against an outside, private repository, whereas a rebuild could simply have been imported straight into Koji, or at least pulled from a public mirror that everyone can access.

Conan-Kudo:
> Why not just fix EPEL in Koji to build against something easily reproducible by the community? The Koji config is really the odd one out and the one that should be fixed. 

The reason Koji uses RHEL is primarily to expose RHEL to the community. When EPEL builds fail due to things like unshipped packages or bugs in the packages that are shipped, RHEL is supposed to care and resolve those problems, because most RHEL customers consider EPEL mandatory to a usable RHEL experience. You take that away, and Red Hat winds up having no incentive to do things like add back packages so that EPEL builds work.

However, this reason isn't good enough to make local users suffer through the issues that RHEL causes people when trying to use it for local package development.

kkofler:
And why would we care whether RHEL makes it even more painful to use their official binaries for development than it already does? It would just mean more developers moving to a rebuild and not cause any problems for EPEL or Fedora. Just maybe for RH if they lose paying customers (but then they will have to care about fixing their mess). So why does EPEL have to be the beta tester for setting a limit to RHEL developer unfriendliness? (I consider even the mere concept of not shipping built subpackages until someone complains to be inherently user- and developer-unfriendly, so I do not see the value in EPEL being the "someone" who complains about each individual missing subpackage.)

carlwgeorge:
> Please do not settle for a compromise that, for fear of offending a handful developers, actually leads to the worst possible user experience.

It's not a compromise, it's the more correction solution.  And I still have not heard a convincing argument about why the `{clone}+epel-*` configs do not satisfy the people that don't want to deal with a subscription.  Not to mention the fact that these files are `%config(noreplace)` and you can easily change your local version to use whatever you like.

> Why not just fix EPEL in Koji to build against something easily reproducible by the community?

One of the founding principles of EPEL is that it builds against RHEL.  Building against a RHEL clone can expose EPEL packagers to inconsistencies and bugs in the clone distributions that are not present in RHEL, as well as their rebuild delay, which will always be longer than the time it takes to sync the released RHEL content into the EPEL koji buildroot.  EPEL isn't changing.  Mock is aligning with reality.

Conan-Kudo:
> And why would we care whether RHEL makes it even more painful to use their official binaries for development than it already does? It would just mean more developers moving to a rebuild and not cause any problems for EPEL or Fedora. Just maybe for RH if they lose paying customers (but then they will have to care about fixing their mess).

Well, if RHEL isn't exposed to the needs of real people, or worse isn't used and people don't pay for it, you don't _get_ RHEL rebuilds _either_.

> So why does EPEL have to be the beta tester for setting a limit to RHEL developer unfriendliness? (I consider even the mere concept of not shipping built subpackages until someone complains to be inherently user- and developer-unfriendly, so I do not see the value in EPEL being the "someone" who complains about each individual missing subpackage.)

While on some level I agree with you, I also get the reason why Red Hat does this. A big part of how RHEL is supportable is that Red Hat works very hard to provide a few, well-defined integration points that they understand rather than lots of poorly-understood ones. It makes sense to only expand those paths as people need them and they can allocate resources to support those paths.

Conan-Kudo:
Closing in favor of #814.

PR #804, opened 2021-11-22
config: Add AlmaLinux+EPEL configs

Conan-Kudo:
This allows people to use AlmaLinux as the base OS for building packages
for Fedora EPEL.

### DIFF
--- a/mock-core-configs/etc/mock/alma+epel-8-aarch64.cfg
+++ b/mock-core-configs/etc/mock/alma+epel-8-aarch64.cfg
@@ -1,0 +1,6 @@
+include('templates/almalinux-8.tpl')
+include('templates/epel-8.tpl')
+
+config_opts['root'] = 'alma+epel-8-aarch64'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
This is split out of #803.

Conan-Kudo:
The ppc64le config is missing because AlmaLinux hasn't published it yet. @jaboutboul previously told me it should be available fairly soon, which is pretty much the blocker for #803.

mock-core-configs/etc/mock/almaepel-8-aarch64.cfg
```diff
@@ -0,0 +1,6 @@
+include('templates/almalinux-8.tpl')
+include('templates/epel-8.tpl')
+
+config_opts['root'] = 'almaepel-8-aarch64'
```
carlwgeorge, line 4: Can we go ahead an introduce these as `alma-epel-8-*`?

carlwgeorge (changes requested):

mock-core-configs/etc/mock/almaepel-8-aarch64.cfg
Conan-Kudo: That's going to mess with the parsing for all the stuff that auto-select configs (copr plugin, fedpkg, etc.)

Conan-Kudo: I've called it `alma+epel-8-*` instead, but that's now done.

xsuchy (approved):

PR #806, opened 2021-11-23
Standardize oraclelinux+epel configs

carlwgeorge:
* Rename files from `oraclelinux-epel-*` to `oraclelinux+epel-*`
* Use corresponding root name in config_opts
* Use the real EPEL

### DIFF
--- a/mock-core-configs/etc/mock/oraclelinux+epel-7-aarch64.cfg
+++ b/mock-core-configs/etc/mock/oraclelinux+epel-7-aarch64.cfg
@@ -1,6 +1,6 @@
 include('templates/oraclelinux-7.tpl')
-include('templates/oraclelinux-epel-7.tpl')
+include('templates/epel-7.tpl')
 
-config_opts['root'] = 'oraclelinux-7-aarch64'
+config_opts['root'] = 'oraclelinux+epel-7-aarch64'
 config_opts['target_arch'] = 'aarch64'
 config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/oraclelinux+epel-7-x86_64.cfg
+++ b/mock-core-configs/etc/mock/oraclelinux+epel-7-x86_64.cfg
@@ -1,6 +1,6 @@
 include('templates/oraclelinux-7.tpl')
-include('templates/oraclelinux-epel-7.tpl')
+include('templates/epel-7.tpl')
 
-config_opts['root'] = 'oraclelinux-7-x86_64'
+config_opts['root'] = 'oraclelinux+epel-7-x86_64'
 config_opts['target_arch'] = 'x86_64'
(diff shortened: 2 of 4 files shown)


Conan-Kudo (approved):

carlwgeorge:
To follow up on the last point in more detail, Oracle's EPEL is not EPEL.  Currently there is significant divergence.

* epel7 vs Oracle's epel7 (x86_64):
	* 13673 package names in real epel7
	* 15642 package names in Oracle's epel7
	* 86 package names only in the real epel7
	* 2055 package names only in Oracle's epel7
* epel8 vs Oracle's epel8 (x86_64):
	* 8076 package names in real epel8
	* 9619 package names in Oracle's epel8
	* 330 package names only in the real epel8
	* 1873 package names only in Oracle's epel8

Conan-Kudo:
The more interesting question is why don't they contribute to mainline EPEL...?

carlwgeorge:
Agreed, it's an interesting question, but I'm not sure where to ask it to get a substantial answer.  Hopefully in the future we can encourage them to contribute more to the real EPEL.  But for now all there is to say is that Oracle's EPEL does not have parity with the real EPEL, so to avoid confusion for mock users we should use the real EPEL.  Which is what we'll be doing since this is merged, thanks.  :grin:

Conan-Kudo:
Well, I guess the best way to ask is to ask @Djelibeybi here...

@Djelibeybi, any chance you could take a stab at answering that?

Djelibeybi:
We have a policy of only distributing binaries built by Oracle, and we had customers request EPEL packages, so we started building it. The package discrepencies may be because we ship source and binaries in the same repo. I'll need to look into it further, because I thought we had (finally) achieved partity (excluding the EPEL8 Modular stuff).

carlwgeorge:
Filtering out source rpms does remove some of the differences, but there are still significant differences.

epel7:
13673 package names in the real epel7
14806 package names in Oracle's epel7
86 package names only in the real epel7
1219 package names only in Oracle's epel7
13587 package names in both

epel8:
8082 package names in the real epel8
8544 package names in Oracle's epel8
337 package names only in the real epel8
799 package names only in Oracle's epel8
7745 package names in both

If you're curious I'm generating these numbers using [this script](https://gist.github.com/carlwgeorge/50ed939bf615d1bc22d617336cc0f47b).

There is nothing wrong with Oracle doing a rebuild of EPEL packages for customers that require that, it's just confusing to call that EPEL when it 1) doesn't have all the packages in EPEL and 2) has additional packages that have not been contributed to EPEL.  In any case mock will now use the real EPEL for the `oraclelinux+epel-*` chroots.

Djelibeybi:
Fair enough and no objection from me. I'll also point our build folks at your script to see if we can determine why there's such a large discrepency.

PR #809, opened 2021-11-25
2026571 - expand dnf_vars

xsuchy:
Because yum (we still support EL7) does not understand it.

Resolves: BZ#2026571

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-8.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-8.tpl
@@ -36,118 +36,118 @@ skip_if_unavailable=False
 
 [baseos]
 name=CentOS Stream $releasever - BaseOS
-mirrorlist=http://mirrorlist.centos.org/?release=$stream&arch=$basearch&repo=BaseOS&infra=$infra
-#baseurl=http://mirror.centos.org/$contentdir/$stream/BaseOS/$basearch/os/
+mirrorlist=http://mirrorlist.centos.org/?release=$releasever-stream&arch=$basearch&repo=BaseOS&infra=$infra
+#baseurl=http://mirror.centos.org/centos/$releasever-stream/BaseOS/$basearch/os/
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 gpgcheck=1
 skip_if_unavailable=False
 
 [appstream]
 name=CentOS Stream $releasever - AppStream
-mirrorlist=http://mirrorlist.centos.org/?release=$stream&arch=$basearch&repo=AppStream&infra=$infra
-#baseurl=http://mirror.centos.org/$contentdir/$stream/AppStream/$basearch/os/
+mirrorlist=http://mirrorlist.centos.org/?release=$releasever-stream&arch=$basearch&repo=AppStream&infra=$infra
+#baseurl=http://mirror.centos.org/centos/$releasever-stream/AppStream/$basearch/os/
 gpgcheck=1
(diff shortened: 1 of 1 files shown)


praiskup:
+1, but I'd appreciate a not why "stream-9 doesn't need fix" :-) (and I'll give others a day for review before merge)

xsuchy:
Because c9s does not use `dnf_vars`. It uses pseudo expanded `$releasever-stream`

Conan-Kudo:
Because I just decided not to use the stupid `$stream` variable for the Mock config.

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
```diff
@@ -36,23 +36,23 @@ skip_if_unavailable=False
 
 [baseos]
 name=CentOS Stream $releasever - BaseOS
-mirrorlist=http://mirrorlist.centos.org/?release=$stream&arch=$basearch&repo=BaseOS&infra=$infra
-#baseurl=http://mirror.centos.org/$contentdir/$stream/BaseOS/$basearch/os/
+mirrorlist=http://mirrorlist.centos.org/?release=8-stream&arch=$basearch&repo=BaseOS&infra=$infra
```
Conan-Kudo, line 39: Instead of `8-stream`, can you use `$releasever-stream`?

Conan-Kudo (changes requested):

praiskup:
I meant "in the commit message" :-) I didn't check.  The reason I thought it is that we don't care, as EL9 is not installable by yum anyway...

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
xsuchy: I can, Done.

praiskup:
Trying `vim -d ./mock-core-configs/etc/mock/templates/centos-stream-9.tpl ./mock-core-configs/etc/mock/templates/centos-stream-8.tpl` ...
Do we actually need those `dnf_vars` defined for something, then?

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
```diff
@@ -36,118 +36,118 @@ skip_if_unavailable=False
 
 [baseos]
 name=CentOS Stream $releasever - BaseOS
-mirrorlist=http://mirrorlist.centos.org/?release=$stream&arch=$basearch&repo=BaseOS&infra=$infra
-#baseurl=http://mirror.centos.org/$contentdir/$stream/BaseOS/$basearch/os/
+mirrorlist=http://mirrorlist.centos.org/?release=$releasever-stream&arch=$basearch&repo=BaseOS&infra=$infra
+#baseurl=http://mirror.centos.org/centos/releasever-stream/BaseOS/$basearch/os/
```
Conan-Kudo, line 40: Missing `$`

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
xsuchy: nice catch. fixed.

xsuchy:
>  Do we actually need those `dnf_vars` defined for something, then?

No. And yes. The original c8s defines it. And it is available there. If users craft a custom config they may use it. It is definitely no harm to have it there.

praiskup:
+1 (the commit message could have a nicer format, but that's a nit)

Conan-Kudo (approved):

PR #812, opened 2021-11-29
add fedora buttons

xsuchy:


docs/index.md
```diff
@@ -10,6 +10,10 @@ Mock also offers a multi-package command (`--chain`), that can build chains of p
 
 Mock is capable of building SRPMs from source configuration management if the `mock-scm` package is present, then building the SRPM into RPMs. See `--scm-enable` in the documentation.
 
+![Fedora package](https://img.shields.io/fedora/v/mock/rawhide)
+
+![Fedora package](https://img.shields.io/fedora/v/mock-core-configs/rawhide)
```
Conan-Kudo, line 15: This looks confusing since the package names aren't in the badge titles.

mock/README.md
```diff
@@ -1,5 +1,10 @@
 # Mock
 
+![Fedora package](https://img.shields.io/fedora/v/mock/rawhide)
+
+![Fedora package](https://img.shields.io/fedora/v/mock-core-configs/rawhide)
```
Conan-Kudo, line 5: This looks confusing since the package names aren't in the badge titles.

Conan-Kudo (changes requested):

docs/index.md
xsuchy: I am confused because I do not understand what exactly you mean by badge titles. :)

praiskup: It would be better to have:
`mock: <the badge>`
`mock-core-configs`: <the badge>`
Because the badge itself doesn't claim what package it belongs to. Just "Fedora: v36.3".

xsuchy:
Better?

Conan-Kudo (approved):

Conan-Kudo (changes requested):
You might want to put the two badges next to each other instead of on newlines.

praiskup:
Looks good.  I don't know how to put two images next to each other... so it's not my requirement.

Conan-Kudo:
Just put the two badge Markdown things right next to each other with a space in between. That'll do it just fine.

xsuchy:
> You might want to put the two badges next to each other instead of on newlines.

Hmm. Markdown table (visible lines)?  Or pure HTML (without visible lines of the border)?

xsuchy:
> Just put the two badge Markdown things right next to each other with a space in between. That'll do it just fine.

Ahh. Reading comments via email is slow :) I will try this. Thank you for a tip.

xsuchy:
Updated. But that aligns vertically to the bottom. I would prefer top alignment thou :(

praiskup:
Ugly, but you could use [html](https://stackoverflow.com/questions/40878670/how-to-vertically-align-elements-inside-github-markdown) for that?

xsuchy:
When I used <td>[!img] it did not work. GH does not render markup inside of HTML tags. So I used  at the end.
Now it renders fine.

praiskup:
I still see this:
Screenshot_20211202_171149
But I think it is good enough, so I'm +1 anyway (if you don't want to iterate more).

xsuchy:
I see it top alligned. And when I click on "view file" so rendered using GH, it is on top as well.

praiskup:
Ok, thank you!

praiskup:
Seems it is eventually *not* top-aligned, but we can fix that later:
https://rpm-software-management.github.io/mock/

PR #813, opened 2021-12-04
Add epel9 repos to epel9 template

carlwgeorge:
Originally the plan was to launch epel9-next before epel9.  We (the EPEL Steering Committee) changed course and decided to launch them together. This change adds the repo definitions that were previously missing from the template file and renames the epel-next-9-* config files to match the new {base}+epel pattern.

https://communityblog.fedoraproject.org/epel-9-is-now-available/


Conan-Kudo:
Nobody agreed to the elimination of the `epel-*` or `epel-next-*` configs yet, as far as I know.

carlwgeorge:
Yes, multiple people at the last meeting agreed that was the best path forward was to remove the `epel-8-*` configs and add extra error output to mock.

- https://pagure.io/epel/issue/133#comment-765381
- https://meetbot.fedoraproject.org/fedora-meeting/2021-12-01/epel.2021-12-01-21.00.log.html
- `21:54:06 <tdawson> Huh ... did we just agree on this?`

Conan-Kudo:
Okay then, but we don't need to add any logic to Mock then. The lack of an `epel-*`/`epel-next-*` config will be sufficient to break everyone anyway. Only `fedpkg(1)` will need changes.

PR #814, opened 2021-12-04
Align all the EPEL configs with the EPEL Steering Committee decision

Conan-Kudo:
The EPEL Steering Committee agreed to eliminate EPEL configs for exclusively distro+repo configs.

Additionally, fix EPEL 9 to have split configs for EPEL 9 and EPEL Next 9.

PR #817, opened 2021-12-08
Give user alternative help for 'epel-8-*' configs

praiskup:
Per discussion in: https://pagure.io/epel/issue/133


praiskup:
The initial error output looks like:
```
$ mock -r epel-8-x86_64 --shell
ERROR: Could not find required config file: /etc/mock/epel-8-x86_64.cfg
ERROR: There are those alternatives:
ERROR: [1] /etc/mock/alma+epel-8-x86_64.cfg
ERROR:     Use like: mock -r alma+epel-8-x86_64 ...
ERROR:     Builds against AlmaLinux repositories, together with the official EPEL repositories.Project page: https://almalinux.org/
ERROR: [2] /etc/mock/centos+epel-8-x86_64.cfg
ERROR:     Use like: mock -r centos+epel-8-x86_64 ...
ERROR:     Builds against CentOS Linux repositories, together with the official EPEL repositories.
ERROR:     This will reach end-of-life in January 2021.
ERROR: [3] /etc/mock/rhel+epel-8-x86_64.cfg
ERROR:     Use like: mock -r rhel+epel-8-x86_64 ...
ERROR:     Builds against Red Hat Enterprise Linux, together with the official EPEL repositories.
ERROR:     This mimics what is done in the official EPEL build system, though you need a Red Hat subscription:
ERROR:     https://rpm-software-management.github.io/mock/Feature-rhelchroots
ERROR: Alternatively you can install a permanent symbolic link pointing to one of
ERROR: the above files on path ~/.config/mock/epel-8-x86_64.cfg, and re-try the same command
```

praiskup:
Trying to build the updated packages here:
https://copr.fedorainfracloud.org/coprs/praiskup/mock-no-epel-8/

Conan-Kudo:
Is there a way we could generalize this to provide help when someone tries to invoke a config that doesn't exist but might be similar to something we _do_ have?

mock-core-configs/etc/mock/chroot-aliases.cfg
```diff
@@ -0,0 +1,32 @@
+config_opts["no-config"]["epel-8"] = {
+    "/etc/mock/alma+epel-8": {
```
Conan-Kudo, line 2: Let's not hardcode paths here if we can help it.

Conan-Kudo (changes requested):

praiskup:
What kind of non-existent configs do you have in mind?  Note there's bash completion,
so it's quite unlikely to make typos.

mock-core-configs/etc/mock/chroot-aliases.cfg
praiskup: The /etc/mock is widely hardcoded, unfortunately.  But in this case, it is doable, I think.

Conan-Kudo:
I'm thinking of when we retire CentOS 8 configs and for directing people to Stream for both 8 and 9.

mock-core-configs/etc/mock/chroot-aliases.cfg
```diff
@@ -0,0 +1,32 @@
...
+        "description": [
+            "Builds against AlmaLinux repositories, "
+            "together with the official EPEL repositories.",
+            "Project page: https://almalinux.org/"
```
xsuchy, line 6: I would love to see here command that will resolve it. E.g.:

    ln -s alma+epel-8-{{arch}} /etc/mock/epel-8-{{arch}}

So users can copy'n'paste it directly

mock/py/mockbuild/config.py
```diff
@@ -313,6 +313,8 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['stderr_line_prefix'] = ""
     config_opts['additional_packages'] = None
 
+    config_opts["no-config"] = {}
```
xsuchy, line 316: Documenting this in site-defaults will be nice.

mock/py/mockbuild/config.py
```diff
@@ -687,9 +741,12 @@ def load_config(config_path, name, uidManager, version, pkg_python_dir):
             chroot_cfg_path = '%s/%s.cfg' % (config_path, name)
     config_opts['config_file'] = chroot_cfg_path
 
-    cfg = os.path.join(config_path, 'site-defaults.cfg')
-    do_update_config(log, config_opts, cfg, uidManager, name)
+    # load the global config files
+    for cfg_file in ["site-defaults.cfg", "chroot-aliases.cfg"]:
```
xsuchy, line 745: I wonder why we need new config for that? This can be in default config (when done in upstream) or in site-defaults or .config/mock.cfg when user wants define their own.

mock-core-configs/etc/mock/chroot-aliases.cfg
xsuchy: I would use the basename too. And in code concat it with `config_path`

mock/py/mockbuild/config.py
praiskup: Because site-defaults.cfg is empty upstream, and users put there their stuff.
If we add something there it will not be delivered to many of our users.

mock-core-configs/etc/mock/chroot-aliases.cfg
praiskup: For N alternatives, N-times `ln -s` command in the output?

mock/py/mockbuild/config.py
praiskup: Because "noreplace"

praiskup:
Shouldn't we use "/usr/lib" instead of "/usr/share"?

mock/py/mockbuild/config.py
praiskup: I think that it would eventually confuse users;  this is never to be configured by users (or at least very unlikely).

mock-core-configs/etc/mock/chroot-aliases.cfg
praiskup: Fixed.

mock/py/mockbuild/config.py
praiskup: I eventually moved this to /usr/lib/mock/site-defaults.cfg as we anyway plan to move
everything there in v3.0+.

Conan-Kudo: The config files are architecture independent (we ship them in a _noarch_ package, so `/usr/share` is much more sensible).

praiskup: I tried to mimic how this is done in rpm/systemd (note it is not %_libdir).

Conan-Kudo: The reasons why systemd used `/usr/lib` don't apply here, and rpm's usage of `/usr/lib` is an ugly accident.

praiskup: Citation needed!

Conan-Kudo: I'd have to dig up the old references, but the reason systemd uses `/usr/lib` instead of `/usr/share` was that Debian wasn't willing to put a top level `/share` directory for its immutable files, so `/lib` is where it went. RPM's usage of `/usr/lib` for vendor/static configuration is pretty much an accident as it was originally for helper binaries and scripts (which was supposed to to be in `/usr/libexec`, but back in 1995, nobody was sure of anything yet). The path got co-opted later for other things. One day, it's on my TODO to actually fix that mess in RPM, but it'll be a while.

praiskup: > One day, it's on my TODO to actually fix that mess in RPM, but it'll be a while.

It would be nice if this was tracked somewhere else than your personal TODO list,
because it's worth seeing the feedback from the community about this.

I tend to fall back to keep the stuff in /etc till we have some authoritative answer
on how to handle this scenario :-(

praiskup:
- moved /usr/lib to just /etc for now, but as `%config` not `%config(noreplace)`
- moved `no-config.<chroot>.<alternative>` to `no-config.<chroot>.alternatives.<alternative>` structure so we can e.g. add some other metadata for given `chroot`, e.g. "boilerplate" text or anything

mock/py/mockbuild/config.py
```diff
@@ -638,20 +641,85 @@ def update_config_from_file(config_opts, config_file, uid_manager):
...
+        log.error("%sEnable permanently by:", pfx)
+        homeconfig = os.path.join(os.path.expanduser('~'), '.config', 'mock')
+        if not os.path.exists(homeconfig):
+            log.error("%s$ mkdir -p %s", pfx, homeconfig)
+
+        log.error("%s$ ln -s %s %s/%s-%s.cfg", pfx, cfg_path,
+                  homeconfig, alias_name, arch)
+
+    if any_alternative:
+        pass
```
praiskup, line 690: This is redundant.

praiskup:
Trying to understand the review workflow ... I still see "Changes requested" here by @Conan-Kudo, but
that is the chunk which no longer exists and I marked the discussion as "resoleved".  Ideas?

Conan-Kudo:
> Trying to understand the review workflow ... I still see "Changes requested" here by @Conan-Kudo, but that is the chunk which no longer exists and I marked the discussion as "resoleved". Ideas?

I basically need to re-review, which I'm doing now.

Conan-Kudo (approved):

praiskup:
I'm curious if we really have to do this for `epel-7`, or if we can keep epel==centos at least there?

Conan-Kudo:
> I'm curious if we really have to do this for `epel-7`, or if we can keep epel==centos at least there?

We can leave it be for 7. I think we're going to want to do some kind of thing for epel-9 though.

praiskup:
But we need to re-add the symlinks, right?  After #814.

Conan-Kudo:
> But we need to re-add the symlinks, right? After #814.

Hmm, good question. I'd kind of hoped we'd have a more generic solution so that we wouldn't have to add logic or whatnot, but yeah, probably symlinks for the epel-7 ones makes sense. I'll make a PR for that.

Conan-Kudo:
I've added symlinks for EPEL 7 in #829

mock-core-configs/mock-core-configs.spec
```diff
@@ -89,7 +89,10 @@ cp -a etc/mock/eol/templates/*.tpl %{buildroot}%{_sysconfdir}/mock/eol/templates
 # generate files section with config - there is many of them
 echo "%defattr(0644, root, mock)" > %{name}.cfgs
 find %{buildroot}%{_sysconfdir}/mock -name "*.cfg" -o -name '*.tpl' \
+    | grep -v chroot-aliases \
```
sincerelysaucy, line 92: Never rely on unqualified "grep" statements, they can accidentally filter peculiar names like "epel-8-test-chroot-aliases.cfg", and hilarity could ensue. And *always* put a dangling slash on a directory name, in case someone used a symlink for /etc/mock for whatever reason.

      ind %{buildroot}%{_sysconfdir}/mock/ -name 'chroot-aliases -prune -o -type f -name "*.cfg" -print -o -print -name '*.tpl' -print \

praiskup: Oh, have you been bitten by this?  Note that we talk about `%{buildroot}` which is automatically cleaned by rpmbuild/mock when building Mock rpm. Such an unexpected file shouldn't exist at all.

Even if, this shouldn't cause any big headaches to mock package maintainers because the build log clearly says what is happening.  If you believe this is worth fixing, open a PR.

PR #822, opened 2021-12-13
config: Add CentOS Stream 9 + EPEL Next 9

### LINKED ISSUE #821: centos-stream+epel-9-next config is missing?

Looking into https://github.com/rpm-software-management/mock/tree/main/mock-core-configs/etc/mock it seems that the `centos-stream+epel-9-next` configs were renamed to `centos-stream+epel-9`, but there should be both of them present. One with EPEL 9 and the other one with EPEL 9 Next. Also, the EPEL 9 Next configs should probably contain both Next and "no-Next" EPEL repos.

Conan-Kudo:
This allows people to build against EPEL Next 9.

Fixes #821.

### DIFF
--- a/mock-core-configs/etc/mock/centos-stream+epel-next-9-aarch64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-next-9-aarch64.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel-next"
+include('templates/centos-stream-9.tpl')
+include('templates/epel-9.tpl')
+include('templates/epel-next-9.tpl')
+
+config_opts['root'] = 'centos-stream+epel-next-9-aarch64'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/centos-stream+epel-next-9-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-next-9-ppc64le.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel-next"
+include('templates/centos-stream-9.tpl')
+include('templates/epel-9.tpl')
+include('templates/epel-next-9.tpl')
+
+config_opts['root'] = 'centos-stream+epel-next-9-ppc64le'
+config_opts['target_arch'] = 'ppc64le'
+config_opts['legal_host_arches'] = ('ppc64le',)
--- a/mock-core-configs/etc/mock/centos-stream+epel-next-9-s390x.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-next-9-s390x.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel-next"
+include('templates/centos-stream-9.tpl')
+include('templates/epel-9.tpl')
+include('templates/epel-next-9.tpl')
+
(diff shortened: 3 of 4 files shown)


Conan-Kudo:
cc: @hroncok

mock-core-configs/etc/mock/centos-stream+epel-next-9-aarch64.cfg
```diff
@@ -0,0 +1,7 @@
+include('templates/centos-stream-9.tpl')
+include('templates/epel-9.tpl')
+include('templates/epel-next-9.tpl')
```
hroncok, line 3: All 2 EPEL templates have a local repo. Does the latter win, or how does that work?

Conan-Kudo: That's... a good question. I'm not sure what happens here, but in any case, we need to dedupe these references. We have _three_ `local` references across these templates. And that's true for both 8 and 9.

I'm tempted to rename the EPEL ones to `epel-local` and `epel-next-local`. We'd need to document this change somewhere so people know how to use it, though.

Conan-Kudo: That's... a good question. I'm not sure what happens here, but in any case, we need to dedupe these references. We have _three_ `local` references across these templates. And that's true for both 8 and 9.

I'm tempted to rename the EPEL ones to `epel-local` and `epel-next-local`. We'd need to document this change somewhere so people know how to use it, though.

@carlwgeorge, what do you think?

praiskup: Perhaps rename them to `local-epel` and `local-epel-next`, so users can just do `--enablerepo 'local*'`.

Conan-Kudo: Oh, that makes sense! I'll make a separate PR for that.

Conan-Kudo: Done in #824

praiskup:
Do we want to have this in the very next mock-core-configs release, or can this wait another 14+ days?

Conan-Kudo:
We need this and #824 for the next core-configs release. Right now building stuff for EPEL 9 is broken. :(

praiskup:
I mean, can the 'epel-next-9' thing be delayed?  I agree that we want to have epel-9 working.

Conan-Kudo:
Maybe? I don't think so, though. People are building for EPEL Next 9: https://bodhi.fedoraproject.org/releases/EPEL-9N

hroncok:
> People are building for EPEL Next 9

Those are [mostly mistakes](https://pagure.io/releng/issue/10463) or [builds that happened before the EPEL 9 announcement](https://pagure.io/releng/issue/10456) though. Currently both EPEL 9 and EPEL 9 Next build against c9s, so there is no legitimate reason to build for EPEL 9 Next.

praiskup:
... or do we want to delay this when probably nobody wants this, yet?

Conan-Kudo:
I'm syncing it now, one sec

Conan-Kudo:
@praiskup Done!

praiskup:
LGTM, thanks.

praiskup:
OK, merging soon as well, so it can be tested before the release.  Feel free to send follow-up patches.

PR #824, opened 2021-12-13
config: Add EPEL suffixes to 'local' repo in EPEL configs

Conan-Kudo:
This ensures they are unique regardless of which base RHELish
distribution EPEL is combined with.

### DIFF
--- a/mock-core-configs/etc/mock/templates/epel-9.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-9.tpl
@@ -50,8 +50,8 @@ gpgcheck=1
 gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-$releasever
 skip_if_unavailable=False
 
-[local]
-name=local
+[local-epel]
+name=local-epel
 baseurl=https://kojipkgs.fedoraproject.org/repos/epel$releasever-build/latest/$basearch/
 cost=2000
 enabled=0
--- a/mock-core-configs/etc/mock/templates/epel-8.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-8.tpl
@@ -15,13 +15,6 @@ enabled=0
 mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=testing-epel8&arch=$basearch
 skip_if_unavailable=False
 
-[local]
-name=local
-baseurl=https://kojipkgs.fedoraproject.org/repos/epel8-build/latest/$basearch/
-cost=2000
-enabled=0
-skip_if_unavailable=False
-
 [epel-debuginfo]
 name=Extra Packages for Enterprise Linux $releasever - $basearch - Debug
 mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=epel-debug-8&arch=$basearch
@@ -49,6 +42,13 @@ skip_if_unavailable=False
 [epel-modular-source]
 name=Extra Packages for Enterprise Linux Modular $releasever - $basearch - Source
 mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=epel-modular-source-8&arch=$basearch
+enabled=0
+skip_if_unavailable=False
+
+[local-epel]
+name=local-epel
+baseurl=https://kojipkgs.fedoraproject.org/repos/epel8-build/latest/$basearch/
+cost=2000
 enabled=0
 skip_if_unavailable=False
 """
--- a/mock-core-configs/etc/mock/templates/epel-next-8.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-next-8.tpl
@@ -48,8 +48,8 @@ gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-8
 gpgcheck=1
 skip_if_unavailable=False
 
-[local]
-name=local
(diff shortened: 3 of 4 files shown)


Conan-Kudo:
cc: @hroncok @carlwgeorge @praiskup

praiskup:
Is it really OK to mixup epel-next with epel repositories in one build?  As I understand it, EPEL Next is not a complement to EPEL?

Conan-Kudo:
EPEL Next is a layer on top of EPEL. You need EPEL and EPEL Next together to create the full build and runtime environment.

praiskup:
So, does that apply to the Koji's "local" directories, too?  I thought that at least the "epel-next-local" is self-standing.

hroncok:
> So, does that apply to the Koji's "local" directories, too? I thought that at least the "epel-next-local" is self-standing.

Honestly, I think that epel-next-local is not in a working state, I've asked 2 questions at https://lists.fedoraproject.org/archives/list/epel-devel@lists.fedoraproject.org/thread/WQGFOV67QKUTN6V345GCMSSSBGH3M4AA/

mock-core-configs/etc/mock/templates/epel-9.tpl
```diff
@@ -50,8 +50,8 @@ gpgcheck=1
 gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-$releasever
 skip_if_unavailable=False
 
-[local]
-name=local
+[local-epel]
+name=local-epel
```
praiskup, line 54: Would you mind keeping local here?  It would minimize changes...

IMO we can change only the `local-epel-next` below (or even say `local-next`).

Conan-Kudo: It conflicts with `local` from CS9 base, so no, we can't.

praiskup: Ah, right.

praiskup:
I'm still a bit worried about this.  Users are used to just do `--enablerepo=local`
quite often, and the question is what they expect..

If they do `mock -r alma+epel-8-x86_64 --enablerepo=local` they definitely want to
enable "epel local" repo.  Ditto if they have `epel-8` symlink and they do
`-r epel-8-x86_64`.
Still, if they do `mock -r centos-stream-9-x86_64 --enablerepo=local`, they want
c9s local repo.
If they do `-r centos-stream+epel-9`, they likely want EPEL's local IMO
If they do `-r centos-stream+epel-next-8` they want EPEL Next's local.

I still don't think that `local-epel-next` should ever be enabled at the same time
with `local-epel`.  That's something which IMO never happens in Koji.

What about a Jinja2 trick, in `centos-stream-9` template the `local` could have this:
     `[local{% if epel_included %}-centos-stream]`
Then, `local` always means Koji's EPEL local?

Conan-Kudo:
I think we would still want to offer access to the CS9 local repo so that people can see whether the dependency exists there.

As for the `local-epel` + `local-epel-next` thing, I'm pretty sure epel9-next inherits from epel9: https://koji.fedoraproject.org/koji/taginfo?tagID=45736

hroncok:
> As for the `local-epel` + `local-epel-next` thing, I'm pretty sure epel-next inherits from epel: https://koji.fedoraproject.org/koji/taginfo?tagID=45736

Which will lead to bad results once epel9-next has older builds than epel9.

Conan-Kudo:
> > As for the `local-epel` + `local-epel-next` thing, I'm pretty sure epel-next inherits from epel: https://koji.fedoraproject.org/koji/taginfo?tagID=45736
> 
> Which will lead to bad results once epel9-next has older builds than epel9.

I believe @tdawson is working on a process to untag builds from epel-next once they've been superseded by epel builds, but yeah, this can be a problem.

Conan-Kudo:
Note that _none_ of the `local` repos are enabled by default, so you have to enable them to get their content. This just de-conflicts their IDs.

praiskup:
Yes, I get that.. but the thing is that `--enablerepo local` is a pretty known/common option.
It would be nice if we could have this without a breakage, if possible.

Conan-Kudo:
> Yes, I get that.. but the thing is that `--enablerepo local` is a pretty known/common option. It would be nice if we could have this without a breakage, if possible.

I'm not sure that's possible, unless we create a new custom option and wire that in to all the mock configs so that the correct knobs are flipped...?

praiskup:
Hmm, I would have to try ... but I thought that flipping that knob explicitly is needed only in affected chroots (those that include multiple local repos).

Conan-Kudo:
Well, if we're designing a custom knob, we might as well make it so it works for both the singular local and multiple local case.

praiskup:
I am not sure I follow.  Are you OK to experiment with this?

Conan-Kudo:
Personally, I don't think we need a knob, so I'd just be happy updating docs to indicate `epel-local` and such...

praiskup:
Then please give me a chance to take a look at this, hopefully today.

hroncok:
What a discussion I've started with a simple question :)

Conan-Kudo:
> Then please give me a chance to take a look at this, hopefully today.

Sure, go for it. 👍🏾

praiskup:
Proposed in #830.

Conan-Kudo:
Closing in favor of #830

PR #807, opened 2021-11-24
remove failovermethod=priority for dnf repos

istiak101:
Removed failovermethod=priority as this is removed from dnf lately.

PR #825, opened 2021-12-14
config: keep failovermethod=priority for a while

praiskup:
We still support EL7 in the release process.  We can remove this once we
have 'mock-2' branch separated from 'main' (for the purpose of EL7
mock/mock-core-configs maintenance).

This causes ugly warnings on modern fedora systems, but these are just
warnings which are going to disappear soon:
https://github.com/rpm-software-management/libdnf/pull/1276

This reverts commit d4b504c1c1d026f99f9e84852ec7543ce615d1bb.
Relates: #807


Conan-Kudo (approved):
They don't really do anything meaningful, but okay, I guess...

PR #830, opened 2021-12-15
Resolve the multiple "local" repo problems

praiskup:
Deterministically resolve which of the "local" repositories is
the primary one:

- 'centos-stream+epel' => local is EPEL buildroot
- 'centos-stream+epel-next' => local for EPEL Next buildroot
- 'centos-stream' => local is just CentOS Stream buildroot

The only requirement is to specify the 'koji_primary_repo' variable to
'epel', 'epel-next' or 'centos-stream'.

Users may still opt-in to use '--enablerepo "local*"' if they really
want to enable all the Koji buildroots.

### DIFF
--- a/mock-core-configs/etc/mock/templates/epel-8.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-8.tpl
@@ -17,17 +17,6 @@ mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=testing-epel8&arch=$
 failovermethod=priority
 skip_if_unavailable=False
 
-{% if koji_primary_repo != None and koji_primary_repo != "epel" %}
-[local-epel]
-{% else %}
-[local]
-{% endif %}
-name=local
-baseurl=https://kojipkgs.fedoraproject.org/repos/epel8-build/latest/$basearch/
-cost=2000
-enabled=0
-skip_if_unavailable=False
-
 [epel-debuginfo]
 name=Extra Packages for Enterprise Linux $releasever - $basearch - Debug
 mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=epel-debug-8&arch=$basearch
@@ -60,6 +49,17 @@ skip_if_unavailable=False
 name=Extra Packages for Enterprise Linux Modular $releasever - $basearch - Source
 mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=epel-modular-source-8&arch=$basearch
 failovermethod=priority
+enabled=0
+skip_if_unavailable=False
+
+{% if koji_primary_repo != None and koji_primary_repo != "epel" %}
+[local-epel]
+{% else %}
+[local]
+{% endif %}
+name=Extra Packages for Enterprise Linux $releasever - Koji Local - BUILDROOT ONLY!
+baseurl=https://kojipkgs.fedoraproject.org/repos/epel$releasever-build/latest/$basearch/
+cost=2000
 enabled=0
 skip_if_unavailable=False
 """
--- a/mock-core-configs/etc/mock/templates/epel-8.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-8.tpl
@@ -17,7 +17,11 @@ mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=testing-epel8&arch=$
 failovermethod=priority
 skip_if_unavailable=False
 
+{% if koji_primary_repo != None and koji_primary_repo != "epel" %}
+[local-epel]
+{% else %}
 [local]
+{% endif %}
 name=local
 baseurl=https://kojipkgs.fedoraproject.org/repos/epel8-build/latest/$basearch/
 cost=2000
--- a/mock-core-configs/etc/mock/templates/rocky-8.tpl
+++ b/mock-core-configs/etc/mock/templates/rocky-8.tpl
@@ -84,7 +84,7 @@ enabled=0
 gpgkey=file:///usr/share/distribution-gpg-keys/rocky/RPM-GPG-KEY-rockyofficial
 
 [devel]
-name=Rocky Linux $releasever - Devel WARNING! FOR BUILDROOT AND KOJI USE
+name=Rocky Linux $releasever - Devel (WARNING: UNSUPPORTED - FOR BUILDROOT USE ONLY!)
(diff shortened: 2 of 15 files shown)


Conan-Kudo (approved):
I like this solution, very clever!

mock-core-configs/etc/mock/templates/epel-8.tpl
```diff
@@ -17,8 +17,12 @@ mirrorlist=http://mirrors.fedoraproject.org/mirrorlist?repo=testing-epel8&arch=$
...
+{% if koji_primary_repo != None and koji_primary_repo != "epel" %}
+[local-epel]
+{% else %}
 [local]
-name=local
+{% endif %}
+name=Extra Packages for Enterprise Linux $releasever - Koji Local WARNING! FOR BUILDROOT ONLY!
 baseurl=https://kojipkgs.fedoraproject.org/repos/epel8-build/latest/$basearch/
 cost=2000
 enabled=0
```
Conan-Kudo, line 28: Can you move this down to the bottom like how epel-9.tpl has it?

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/epel-8.tpl
praiskup: Done, in separate commit.

mock-core-configs/etc/mock/templates/rocky-8.tpl
```diff
@@ -133,7 +133,7 @@ enabled=0
 gpgkey=file:///usr/share/distribution-gpg-keys/rocky/RPM-GPG-KEY-rockyofficial
 
 [devel-debug]
-name=Rocky Linux $releasever - Devel WARNING! FOR BUILDROOT AND KOJI USE
+name=Rocky Linux $releasever - Devel - BUILDROOT ONLY!
 mirrorlist=https://mirrors.rockylinux.org/mirrorlist?arch=$basearch&repo=Devel-$releasever-debug
 #baseurl=http://dl.rockylinux.org/pub/rocky/$releasever/Devel/$basearch/debug/tree
 gpgcheck=1
```
Conan-Kudo, line 139: Why does this exist here? We can't use debuginfo for anything useful and only the `devel` repo has one? @nazunalika?

mock-core-configs/etc/mock/templates/almalinux-8.tpl
```diff
@@ -65,7 +65,7 @@ gpgkey=file:///usr/share/distribution-gpg-keys/alma/RPM-GPG-KEY-AlmaLinux
 fastestmirror=1
 
 [devel]
-name=AlmaLinux $releasever - Devel (WARNING: UNSUPPORTED - FOR BUILDROOT USE ONLY!)
```
Conan-Kudo, line 68: I think we should synchronize with `WARNING:` and note it's all unsupported.

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/rocky-8.tpl
praiskup: Honestly, speak fast here (today) :-) or delay this to the next release!

Conan-Kudo: I'd rather have this removed now and if we want it back, let's have @nazunalika add them all back correctly later.

mock-core-configs/etc/mock/templates/almalinux-8.tpl
praiskup: Can you spell the exact wording you prefer?  I think it is OK now, but I'll accept your variant.  I just don't want to spend too much time on this PR.

Conan-Kudo: `(WARNING: UNSUPPORTED - FOR BUILDROOT USE ONLY!)`

Conan-Kudo: For EPEL, the `local` repos are mainly pre-compose (direct koji access), so the unsupported/warning doesn't apply there.

praiskup: Hmm, strictly speaking, is it different?  Those packages are neither supported ..

praiskup: I mean, it contains all sort of buildoverrides, and are not signed, and anything may disappear.

Conan-Kudo: Fair point, actually. But generally, the expectation is that Koji stuff eventually ships out for EPEL.

praiskup: Ack, and "local" repos are opt-in.

Conan-Kudo: Yup!

praiskup: Ouch, the Devel repo is opt-in intentionally?

mock-core-configs/etc/mock/templates/rocky-8.tpl
praiskup: It is enablerepo=0, so I just dropped the "BUILDROOT ONLY" warning which is nowhere else in the debuginfo repos.

mock-core-configs/etc/mock/templates/almalinux-8.tpl
praiskup: Ok, according to the latest info those repos are mostly empty anyway.

Conan-Kudo: For CentOS, yes, because that's all moved to `local`, but for Alma/Rocky, they're full of everything that's normally in `local`.

praiskup: This is off-topic, but I'm curious why ...?  Devel (mirrored) is orthogonal thing to 'local'?

praiskup:
Updated.  I fixed one more C&P error (s/Source/Debuginfo)

Conan-Kudo (approved):

praiskup:
This is mostly blocking release testing, and the pull request #822, if I get it correctly.  So merging now, thank you!

PR #834, opened 2021-12-16
mock: don't print coverage stats for RPM builds

praiskup:
This is breaking build on F34 (#2033331), but it also doesn't make any
sense to run coverage analysis for every RPM build.  By default, if run
locally, the ./run-tests.sh script still performs coverage analysis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,6 @@ matrix:
 
 install:
   - "pip install -r mock/requirements.txt"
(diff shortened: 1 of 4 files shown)


praiskup:
Of course "unrecognized arguments: --cov-report --cov".

praiskup:
Fixed.

praiskup:
This is breaking other PRs, so merging early (it is a detail anyway).

Issue #811, opened 2021-11-26
`Curl error (6): Couldn't resolve host name` within Fedora 35 mock buidlroot on CentOS 8 Linux host


anoopcs9:
### Short description of the problem
On a CentOS 8 Linux host, Fedora 35 mock buildoort fails to setup during package repository sync with following error:
```
fedora                                                                                                                         0.0  B/s |   0  B     00:00    
Errors during downloading metadata for repository 'fedora':
  - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-35&arch=x86_64 [getaddrinfo() thread failed to start]
Error: Failed to download metadata for repo 'fedora': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-35&arch=x86_64 [getaddrinfo() thread failed to start]
ERROR: Command failed: 
 # /usr/bin/systemd-nspawn -q -M 406e5be982ab4cfa8455da19e7774eda -D /var/lib/mock/fedora-35-x86_64-bootstrap/root -a --capability=cap_ipc_lock --bind=/tmp/mock-resolv.wu_fyfix:/etc/resolv.conf --setenv=TERM=vt100 --setenv=SHELL=/bin/bash --setenv=HOME=/var/lib/mock/fedora-35-x86_64/root/installation-homedir --setenv=HOSTNAME=mock --setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin --setenv=PROMPT_COMMAND=printf "\033]0;<mock-chroot>\007" --setenv=PS1=<mock-chroot> \s-\v\$  --setenv=LANG=C.UTF-8 --setenv=LC_MESSAGES=C.UTF-8 --resolv-conf=off /usr/bin/dnf --installroot /var/lib/mock/fedora-35-x86_64/root/ --releasever 35 --setopt=deltarpm=False --allowerasing --disableplugin=local --disableplugin=spacewalk --disableplugin=versionlock install @buildsys-build --setopt=tsflags=nocontexts
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
fedora                                                                                                                         0.0  B/s |   0  B     00:00    
Errors during downloading metadata for repository 'fedora':
[...]

PR #835, opened 2021-12-16
nspawn: disable system call filtering

praiskup:
It happened several times before that the software inside container hit
the systemd-nspawn's seccomp filters.  It is proven to cause troubles,
and the filter updates take some time.

It is not the purpose of Mock to provide full isolation (see "Scope"
in https://rpm-software-management.github.io/mock/ and the notes about
isolation), and thus it looks it should be OK to disable this feature.

Fixes: #811


Conan-Kudo (approved):

praiskup:
One more update here :-( I realized we were setting the systemd variables on two places.

praiskup:
Seems to pass tests...

PR #844, opened 2021-12-24
core-configs: Fix CentOS templates

dpward:
This fixes differences between the CentOS templates in mock, and the actual repository configurations for CentOS, which results in build issues depending on the repositories and/or architectures.

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-7.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-7.tpl
@@ -51,7 +51,6 @@ skip_if_unavailable=False
 
 [updates]
 name=CentOS-$releasever - Updates
-enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=updates&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/updates/$basearch/
 failovermethod=priority
@@ -85,9 +84,11 @@ enabled=0
 name=CentOS-$releasever - Plus
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=centosplus&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/centosplus/$basearch/
+failovermethod=priority
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
        {{- centos_7_arch_gpg_key }}
 gpgcheck=1
+skip_if_unavailable=False
 enabled=0
 
 {% if target_arch == 'x86_64' %}
@@ -95,6 +96,7 @@ enabled=0
 name=CentOS-$releasever - SCLo sclo
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=sclo-sclo&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/sclo/$basearch/sclo/
+failovermethod=priority
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-SIG-SCLo
 gpgcheck=1
 includepkgs=devtoolset*
@@ -106,6 +108,7 @@ skip_if_unavailable=False
 name=CentOS-$releasever - SCLo rh
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=sclo-rh&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/sclo/$basearch/rh/
+failovermethod=priority
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-SIG-SCLo
 gpgcheck=1
 includepkgs=devtoolset*
--- a/mock-core-configs/etc/mock/templates/centos-7.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-7.tpl
@@ -30,12 +30,13 @@ best=1
 protected_packages=
 user_agent={{ user_agent }}
 
-{% set centos_7_gpg_keys = 'file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7' %}
+{% set centos_7_arch_gpg_key -%}
 {% if target_arch in ['ppc64le', 'ppc64'] %}
-{%   set centos_7_gpg_keys = centos_7_gpg_keys + ',file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-SIG-AltArch-7-' + target_arch %}
-{% elif target_arch in ['aarch64'] %}
-{%   set centos_7_gpg_keys = centos_7_gpg_keys + ',file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7-aarch64' %}
-{% endif %}
+       file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-SIG-AltArch-7-{{ target_arch }}
+{%- elif target_arch == 'aarch64' %}
+       file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7-aarch64
+{%- endif -%}
+{% endset -%}
 
 # repos
 [base]
@@ -43,7 +44,8 @@ name=CentOS-$releasever - Base
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=os&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/os/$basearch/
 failovermethod=priority
-gpgkey={{ centos_7_gpg_keys }}
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
+       {{- centos_7_arch_gpg_key }}
 gpgcheck=1
 skip_if_unavailable=False
 
@@ -53,7 +55,8 @@ enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=updates&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/updates/$basearch/
 failovermethod=priority
-gpgkey={{ centos_7_gpg_keys }}
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
+       {{- centos_7_arch_gpg_key }}
 gpgcheck=1
 skip_if_unavailable=False
 
@@ -62,7 +65,8 @@ name=CentOS-$releasever - Extras
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=extras&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/extras/$basearch/
 failovermethod=priority
-gpgkey={{ centos_7_gpg_keys }}
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
+       {{- centos_7_arch_gpg_key }}
 gpgcheck=1
 skip_if_unavailable=False
 
@@ -71,7 +75,8 @@ name=CentOS-$releasever - fasttrack
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=fasttrack&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/fasttrack/$basearch/
 failovermethod=priority
-gpgkey={{ centos_7_gpg_keys }}
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
+       {{- centos_7_arch_gpg_key }}
 gpgcheck=1
 skip_if_unavailable=False
 enabled=0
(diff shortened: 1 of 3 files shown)


mock-core-configs/etc/mock/templates/centos-stream-8.tpl
```diff
@@ -27,132 +27,128 @@ protected_packages=
...
+{% set contentdir = dnf_vars['contentdir'] -%}
+
 {% if koji_primary_repo != None and koji_primary_repo != "centos-stream" %}
 [local-centos-stream]
 {% else %}
 [local]
 {% endif %}
 name=CentOS Stream $releasever - Koji Local - BUILDROOT ONLY!
-baseurl=https://koji.mbox.centos.org/kojifiles/repos/dist-c{{ releasever }}-stream-build/latest/$basearch/
+baseurl=https://koji.mbox.centos.org/kojifiles/repos/dist-c{{ stream }}-build/latest/$basearch/
```
Conan-Kudo, line 39: Please don't switch away from `$releasever-stream`

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
dpward: @Conan-Kudo The goal here is to keep variables in the same places that [they appear upstream](https://git.centos.org/rpms/centos-repos/blob/11643f64f7c88715c5b0564158ff2a851b713b1e/f/SOURCES/CentOS-Stream-BaseOS.repo#_14). Is your comment specific to Koji?

Conan-Kudo: We deliberately deviate because extended variable support does not exist in yum. None of those extra variables you're adding will work on EL7. Additionally, there have been compatibility issues with extended vars on the host environment. That's why we avoid them. Being identical with the ones in the repos package is not required or in some cases very undesirable.

dpward: > We deliberately deviate because extended variable support does not exist in yum.
That's not correct. Yum supports extended variables — [even in RHEL 6](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_yum_variables). It is mock that doesn't handle them.

> None of those extra variables you're adding will work on EL7.
These are Jinja2 variables, not Yum variables, so yes they do work.

> Additionally, there have been compatibility issues with extended vars on the host environment. That's why we avoid them. Being identical with the ones in the repos package is not required or in some cases very undesirable.
Can you please elaborate on why they aren't desirable? I gave reasons why they are.


Please keep in mind the comments on #809:
> > Do we actually need those dnf_vars defined for something, then?
>
> No. And yes. The original c8s defines it. And it is available there. If users craft a custom config they may use it. It is definitely no harm to have it there.

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -91,7 +97,8 @@ skip_if_unavailable=False
 {% if target_arch in ['x86_64', 'ppc64le', 'aarch64'] %}
 [centos-sclo-rh]
 name=CentOS-$releasever - SCLo rh
-mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=sclo-rh
+mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=sclo-rh&infra=$infra
```
praiskup, line 135: What the `infra=$infra` means?

praiskup:
> which results in build issues depending on the repositories and/or architectures.

Can you please document what are the issues you fixed?  Error messages, etc.?

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -7,8 +7,6 @@ config_opts['releasever'] = '7'
 config_opts['bootstrap_image'] = 'quay.io/centos/centos:7'
 config_opts['package_manager'] = 'yum'
 
-config_opts['yum_install_command'] += "{% if target_arch in ['x86_64', 'ppc64le', 'aarch64'] %} --disablerepo=centos-sclo*{% endif %}"
-
```
Conan-Kudo, line 11: This also breaks EPEL 7 builds, since we do need this to ensure SCLo content is only used at the right stage for EPEL 7 builds.

Conan-Kudo (changes requested):

dpward:
> > which results in build issues depending on the repositories and/or architectures.
> 
> Can you please document what are the issues you fixed? Error messages, etc.?

The issues are mentioned in the commit log already, but to illustrate with only a few examples:

* Misspelled repository names
```
$ mock -r centos-7-x86_64 --enablerepo=fasttrack [...]
Error: Unknown repo: fasttrack
```
* Missing repositories
```
$ mock -r centos-stream-8-x86_64 --enablerepo=resilientstorage --init
Error: Unknown repo: resilientstorage
```
* Missing commented `baseurl=` settings
```
$ sudo sed -i -e '/^mirrorlist/s/^/#/' -e '/^#baseurl=/s/^#//' /etc/mock/templates/centos*-[78].tpl
$ mock -r centos-7-x86_64 --init
Error: cannot find a valid baseurl for repo: base
$ mock -r centos-8-x86_64 --init
Error: cannot find a valid baseurl for repo: baseos
$ mock -r centos-stream-8-x86_64 --init
Error: cannot find a valid baseurl for repo: baseos
```

Conan-Kudo:
Your changes are too invasive. Some of them (like the SCLo changes) will break setting up the build environment properly for EPEL. Things like repo ID names being wrong are easy fixes we could take, but a lot of the others are too much.

Can you please make separate PRs for the fixes so we can figure them out?

Also, we don't release `mock-core-configs` from the old mock branches, so there's no point in making changes there.

praiskup:
> Also, we don't release mock-core-configs from the old mock branches, so there's no point in making changes there.

I think that it will be complicated to release just from main, and keep compatibility with both mock 2 and 3.  At least
I thought we will release el7-only core configs from `mock-2` branch. :-)

dpward:
I will restructure the PRs, but kindly keep in mind that CentOS 7 is not EOL until 2024 and we need to continue supporting it as a build target. The fact is that it doesn't work properly in many cases right now and we need to fix those.

Conan-Kudo:
> > Also, we don't release mock-core-configs from the old mock branches, so there's no point in making changes there.
> 
> I think that it will be complicated to release just from main, and keep compatibility with both mock 2 and 3. At least I thought we will release el7-only core configs from `mock-2` branch. :-)

If we do that, we'll need to ensure the versioning doesn't conflict somehow. That said, I don't think there's much changing right now with the config file format. We haven't significantly changed them in several years...

mock-core-configs/etc/mock/templates/centos-7.tpl
dpward: I will adjust this, but can you please provide a reproducer? I have tested on multiple host/target combinations and I cannot see where this breaks.

mock-core-configs/etc/mock/templates/centos-7.tpl
dpward: It is part of the actual CentOS repository configurations. Mock already includes it for [CentOS 8](https://github.com/rpm-software-management/mock/blob/0fb40f3a3dc03864a6a23cdc5d2df8eaa3cf96b0/mock-core-configs/etc/mock/templates/centos-8.tpl#L31) and [CentOS Stream 8](https://github.com/rpm-software-management/mock/blob/0fb40f3a3dc03864a6a23cdc5d2df8eaa3cf96b0/mock-core-configs/etc/mock/templates/centos-stream-8.tpl#L43).

praiskup: Why do we need it here in mock?

dpward: The repository configurations in CentOS are normally provided in the packages centos-release, epel-release, centos-release-scl, etc. But when the buildroot is first initialized, these packages aren't installed inside of it, so there is a bootstrapping problem. That's why mock has to write these repository configurations at all, so that it can create the buildroot.

As mentioned above, sometimes an administrator has to modify the provided configs in the OS, for example to point to a local mirror. The same changes would need to be made in the mock template, and this is far easier to do when it matches what is provided in CentOS.

So, to ask your question the other way around: why does mock need to use anything different than what is provided by CentOS?

Including an additional repository for koji is certainly understandable. But can we point to a reason that the URLs need to be different for the CentOS-provided repos?

praiskup: > The only reason distro repo configs are needed in mock at all is because they don't exist in the chroot when it is first created. Otherwise, the .repo files would already be provided by (in this case) the centos-release* packages.

Even if the centos-relese with repo files is provided we still intentionally use configs from mock.  This is
mostly an innocent bug (we waste the buildroot storage with useless repo files).

> To ask your question the other way around: why does mock need to use anything other than the URLs provided by CentOS for the CentOS repos? Are there specific reasons?

Any change needs to have a reason, as it forces developers to ask/focus to useless things.  So if we don't need the change, I'm -1.  I already asked several times about this, and haven't got the answer.

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -1,11 +1,15 @@
 # This list is taken from 'epel-7-x86_64' @buildsys-build group, minus the
 # 'epel-*' specific stuff.
 config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
+config_opts['chroot_additional_packages'] = 'scl-utils-build'
```
praiskup, line 4: This package is not installed by default in Koji EPEL, so we can not add it here.

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -1,11 +1,15 @@
...
 
 config_opts['dist'] = 'el7'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '7'
 config_opts['bootstrap_image'] = 'quay.io/centos/centos:7'
 config_opts['package_manager'] = 'yum'
+config_opts['yum_vars'] = { 'contentdir': "{% if target_arch in ['x86_64'] -%} centos {%- else -%} altarch {%- endif %}",
+                            'infra': 'stock',
```
praiskup, line 11: Before we do this, we should have good arguments about why we do so.

praiskup:
I agree with Neal that this can be made simpler, can you split or drop the commented patches?

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -27,74 +31,113 @@ best=1
...
 gpgcheck=1
-skip_if_unavailable=False
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
+       {{- centos_7_arch_gpg_key }}
 
 [extras]
 name=CentOS-$releasever - Extras
-mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=extras
-#baseurl=http://mirror.centos.org/centos/$releasever/extras/$basearch/os/
-failovermethod=priority
```
praiskup, line 59: failovermethod=priority should stay enabled in EL7 configs

mock-core-configs/etc/mock/templates/centos-7.tpl
dpward: https://github.com/rpm-software-management/mock/pull/809#issuecomment-979231466

dpward:
> I agree with Neal that this can be made simpler, can you split or drop the commented patches?

My earlier comment seems to have gotten lost here.

The end result of these (adjusted) changes is to make the repository configurations match what is found upstream in CentOS. Mock should be able to handle those configurations as-is, for several reasons (mentioned above).

The current configurations in mock are out-of-sync with CentOS and do not work in a number of cases. (A further example is building packages that use Software Collections other than devtoolset.)

I have tested these (with #847) on several host/target combinations, as well as using `--no-bootstrap-chroot`, `--with-bootstrap-image`, etc. If there are specific examples of issues that these changes causes, could you please provide more details so I can reproduce?

mock-core-configs/etc/mock/templates/centos-7.tpl
dpward: Ok, should we use it in EL8 as well? Because it is missing from all repos there (except one).

mock-core-configs/etc/mock/templates/centos-7.tpl
dpward: Does Koji EPEL have the SCLo repositories (and enabled by default), as the Mock configs do?
The repo definitions in Mock alone aren't enough to build packages that use SCLo. I would suggest we consider dropping them if we aren't going to have the corresponding RPM macros. Or this could be moved to a separate config.

Conan-Kudo: Yes.

Conan-Kudo:
Fundamentally, I think the disagreement here is that making the configurations identical with `centos-repos` is useful for Mock's purpose. While at least having identical repo IDs makes sense, the underlying content of the configurations don't matter so much. And we have differences intentionally to support builder use-cases and minimize the weird things we need to do in Mock.

Some of the choices CentOS does (like using weird extended yum/dnf vars) are not necessary for Mock, and eventually I'd like to eliminate them in CentOS too.

mock-core-configs/etc/mock/templates/centos-7.tpl
praiskup: But scl-utils-build is not in the buildroot, note this: https://github.com/rpm-software-management/mock/blob/691d914f561ca164ad777bd456aa9ea8011d0650/mock-core-configs/etc/mock/templates/centos-7.tpl#L87

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -27,74 +31,133 @@ best=1
 protected_packages=
 user_agent={{ user_agent }}
 
-{% set centos_7_gpg_keys = 'file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7' %}
+{% set centos_7_arch_gpg_key -%}
```
praiskup, line 34: How does this work now?  Seems this template variable is never set.

praiskup:
I'm closing this for inactivity.  But feel free to reopen.

PR #845, opened 2021-12-24
core-configs: Fix CentOS templates (mock-2)

dpward:
This fixes differences between the CentOS templates in mock, and the actual repository configurations for CentOS, which results in build issues depending on the repositories and/or architectures.

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-7.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-7.tpl
@@ -51,7 +51,6 @@ skip_if_unavailable=False
 
 [updates]
 name=CentOS-$releasever - Updates
-enabled=1
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=updates&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/updates/$basearch/
 failovermethod=priority
@@ -85,9 +84,11 @@ enabled=0
 name=CentOS-$releasever - Plus
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=centosplus&infra=$infra
 #baseurl=http://mirror.centos.org/$contentdir/$releasever/centosplus/$basearch/
+failovermethod=priority
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
        {{- centos_7_arch_gpg_key }}
 gpgcheck=1
+skip_if_unavailable=False
 enabled=0
 
 {% if target_arch == 'x86_64' %}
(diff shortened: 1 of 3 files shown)


mock-core-configs/etc/mock/templates/centos-stream-8.tpl
```diff
@@ -27,132 +27,128 @@ protected_packages=
...
+{% set contentdir = dnf_vars['contentdir'] -%}
+
 {% if koji_primary_repo != None and koji_primary_repo != "centos-stream" %}
 [local-centos-stream]
 {% else %}
 [local]
 {% endif %}
 name=CentOS Stream $releasever - Koji Local - BUILDROOT ONLY!
-baseurl=https://koji.mbox.centos.org/kojifiles/repos/dist-c{{ releasever }}-stream-build/latest/$basearch/
+baseurl=https://koji.mbox.centos.org/kojifiles/repos/dist-c{{ stream }}-build/latest/$basearch/
```
Conan-Kudo, line 39: Please do not switch away from `$releasever-stream`

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-8.tpl
dpward: @Conan-Kudo The goal here is to keep variables in the same places that [they appear upstream](https://git.centos.org/rpms/centos-repos/blob/11643f64f7c88715c5b0564158ff2a851b713b1e/f/SOURCES/CentOS-Stream-BaseOS.repo#_14). Is your comment specific to Koji?

Conan-Kudo: We deliberately deviate because extended variable support does not exist in yum. None of those extra variables you're adding will work on EL7.

dpward: > We deliberately deviate because extended variable support does not exist in yum.

That's not correct. Yum supports extended variables — [even in RHEL 6](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_yum_variables). It is mock that doesn't handle them.

> None of those extra variables you're adding will work on EL7.

These are Jinja2 variables, not Yum variables, so yes they do work.

mock-core-configs/etc/mock/templates/centos-7.tpl
```diff
@@ -27,74 +31,113 @@ best=1
...
 gpgcheck=1
-skip_if_unavailable=False
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
+       {{- centos_7_arch_gpg_key }}
 
 [updates]
 name=CentOS-$releasever - Updates
-enabled=1
-mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=updates
-failovermethod=priority
```
praiskup, line 50: I don't want to drop the "failovermethod=priority" here in `mock-2` branch.

praiskup:
This is too invasive for mock-2 branch.  Let's do the cleanups v3 only.

praiskup:
That said, thank you for the PR anyway  -- we'll continue in #844

PR #847, opened 2022-01-02
Set extended variables for yum

dpward:
yum supports extended variables, even in RHEL 6. If either `yum_vars` or `dnf_vars` are set in the mock configuration, then set extended variables for the appropriate package manager. These are used in URLs for CentOS. (This change precedes #844.)

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -50,7 +50,8 @@ def nspawn_supported():
 @traceLog()
 def setup_default_config_opts(unprivUid, version, pkgpythondir):
     "sets up default configuration."
-    config_opts = TemplatedDictionary(alias_spec={'dnf.conf': ['yum.conf']})
+    config_opts = TemplatedDictionary(alias_spec={'dnf.conf': ['yum.conf'],
+                                                  'dnf_vars': ['yum_vars']})
     config_opts['config_paths'] = []
     config_opts['version'] = version
     config_opts['basedir'] = '/var/lib/mock'  # root name is automatically added to this
@@ -247,7 +248,6 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['target_arch'] = 'i386'
     config_opts['releasever'] = None
     config_opts['rpmbuild_arch'] = None  # <-- None means set automatically from target_arch
-    config_opts['dnf_vars'] = {}
     config_opts['yum_builddep_opts'] = []
     config_opts['yum_common_opts'] = []
     config_opts['update_before_build'] = True
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -179,6 +179,15 @@ class _PackageManager(object):
             return self.config['yum.conf']
 
     @traceLog()
+    def get_pkg_manager_vars(self):
+        if 'dnf_vars' in self.config:
+            return self.config['dnf_vars']
+        elif 'yum_vars' in self.config:
+            return self.config['yum_vars']
+        else:
+            return None
+
+    @traceLog()
     def execute(self, *args, **kwargs):
         self.plugins.call_hooks("preyum")
         pm_umount = False
@@ -360,7 +369,7 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
 
     def expand_url_vars(self, string):
         """
-        Expand DNF variables like $baseurl to proper values in string, and
+        Expand variables (like $basearch or $releasever) in a string, and
         return it.
         """
         expand = {
@@ -368,9 +377,9 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
             "releasever": self.config["releasever"] or '<undef>',
         }
 
-        if 'dnf_vars' in self.config:
-            for key in self.config['dnf_vars']:
-                expand[key] = self.config['dnf_vars'][key]
+        pkg_manager_vars = self.get_pkg_manager_vars()
+        if pkg_manager_vars:
+            expand.update(pkg_manager_vars)
 
         for key, value in expand.items():
             # replace both $key and ${key}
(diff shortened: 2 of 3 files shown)


mock/py/mockbuild/config.py
```diff
@@ -50,7 +50,8 @@ def nspawn_supported():
 @traceLog()
 def setup_default_config_opts(unprivUid, version, pkgpythondir):
     "sets up default configuration."
-    config_opts = TemplatedDictionary(alias_spec={'dnf.conf': ['yum.conf']})
+    config_opts = TemplatedDictionary(alias_spec={'dnf.conf': ['yum.conf'],
+                                                  'dnf_vars': ['yum_vars']})
```
praiskup, line 54: The movement being done now is that mock v3 doesn't have to be compatible with yum at all.
We should start removing yum-compat hacks, rather than add new ones.

dpward: We still need to be able to build for targets that use yum, even if the host does not.

praiskup: I personally don't think we need this rather risky change which fixes a cosmetic issue.

xsuchy:
This will not work on EL7 without bootstrap.

mock/py/mockbuild/config.py
dpward: This is not a cosmetic issue. Please read again:
https://github.com/rpm-software-management/mock/pull/809#issuecomment-979231466

Because the same thing applies to yum.

praiskup: The fact we have it expanded is a cosmetic issue.

dpward:
> This will not work on EL7 without bootstrap.

With these changes applied,

creating a centos+epel-7-x86_64 chroot on a CentOS 7 host:
```
$ mock --scrub=all
<...>
$ mock --no-bootstrap-chroot --init
<...>
Finish: run
```
(I also relocated /etc/yum/vars on the host to ensure it wasn't using that.) Here is [root.log](https://paste.centos.org/view/25e81b8a).

creating a centos+epel-7-x86_64 chroot on a Fedora 35 host:
```
$ mock -r centos+epel-7-x86_64 --scrub=all
<...>
$ mock -r centos+epel-7-x86_64 --no-bootstrap-chroot --init
<...>
Finish: run
```
Note that because DNF is used here, it writes out `yum_vars` as DNF variables instead:
```
$ ls /var/lib/mock/centos+epel-7-x86_64/root/etc/dnf/vars/
contentdir  infra
```
Here is [root.log](https://paste.centos.org/view/56705c0d).

mock/py/mockbuild/config.py
dpward: You're missing this part:
> If users craft a custom config they may use it.

mock/py/mockbuild/config.py
```diff
@@ -247,7 +248,6 @@ def setup_default_config_opts(unprivUid, version, pkgpythondir):
     config_opts['target_arch'] = 'i386'
     config_opts['releasever'] = None
     config_opts['rpmbuild_arch'] = None  # <-- None means set automatically from target_arch
-    config_opts['dnf_vars'] = {}
```
praiskup, line 250: When you remove this, users will have to do this in the configuration file anyway which is inconvenient.

mock/py/mockbuild/package_manager.py
```diff
@@ -178,6 +178,15 @@ def get_pkg_manager_config(self):
         else:
             return self.config['yum.conf']
 
+    @traceLog()
+    def get_pkg_manager_vars(self):
```
praiskup, line 182: The 'alias_spec' you specified should cause that this wrapper isn't needed, IIRC.

mock/py/mockbuild/package_manager.py
```diff
@@ -496,6 +505,15 @@ def __init__(self, config, buildroot, plugins, bootstrap_buildroot, is_bootstrap
...
             self._check_command()
 
+    def initialize_vars(self):
+        self.buildroot.root_log.debug('configure yum vars')
+        pkg_manager_vars = self.get_pkg_manager_vars()
+        var_path = self.buildroot.make_chroot_path('etc/yum/vars/')
+        if pkg_manager_vars:
+            for key in pkg_manager_vars.keys():
+                with open(os.path.join(var_path, key), 'w+') as conf_file:
+                    conf_file.write(pkg_manager_vars[key])
```
praiskup, line 515: I think this could be deduplicated, and moved to the parent class.  The decision
where to store the variables can be done per `package_manager` content option.

mock/py/mockbuild/package_manager.py
```diff
@@ -586,10 +605,12 @@ def __init__(self, config, buildroot, plugins, bootstrap_buildroot, is_bootstrap
 
     def initialize_vars(self):
         self.buildroot.root_log.debug('configure DNF vars')
+        pkg_manager_vars = self.get_pkg_manager_vars()
         var_path = self.buildroot.make_chroot_path('etc/dnf/vars/')
-        for key in self.config['dnf_vars'].keys():
-            with open(os.path.join(var_path, key), 'w+') as conf_file:
-                conf_file.write(self.config['dnf_vars'][key])
```
praiskup, line 592: I don't think we have to touch this at all.

mock/py/mockbuild/package_manager.py
praiskup: Ping on this.

mock/py/mockbuild/package_manager.py
praiskup: Ping on this as well.

mock/py/mockbuild/package_manager.py
praiskup: And here.

mock/py/mockbuild/config.py
praiskup: I got your point, and I agree.  Some users might want to use this in the custom configs.

praiskup:
Ping on this.

praiskup:
Closing for now.  Feel free to reopen.

PR #848, opened 2022-01-02
Set extended variables for yum (mock-2)

dpward:
yum supports extended variables, even in RHEL 6. If either `yum_vars` or `dnf_vars` are set in the mock configuration, then set extended variables for the appropriate package manager. These are used in URLs for CentOS. (This change precedes #845.)


praiskup:
I don't think we need to have this feature in EL7 Mock.  Let's keep this in v3+.

praiskup:
That means, we should continue with #847 (needs some fixes, but can be accepted).  Thank you so far!

PR #853, opened 2022-01-07
Remove failovermethod=priority from c8s and epel8 templates

michel-slm:
These were removed in DNF, and now cause warnings when populating the
mock chroot:

```
Invalid configuration value: failovermethod=priority in /var/lib/mock/epel-8-x86_64-zstd/root/etc/dnf/dnf.conf; Configuration: OptionBinding with id "failovermethod" does not exist
```

The C8S and EPEL repo definitions no longer set this key.

Signed-off-by: Michel Alexandre Salim <salimma@fedoraproject.org>

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-8.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-8.tpl
@@ -42,7 +42,6 @@ skip_if_unavailable=False
 name=CentOS Stream $releasever - BaseOS
 mirrorlist=http://mirrorlist.centos.org/?release=$releasever-stream&arch=$basearch&repo=BaseOS&infra=$infra
 #baseurl=http://mirror.centos.org/centos/$releasever-stream/BaseOS/$basearch/os/
-failovermethod=priority
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 gpgcheck=1
 skip_if_unavailable=False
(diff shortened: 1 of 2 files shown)


praiskup:
> These were removed in DNF, and now cause warnings when populating the
mock chroot:

But dnf was fixed not to produce these warnings, see this commit ebff6d0a817623f70f35b96c0439ca82b2d03aa3
Perhaps you should mention it is a revert.

michel-slm:
> > These were removed in DNF, and now cause warnings when populating the
> > mock chroot:
> 
> But dnf was fixed not to produce these warnings, see this commit [ebff6d0](https://github.com/rpm-software-management/mock/commit/ebff6d0a817623f70f35b96c0439ca82b2d03aa3) Perhaps you should mention it is a revert.

ah, I did not realize I was reverting something introduced recently, sorry. That would explain why I did not remember seeing this before.

Not sure why I'm seeing this though. I'm on 4.9.0-1.fc35, and looking at the spec I don't see substantial changes from 4.9.0-1.fc34 which is marked as closing https://bugzilla.redhat.com/show_bug.cgi?id=1961083

praiskup:
> ah, I did not realize I was reverting something introduced recently, sorry. That would explain why I did not remember seeing this before.

Actually the configs variant _without_ the failoverpriority option was never released.

> Not sure why I'm seeing this though. I'm on 4.9.0-1.fc35

This was fixed in libdnf, not dnf.  Sorry for confusion.  The version fixed is if I read correctly 0.64
https://github.com/rpm-software-management/libdnf/pull/1276

You see this warning because bootstrap is ON;  bootstrap is installed on fixed Fedora, then we switch
into the EL 8 where is an older libdnf - and there's the warning.

I really think we should finally drop this, if it stays only in the `main` branch.  The option needs
to stay set for mock.el7 (but that's in `mock-2` branch).

I can accept the patch, but please mention the commit you are reverting in the commit message.

PR #855, opened 2022-01-27
Use host_arch to select appropriate qemu-user-static binary

michel-slm:
Currently mock is hardcoded to use `qemu-x86_64-static`, which is
only valid on x86_64.

Signed-off-by: Michel Alexandre Salim <salimma@fedoraproject.org>

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -557,7 +557,7 @@ def check_arch_combination(target_arch, config_opts):
 
     config.multiply_platform_multiplier(config_opts)
     if config_opts['forcearch']:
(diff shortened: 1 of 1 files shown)


praiskup:
We don't use that binary actually.  This is only a way to check if the package
is present on the system (so we don't have to ask for package names that may
differ accros distrubtions).

michel-slm:
Ah, fair enough. my issue of trying to get `mock --forcearch=armv7hl` on an aarch64 VM is not fixed by this anyway, so I'll just close this and try and find the real issue. Thanks for the response!

PR #858, opened 2022-02-02
configs: Fedora 36 branching

praiskup:
Done by `$ ./releng/rawhide-branching.sh`.

Branching is very close:
https://fedorapeople.org/groups/schedule/f-36/f-36-key-tasks.html

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -18,7 +18,7 @@ BuildArch:  noarch
 Provides: mock-configs
 
 # distribution-gpg-keys contains GPG keys used by mock configs
-Requires:   distribution-gpg-keys >= 1.63
+Requires:   distribution-gpg-keys >= 1.64
 # specify minimal compatible version of mock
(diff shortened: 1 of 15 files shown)


mock-core-configs/mock-core-configs.spec
```diff
@@ -18,7 +18,7 @@ BuildArch:  noarch
 Provides: mock-configs
 
 # distribution-gpg-keys contains GPG keys used by mock configs
-Requires:   distribution-gpg-keys >= 1.63
+Requires:   distribution-gpg-keys >= 1.64
```
Conan-Kudo, line 21: Do we need to bump this again?

Conan-Kudo (changes requested):

mock-core-configs/mock-core-configs.spec
praiskup: Yes, sorry for being too fast on the gpg key side...  We want the rawhdie => F37 symlink there.

Conan-Kudo (approved):

PR #861, opened 2022-02-03
Fix 859

praiskup:


Conan-Kudo (approved):

mock-core-configs/mock-core-configs.spec
```diff
@@ -123,7 +123,26 @@ if [ -s /etc/mageia-release ]; then
...
+cfg=fedora-$ver-$mock_arch.cfg
+%endif
+%if 0%{?rhel}
+# Being installed on RHEL, or a RHEL fork.  Detect it.
+distro_id=$(. /etc/os-release; echo $ID)
+case $distro_id in
+  centos)
+    # This package is EL8+, and there's only CentOS Stream now.
+    distro_id=centos-stream
+    ;;
```
praiskup, line 138: Hmm, almalinux should be mapped to 'alma+epel'.

praiskup: Interesting, we don't have `centos-stream+epel-9` so far.  Going to add this.

praiskup:
There's no s390x in c8s, sorry.  So this should be ready now.

PR #864, opened 2022-02-06
mock: copy /usr/share/pki source CA certificates

derekschrock:
Like commit dcdaf718 if a source CA is copied to /usr/share/pki on the
host update-ca-trust inside the chroot will remove these from the trust.

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -303,13 +303,14 @@ Error:      Neither dnf-utils nor yum-utils are installed. Dnf-utils or yum-util
 
     @traceLog()
     def copy_certs(self):
-        cert_path = "/etc/pki/ca-trust"
-        pki_dir = self.buildroot.make_chroot_path(cert_path)
-        try:
-            copy_tree(cert_path, pki_dir)
-        except DistutilsFileError:
-            warning = "Couldn't copy certs from host, %s doesn't exist or is not readable"
-            self.buildroot.root_log.debug(warning, cert_path)
+        cert_paths = ["/etc/pki/ca-trust", "/usr/share/pki/ca-trust-source"]
+        for cert_path in cert_paths:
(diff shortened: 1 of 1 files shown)


praiskup:
So the problem is that something _inside_ the chroot re-generates the ca bundle?
Why is this happening (seems like some bug in package that triggers this in `%post`, or something)

derekschrock:
Yes, like commit dcdaf71 when `ca-certificates` is installed/updated it runs `update-ca-trust` and it will use `/etc/pki/ca-trust` and `/usr/share/pki/ca-trust-source` as possible locations for additional certs.

So the order of event here would be:

0) `mock --clean --scrub=all`    Just to setup a clean env.
1) `mock --init`     During bootstrap this runs without issue since any custom CA would in `/etc/pki/ca-trust/`
2) During bootstrap `ca-certificates` is installed and `update-ca-trust` is ran during `%postinstall` or `%posttrans`.   Since inside the chroot `/usr/share/pki/ca-trust-source/anchors/` is empty the custom CAs are removed.
3) Now the second part of `mock --init`  runs where the base chroot is created.  Since the custom CA is missing (in our case) you'll have SSL issues since we depend on the custom CA root cert.
4) Run `mock --init` again and the base chroot runs since `/etc/pki/ca-trust` is copied back in to the chroot.

I would expect the above to not fail again until there's an update to `ca-certificates` then you need to run `mock ...` watch it fail then run it again where the `/etc/pkg/ca-trust` is copied again.

The original commit handled the case when the host had custom CA certs in `/etc/pki/ca-trust/source/anchors` however since `update-ca-trust` used both `/etc/pki/ca-trust` and `/usr/share/pki/ca-trust-source` as possible locations for certs `/usr/share/pki/ca-trust-source` should also be included.

I don't think you could push this back on the package since it's doing what it should.

praiskup:
Thanks for the patch!

PR #867, opened 2022-02-18
EOL Fedora 37/Rawhide armhfp configs

hroncok:
https://fedoraproject.org/wiki/Changes/RetireARMv7

Not sure if the link should remain as is, happy to remove it.


hroncok:
Alternative: https://github.com/rpm-software-management/mock/pull/868

PR #868, opened 2022-02-18
Delete Fedora 37/Rawhide armhfp configs

hroncok:
https://fedoraproject.org/wiki/Changes/RetireARMv7


Conan-Kudo (approved):
Looks good to me. To note: deleting this makes sense since the configs wouldn't work at all in the future anyway (the compose trees are getting deleted as we speak).

xsuchy (approved):

PR #871, opened 2022-02-23
config: add epel-9 chroot alias

praiskup:
This is needed for the 'fedpkg mockbuild' inside 'epel9' branch (this
defaults to the 'mock -r epel-9-*' command.

Currently there's only CentOS Stream 9 + EPEL config for the 'epel-9'
variant.

### DIFF
--- a/mock-core-configs/etc/mock/chroot-aliases.cfg
+++ b/mock-core-configs/etc/mock/chroot-aliases.cfg
@@ -36,3 +36,14 @@ config_opts["no-config"]["epel-8"]["alternatives"] = {
         ],
     },
 }
+
+config_opts["no-config"]["epel-9"] = {}
+config_opts["no-config"]["epel-9"]["alternatives"] = {
+    "centos-stream+epel-9": {
+        "description": [
+            "Builds against CentOS Stream 9 repositories together "
+            "with the official EPEL 9 repositories.",
+            "Project page: https://www.centos.org/centos-stream/"
+        ],
+    },
+}


hroncok:
Considering this is the only option, should mock-core-configs just contain that symbolic link for now?

xsuchy:
> Considering this is the only option, should mock-core-configs just contain that symbolic link for now?

*nod* or even better epel-9.cfg which includes c9s+epel config. Because that will be handled better during upgrades than a symlink.

praiskup:
I don't know.  RHEL 9 is Beta now.  I would rather concentrate on this long-term config than some temporary solution.
Though I can keep this PR open .... and perhaps take a look at the RHEL 9 Beta configs first.

hroncok:
I don't understand how is the Beta relevant. EPEL 9 will not target RHEL 9 Beta.

praiskup:
Well, I thought that rhel+epel could be used as an alternative variant, too.

hroncok:
Yes, but not with Beta. https://communityblog.fedoraproject.org/epel-9-is-now-available/

> Thanks to [CentOS Stream 9](https://blog.centos.org/2021/12/introducing-centos-stream-9/), we have a close approximation of what RHEL 9 will look like when it is released. We have set up EPEL 9 to build against CentOS Stream 9. **After the RHEL 9 release**, we will switch it to build against RHEL 9 instead. We intend for this switch to be basically invisible to package maintainers and users.

hroncok:
Technically, you *can* create a rhel9beta+epel9 chroot, but it will most certainly have much weird brokenness.

PR #872, opened 2022-02-25
bootstrap_image: podman run with seccomp=unconfined

praiskup:
Related: rhbz#2034209

### DIFF
--- a/mock/py/mockbuild/podman.py
+++ b/mock/py/mockbuild/podman.py
@@ -25,7 +25,8 @@ class Podman:
     @traceLog()
     def get_container_id(self):
         """ start a container and detach immediately """
-        cmd = ["podman", "run", "--quiet", "-i", "--detach", self.image, "/bin/bash"]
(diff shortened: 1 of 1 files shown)


praiskup:
Note this PR is EL7 only, but we disabled seccomp for system-nspawn in 2f1e09c6d8cebb313ea9de12da29f735e30f907c so not sure....  should we do this for `main` branch, too?

xsuchy:
I understand this is needed only on EL7 and only as temporary measurement. I do not see reason why this should be needed in main. Right?
For mock-2 +1

xsuchy (approved):

praiskup:
> I do not see reason why this should be needed in main. Right?

These seccomp rules have bitten us many times before, so I think we'll have to do this sooner or later
in `main` too.

praiskup:
Mirek: We should have a knob for turning on the seccomp (both for podman and systemd-nspawn)

praiskup:
Moving to #885.

Issue #879, opened 2022-03-21
Fedora ELN fails to build


praiskup:
```
[SKIPPED] zstd-1.5.2-1.eln114.x86_64.rpm: Already downloaded                   
Fedora - ELN - Developmental packages for the n 1.6 MB/s | 1.6 kB     00:00    
GPG key at file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary (0x5323552A) is already installed
Fedora - ELN - Developmental packages for the n 1.6 MB/s | 1.6 kB     00:00    
GPG key at file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-37-primary (0x5323552A) is already installed
Fedora - ELN - Developmental packages for the n 1.6 MB/s | 1.6 kB     00:00    
GPG key at file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-36-primary (0x38AB71F4) is already installed
The GPG keys listed for the "Fedora - ELN - Developmental packages for the next Enterprise Linux release" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: dwz-0.14-2.eln112.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary, file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-37-primary, file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-36-primary
Public key for rpmautospec-rpm-macros-0.2.5-1.fc35.noarch.rpm is not installed. Failing package is: rpmautospec-rpm-macros-0.2.5-1.fc35.noarch
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary, file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-37-primary, file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-36-primary
Error: GPG check FAILED
```

praiskup:
See also: 
https://github.com/fedora-eln/eln/issues/86

PR #882, opened 2022-03-22
allow n-2 gpg key for Fedora ELN

xsuchy:
Fixes: #879


praiskup:
@sgallagher WRT https://github.com/fedora-eln/eln/issues/86, do you think packages will be re-signed soon, or should we go with this fix?

sgallagher:
Go with this fix for now. We have no guarantees about resigning ELN packages currently. I'll look into it.

PR #884, opened 2022-03-23
move setup_uid_manager to mockbuild.uid module 

xsuchy:
Two commits:
 * one moves  setup_uid_manager() to mockbuild.uid module 
 * second simplifies it

The motivation is that fedora-review calls load_config without uidManager param
  https://pagure.io/FedoraReview/blob/master/f/src/FedoraReview/mock.py#_180
which fatally fails because recent changes require the uidManager there and it cannot be None.
When we merge this, I will prepare PR for fedora-review and we should synchronize release.

### DIFF
--- a/mock/py/mockbuild/uid.py
+++ b/mock/py/mockbuild/uid.py
@@ -16,7 +16,8 @@ _libc = ctypes.CDLL(None, use_errno=True)
 
 
 @traceLog()
-def setup_uid_manager(mockgid):
+def setup_uid_manager():
+    mockgid = grp.getgrnam('mock').gr_gid
     unprivUid = os.getuid()
     unprivGid = os.getgid()
 
@@ -46,6 +47,7 @@ class UidManager(object):
         self.unprivGid = unprivGid
         self.unprivEnviron = dict(os.environ)
         self.unprivEnviron['HOME'] = pwd.getpwuid(unprivUid).pw_dir
+        self.mockgid = grp.getgrnam('mock').gr_gid
 
     @traceLog()
     def __enter__(self):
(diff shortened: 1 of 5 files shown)


mock/py/mockbuild/uid.py
```diff
@@ -23,6 +47,7 @@ def __init__(self, unprivUid=-1, unprivGid=-1):
         self.unprivGid = unprivGid
         self.unprivEnviron = dict(os.environ)
         self.unprivEnviron['HOME'] = pwd.getpwuid(unprivUid).pw_dir
+        self.mockgid = grp.getgrnam('mock').gr_gid
```
praiskup, line 50: This is now done in two places instead of one.

praiskup:
> The motivation is that fedora-review calls load_config without uidManager param
> which fatally fails because recent changes require the uidManager there and it cannot be None.

These details and plans should be in commit messages ideally (comments contain ephemeral info).

So do I get it right that fedora-review will start using yet another "internal"?  Can we rather create some
"library" method for reading configuration?

praiskup:
> Can we rather create some "library" method

I mean, something which could be easily tested, and what we will not break for fedora-review when we first touch it in the future.

xsuchy:
Comment in the first commit enhanced.

praiskup:
That's why I just call it "library", not a library method.  The point is to maintain the logic here in mock, and test it, rather than implement the logic in fedora-review (I would simply copy-paste the code _they_ are currently using, and call it a day).

mock/py/mockbuild/uid.py
praiskup: Then why not to keep the argument in `setup_uid_manager` (previous approach)?  It sounds
logical.

I get the problem you solve -- third-party app needs to realize what the mockgid is, but it is
fair to have another "library" method for that I think.

praiskup: > It sounds logical.

What I mean here: this is generic "setup_uid_manager()", not "setup_mock_uid_manager()"

xsuchy:
I create `simple_load_config()` which can be run even as `simple_load_config("epel-7-x86_64`) and it will work.
BTW moving the constants to `mockbuild/constants` gives offer to get rid of pass PKGPYTHONDIR to a bunch of functions. But I want to leave that for other PR.

PR #885, opened 2022-03-24
Disable SECCOMP in Podman, too, but invent a knob to opt-in

praiskup:
Previously we disabled SECCOMP rules in systemd (#811).  The SECCOMP
rules were often broken for a long time.  In rhbz#2034209 we realized
that the same problem can happen with the Podman use-cases we have, so
we disable it there as well.

This commit though also implements a knob that allows our users to
change the decision we did, and turn on SECCOMP again both in Podman and
systemd-nspawn.

I realized that util.py:do() method doesn't accept "config_opts"
dictionary (same as do_with_status(), and others).  So instead of
changing all the use-cases of the method I invented yet another ugly
global variable USE_NSPAWN_SECCOMP.

Relates: rhbz#2034209
Replaces: #872


xsuchy (commented):
+1

xsuchy (approved):

PR #886, opened 2022-03-24
Provide 'epel-9' symlinks for 'fedpkg mockbuild'

praiskup:
Later this year, once there are other alternatives, we can get back to
the solution from #871.


xsuchy (approved):
+1

PR #896, opened 2022-05-06
Add RHEL9 templates and configs

carlwgeorge:
Depends on https://github.com/xsuchy/distribution-gpg-keys/pull/63 to be merged first.

### DIFF
--- a/mock-core-configs/etc/mock/chroot-aliases.cfg
+++ b/mock-core-configs/etc/mock/chroot-aliases.cfg
@@ -36,3 +36,24 @@ config_opts["no-config"]["epel-8"]["alternatives"] = {
         ],
     },
 }
+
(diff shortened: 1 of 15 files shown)


Conan-Kudo:
Why are you deleting the CentOS Stream+EPEL 9 config? We have those for EPEL 8, there's no reason not to have it for EPEL 9 too.

Conan-Kudo:
The symlink files need to be dropped, so that we switch to the same system that exists for EPEL 8.

carlwgeorge:
Dropped the symlinks and added the chroot aliases.

praiskup:
Can you please bump the `Requires: distribution-gpg-keys` to  `>= 1.70 `?  Otherwise looks good.

praiskup:
Thank you Carl and Neal!

PR #900, opened 2022-05-16
config: Add AlmaLinux 9 and AlmaLinux 9 + EPEL configs 

Conan-Kudo:
This offers an equivalent alternative to RHEL for building packages
targeted to RHEL.


Conan-Kudo:
This PR is dependent on https://github.com/xsuchy/distribution-gpg-keys/pull/64 being merged.

cc: @xsuchy

PR #909, opened 2022-06-04
config: Use MirrorCache instead of MirrorBrain to improve reliability

Conan-Kudo:
openSUSE has a new system called MirrorCache that offers an improved
mirror service experience for openSUSE users.


Conan-Kudo:
MirrorCache is on HTTPS while MirrorBrain is on HTTP. This is sufficient to switch over.

praiskup:
Thank you!

> MirrorCache is on HTTPS while MirrorBrain is on HTTP. This is sufficient to switch over.

I c&p'ed this statement into the commit message.

PR #921, opened 2022-06-21
[CR-910] Add documentation and examples of modularity enabled templates

nkadel:
See change request 910. Specific modularity channels need to be enabled for specific software bulds. Improve the documentation to reflect the availability of modularity for RHEL, and provide configuration examples in relevant templates.

### DIFF
--- a/mock-core-configs/etc/mock/templates/almalinux-8.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-8.tpl
@@ -4,6 +4,8 @@ config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'quay.io/almalinux/almalinux:8'
+# config_opts['module_enable'] = ['list', 'of', 'modules']
+# config_opts['module_install'] = ['module1/profile', 'module2/profile']
 
 
 config_opts['dnf.conf'] = """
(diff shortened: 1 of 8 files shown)


mock-core-configs/etc/mock/templates/almalinux-8.tpl
```diff
@@ -4,6 +4,8 @@ config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'quay.io/almalinux/almalinux:8'
+# config_opts['module_enable'] = ['list', 'of', 'modules']
+# config_opts['module_install'] = ['module1/profile', 'module2/profile']
```
praiskup, line 8: This is obsoleted method to control modules in buildroot, please use `module_setup_commands`.  This used to be documented in `/etc/mock/site-defaults.cfg`, but now it is in `/usr/share/doc/mock/site-defaults.cfg`.  I don't think it makes sense to document this in every configuration file.

nkadel: Then the documentatation in docs/Feature-modularity.md, which suggests precisely this syntax, is obsolete and should be updated.

Moving such documentation out of site-defaults.cfg has been somewhat awkward, and expecting people to read previous release notes to deduce currently supported configuraiton options is... well, it's demanding a bit much. I'd be happy to add contemporary documentation to docs/Feature-modularity.md instead, though I personally think having working examples is very useful indeed.

nkadel:
OK, I'll drop this one and submit a distinct, documentation only request.

Issue #925, opened 2022-06-25
not solved build mock for ubuntu package


amidevous:
https://github.com/rpm-software-management/mock/issues/923

I have this problem here when I want to make the package there are 2 files that do not exist in the project

consolehelper

https://github.com/rpm-software-management/mock/blob/mock-3.0-1/mock/mock.spec#L152

not found here

https://github.com/rpm-software-management/mock/blob/mock-3.0-1/mock/consolehelper

mock

https://github.com/rpm-software-management/mock/blob/mock-3.0-1/mock/mock.spec#L166

not found here

https://github.com/rpm-software-management/mock/blob/mock-3.0-1/mock/mock

file and folder found online

COPYING
docs
pylintrc
Makefile
etc
requirements.txt
README.md
integration-tests
run-tests.sh
contrib
mock.spec
scripts
create_default_route_in_container.sh
mockchain
setup.cfg
py
tests

```
wget https://github.com/rpm-software-management/mock/releases/download/mock-3.0-1/mock-3.0.tar.gz -O mock_3.0.orig.tar.gz

tar -xvf mock_3.0.orig.tar.gz

cd mock-3.0

dh_make -e email@example.com -s -y

nano debian/rules
```

for information we cannot use a symbolic link in a deb so must copy the file but we cannot copy an existing file

file content

```
#!/usr/bin/make -f
%:
	dh $@
override_dh_auto_install:
	mkdir -p debian/mock/etc/mock/eol/templates
	mkdir -p debian/mock/etc/mock/templates
	install -d debian/mock/usr/bin
	install -d debian/mock/usr/libexec/mock
	install mockchain debian/mock/usr/bin/mockchain
	install py/mock-parse-buildlog.py debian/mock/usr/bin/mock-parse-buildlog
	install py/mock.py debian/mock/usr/libexec/mock/mock
	cp consolehelper debian/mock/usr/bin/mock
	install create_default_route_in_container.sh debian/mock/usr/libexec/mock/
	install -d debian/mock/etc/pam.d
	cp -a etc/pam/* debian/mock/etc/pam.d/
	install -d debian/mock/etc/mock
	cp -a etc/mock/* debian/mock/etc/mock/
	install -d debian/mock/etc/security/console.apps/
	cp -a etc/consolehelper/mock debian/mock/etc/security/console.apps/mock
	install -d debian/mock/usr/share/bash-completion/completions/
[...]

PR #926, opened 2022-06-28
Stop using the gtar symlink, use just tar

praiskup:
On GNU/Linux distributions, tar is just synonym for GNU tar.  At least
we currently don't know of any distribution that would break this rull.

Relates: #925

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -164,9 +164,16 @@
 # config_opts['createrepo_on_rpms'] = False
 # config_opts['createrepo_command'] = '/usr/bin/createrepo_c -d -q -x *.src.rpm'
 
-# You can configure which tar is used (for root cache and SCM plugin)
-# valid options are: "gnutar" or "bsdtar"
-# config_opts['tar'] = "gnutar"
+
+# What tar binary should be used by Mock.
+#config_opts['tar_binary'] = "/bin/tar"
+
+# You can configure what is the tar implementation stored on the
+# config_opts['tar_binary'] path.  Depending on this, Mock will use a different
+# set of command-line options for tar commands.  Valid options are "gnutar" or
+# "bsdtar" (used by root cache and SCM plugin, but also by core Mock for
+# unpacking Podman container file-systems with --use-bootstrap-image option).
+#config_opts['tar'] = "gnutar"
 
 # if you want mock to backup the contents of a result dir before clean
 # config_opts['backup_on_clean'] = False
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -179,12 +179,8 @@ class Buildroot(object):
             podman = Podman(self, self.bootstrap_image)
             podman.pull_image()
             podman.get_container_id()
-            if self.config["tar"] == "bsdtar":
-                __tar_cmd = "bsdtar"
-            else:
-                __tar_cmd = "gtar"
             podman.install_pkgmgmt_packages()
-            podman.cp(self.make_chroot_path(), __tar_cmd)
+            podman.cp(self.make_chroot_path(), self.config["tar_binary"])
             podman.remove()
 
         self._setup_dirs()
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -99,6 +99,7 @@ def setup_default_config_opts(unprivUid):
     config_opts['createrepo_on_rpms'] = False
     config_opts['createrepo_command'] = '/usr/bin/createrepo_c -d -q -x *.src.rpm'  # default command
 
+    config_opts['tar_binary'] = "/bin/tar"
     config_opts['tar'] = "gnutar"
 
     config_opts['backup_on_clean'] = False
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -134,7 +134,6 @@ def setup_default_config_opts(unprivUid):
             'age_check': True,
             'max_age_days': 15,
             'dir': "{{cache_topdir}}/{{root}}/root_cache/",
-            'tar': "gnutar",
             'compress_program': 'pigz',
             'decompress_program': None,
             'exclude_dirs': ["./proc", "./sys", "./dev", "./tmp/ccache", "./var/cache/yum", "./var/cache/dnf",
--- a/mock/py/mockbuild/plugins/root_cache.py
+++ b/mock/py/mockbuild/plugins/root_cache.py
@@ -60,7 +60,7 @@ class RootCache(object):
         plugins.add_hook("postyum", self._rootCachePostShellHook)
         plugins.add_hook("postupdate", self._rootCachePostUpdateHook)
         self.exclude_dirs = self.root_cache_opts['exclude_dirs']
-        self.exclude_tar_cmds = []
+        self.exclude_tar_opts = []
         for ex_dir in self.exclude_dirs:
             self._tarExcludeOption(ex_dir)
 
@@ -70,7 +70,7 @@ class RootCache(object):
         else:
             anchor = ''
 
-        self.exclude_tar_cmds.append('--exclude=' + anchor + ex_dir)
+        self.exclude_tar_opts.append('--exclude=' + anchor + ex_dir)
 
     # =============
     # 'Private' API
@@ -141,10 +141,8 @@ class RootCache(object):
                     prev_cwd = os.getcwd()
                     os.chdir(mockbuild.file_util.find_non_nfs_dir())
                 mockbuild.file_util.mkdirIfAbsent(self.buildroot.make_chroot_path())
-                if self.config["tar"] == "bsdtar":
-                    __tar_cmd = "bsdtar"
-                else:
-                    __tar_cmd = "gtar"
+
+                __tar_cmd = self.config["tar_binary"]
                 mockbuild.util.do(
                     [__tar_cmd] + self.decompressArgs + ["-xf", self.rootCacheFile,
                                                          "-C", self.buildroot.make_chroot_path()],
@@ -206,17 +204,13 @@ class RootCache(object):
                 mockbuild.util.do(["sync"], shell=False)
                 self._root_cache_handle_mounts()
                 self.state.start("creating root cache")
-                if self.config['tar'] == 'bsdtar':
-                    __tar_cmd = ["bsdtar", "--one-file-system"] + self.compressArgs + \
-                                ["-cf", self.rootCacheFile,
-                                 "-C", self.buildroot.make_chroot_path()] + \
-                                self.exclude_tar_cmds + ["."]
-                else:
-                    __tar_cmd = ["gtar", "--one-file-system", "--exclude-caches", "--exclude-caches-under"] + \
-                                 self.compressArgs + \
-                                 ["-cf", self.rootCacheFile,
-                                  "-C", self.buildroot.make_chroot_path()] + \
-                                 self.exclude_tar_cmds + ["."]
+                __tar_cmd = [self.config["tar_binary"], "--one-file-system"]
+                if self.config['tar'] == 'gnutar':
+                    __tar_cmd += ["--exclude-caches", "--exclude-caches-under"]
+                __tar_cmd += self.compressArgs + \
+                        ["-cf", self.rootCacheFile,
+                         "-C", self.buildroot.make_chroot_path()] + \
+                        self.exclude_tar_opts+ ["."]
                 try:
                     mockbuild.util.do(__tar_cmd, shell=False)
                 except:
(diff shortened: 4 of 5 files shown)


xsuchy:
What about https://github.com/rpm-software-management/mock/issues/169#issuecomment-383702706

I personally like the use of gtar, because it reminds us, that there are other implementations we care about (bsdtar).

praiskup:
> What about [#169 (comment)](https://github.com/rpm-software-management/mock/issues/169#issuecomment-383702706)

I think Mandriva is ok, because they already use `config_opts['tar'] = 'bsdtar'`.

> I personally like the use of gtar, because it reminds us, that there are other
> implementations we care about (bsdtar).

Hm, this is valid argument that it is form of documenting ...  but
/bin/gtar may be anything actually, it is nothing formally standardized.

I was too long, too deep in archivers' topics, so this is somewhat clear
to me (there is at least /bin/star in Fedora, for those who care).  And
actually, I haven't seen use of gtar symlink elsewhere than in Mock...

No pressure on this, but the bad thing about the current code is that
/bin/gtar clearly complicates use of Mock on Debian-like systems at least,
where /bin/tar is GNU tar.

praiskup:
> I think Mandriva is ok, because they already use config_opts['tar'] = 'bsdtar'.

Or maybe I'm wrong?  Does Mock on Mandriva still depend on `gnutar`? @Conan-Kudo

Conan-Kudo:
> > I think Mandriva is ok, because they already use config_opts['tar'] = 'bsdtar'.
> 
> Or maybe I'm wrong? Does Mock on Mandriva still depend on `gnutar`? @Conan-Kudo

You'd have to ask @berolinux. I believe they do depend on it still.

Conan-Kudo:
`gtar` is also the name of GNU tar on BSD systems, and I think macOS as well when installed via Homebrew...

praiskup:
Hm, something feels wrong about this tar topic :-) is anyone using the `config_opts[tar] bsdtar` option actually?  Perhaps we can drop the option, or re-design how we use it?

Long-term, IMVHO:

- we should invent something like `config_opts["tar_binary"] = "/bin/tar"`, and use it **everywhere**
- we should use portable tar options where possible and we shouldn't care about the implementation (I believe we already do in many places)
- the places where we are using non-portable options, we should decide per `config_opts["tar"]` what options, but still use the pre-configured `tar_binary`

The default then could stay on /bin/gtar, and the distributions that care can just change the default according to their needs.

amidevous:
ok i just saw your praiskup patch great idea actually gnu tar
 
and better than your gtar because it is present on all distributions without any exception all linux have gnu tar

and gnu tar and also on Mac OSX and Even on Windows

replacing gtar in your config with gnu tar even allows you to open up other possibilities such as adding a mock for windows too and yes there are rpm and dnf in cygwin and cygwin for reahat so maybe that has windows compatibility too with finally its rest to see it is only theoretical

because the chroot in a linux system may cause problems but it can be tested

in any case thank you praiskup for this patch which decide to add it or not directly in the project me on my side it will be applied in patch

on my side I take care of the test and my friend andy will take care of publishing on launchpad (equivalent of copr for ubuntu) the packages and dependencies since he has an account and I don't, he has already started recompiling the missing dependencies for dnf in focal

it will happen here

https://launchpad.net/~andykimpe/+archive/ubuntu/mock

have also vera a 2 for Debian Canonical was to add Debian on launchpad but until today have not yet checked if this has been done

while waiting for a definitive return in the Debian distributions of mock

because I don't think who will put it back on the official repositories on the current version if they do it it will be as usual on the next versions it will be like ffmpeg for my part I already feel it was in precise they removed it in trusty to put it back in focal already only yum or it had yum in trusty they removed it in focal to put dnf back in jammy

it's a strong point of ubuntu and the launchpad system if a package publish in a ppa (Personal Package Archiver) and strongly ask for it and put in the following distribution

this is not the case on rhel distributions proof rpm fusion

all the packages that are in rpm fusion should normally be on the official repositories and yet this is not the case

in short, we have moved away from the subject but it was to explain to you a little how launchpad and canonical work

otherwise Miroslav I'll send you a sweater with the code for a debian package for your project

 distribution-gpg-keys

https://github.com/xsuchy/distribution-gpg-keys/pull/66

Conan-Kudo:
You know, you could just maintain your debian packaging in your own repo...

xsuchy:
> but the bad thing about the current code is that /bin/gtar clearly complicates use of Mock on Debian-like systems at least, where /bin/tar is GNU tar.

fair enough

amidevous:
this is what and planned conan i work with my best friend andy the above have and has 2 on the code

and should all be finished in the next few days

if I understood correctly

debian/patches/version.patch

```
--- mock-3.0.orig/Makefile
+++ /dev/null
@@ -1,45 +0,0 @@
-# vim:noexpandtab:autoindent:tabstop=8:shiftwidth=8:filetype=make:nocindent:tw=0:
-check:
-	./integration-tests/runtests.sh
-
-AUTHORS:
-	(GIT_DIR=.git git log | grep ^Author | sort |uniq > .authors.tmp && mv .authors.tmp AUTHORS; rm -f .authors.tmp) || (touch AUTHORS; echo 'git directory not found: installing possibly empty AUTHORS.' >&2)
-
-install-exec-hook:
-	for i in $(REPLACE_VARS_ON_INSTALL); do      \
-		file=$(DESTDIR)/$$i			;\
-                perl -p -i -e 's|^__VERSION__\s*=.*|__VERSION__="$(RELEASE_VERSION)"|' $$file ;\
-                perl -p -i -e 's|^SYSCONFDIR\s*=.*|SYSCONFDIR="$(sysconfdir)"|' $$file        ;\
-                perl -p -i -e 's|^PYTHONDIR\s*=.*|PYTHONDIR="$(pythondir)"|' $$file           ;\
-                perl -p -i -e 's|^PKGPYTHONDIR\s*=.*|PKGPYTHONDIR="$(mockbuilddir)"|' $$file  ;\
-                perl -p -i -e 's|^PKGDATADIR\s*=.*|PKGDATADIR="$(pkgdatadir)"|' $$file        ;\
-                perl -p -i -e 's|^LIBDIR\s*=.*|LIBDIR="$(libdir)"|' $$file        ;\
-        done
-	mv $(DESTDIR)/$(sbindir)/mock.py $(DESTDIR)/$(sbindir)/mock
-	[ -d $(DESTDIR)/$(bindir) ] || mkdir $(DESTDIR)/$(bindir)
-	mv $(DESTDIR)/$(sbindir)/mockchain.py $(DESTDIR)/$(bindir)/mockchain
-
-
-.PHONY: rpm srpm help install-devel-packages
-rpm:
-	tito build --test --rpm
-
-srpm:
-	tito build --test --srpm
-
-install-devel-packages:
-	yum --disablerepo='beaker*' install mock rpm-build fedora-packager vim-enhanced git-all
-
-help:
-	@echo
-	@echo "Mock Makefile targets:"
-	@echo "	dist	- generate Authors file"
-	@echo "	rpm	- build binary RPM"
-	@echo "	srpm	- build source RPM"
-	@echo "	help	- print this message"
-	@echo "Additionally, all standard automake targets are supported"
-	@echo
-
-# Tell versions [3.59,3.63) of GNU make to not export all variables.
-# Otherwise a system limit (for SysV at least) may be exceeded.
-.NOEXPORT:
--- mock-3.0.orig/docs/mock-parse-buildlog.1
+++ mock-3.0/docs/mock-parse-buildlog.1
@@ -1,4 +1,4 @@
-.TH "mock-parse-buildlog" "1" "@VERSION@" "Silvie Chlupova"
+.TH "mock-parse-buildlog" "1" "3.0"" "Silvie Chlupova"
 .SH NAME
 mock-parse-buildlog \- parse build.log
 .SH SYNOPSIS
--- /dev/null
+++ mock-3.0/docs/mock-parse-buildlog.1.orig
@@ -0,0 +1,16 @@
+.TH "mock-parse-buildlog" "1" "3.0"" "Silvie Chlupova"
+.SH NAME
+mock-parse-buildlog \- parse build.log
+.SH SYNOPSIS
+.B mock-parse-buildlog
+\fB\-\-path\fR \fIPATH\fR
+.SH DESCRIPTION
+.B mock-parse-buildlog
+parses a build.log file saved in logs directory and tries to recognize what caused that a build failed and prints to stdout the cause.
+.SH OPTIONS
+.TP
+\fB\-p\fR PATH, \fB\-\-path\fR \fIPATH\fR
+set the path where is build.log located.
+
+.SH "AUTHORS"
+Silvie Chlupova
--- mock-3.0.orig/docs/mock.1
+++ mock-3.0/docs/mock.1
@@ -1,4 +1,4 @@
-.TH "mock" "1" "@VERSION@" "Seth Vidal" ""
+.TH "mock" "1" "3.0"" "Seth Vidal" ""
 .SH "NAME"
 .LP
 mock \- build SRPMs in a chroot
--- mock-3.0.orig/py/mock.py
+++ mock-3.0/py/mock.py
@@ -93,7 +93,7 @@ signal_names = {1: "SIGHUP",
                 }
 
 # This line is replaced by spec file %install section.
-_MOCK_NVR = None
+_MOCK_NVR = "mock-3.0-1"
 
 
 class RepoCallback(argparse.Action):
--- mock-3.0.orig/py/mockbuild/buildroot.py
+++ mock-3.0/py/mockbuild/buildroot.py
@@ -182,7 +182,7 @@ class Buildroot(object):
             if self.config["tar"] == "bsdtar":
                 __tar_cmd = "bsdtar"
             else:
-                __tar_cmd = "gtar"
+                __tar_cmd = "tar"
             podman.install_pkgmgmt_packages()
             podman.cp(self.make_chroot_path(), __tar_cmd)
             podman.remove()
--- mock-3.0.orig/py/mockbuild/constants.py
+++ mock-3.0/py/mockbuild/constants.py
@@ -6,9 +6,9 @@ import os.path
 import sys
 
 # all of the variables below are substituted by the build system
-VERSION = "unreleased_version"
-SYSCONFDIR = os.path.join(os.path.dirname(os.path.realpath(sys.argv[0])), "..", "etc")
-PYTHONDIR = os.path.dirname(os.path.realpath(sys.argv[0]))
-PKGPYTHONDIR = os.path.join(os.path.dirname(os.path.realpath(sys.argv[0])), "mockbuild")
+VERSION="3.0"
+SYSCONFDIR="/etc"
+PYTHONDIR="/usr/lib/python3/dist-packages"
+PKGPYTHONDIR="/usr/lib/python3/dist-packages/mockbuild"
 MOCKCONFDIR = os.path.join(SYSCONFDIR, "mock")
 # end build system subs
--- mock-3.0.orig/py/mockbuild/plugins/root_cache.py
+++ mock-3.0/py/mockbuild/plugins/root_cache.py
@@ -144,7 +144,7 @@ class RootCache(object):
                 if self.config["tar"] == "bsdtar":
                     __tar_cmd = "bsdtar"
                 else:
-                    __tar_cmd = "gtar"
+                    __tar_cmd = "tar"
                 mockbuild.util.do(
                     [__tar_cmd] + self.decompressArgs + ["-xf", self.rootCacheFile,
                                                          "-C", self.buildroot.make_chroot_path()],
@@ -212,7 +212,7 @@ class RootCache(object):
                                  "-C", self.buildroot.make_chroot_path()] + \
                                 self.exclude_tar_cmds + ["."]
                 else:
-                    __tar_cmd = ["gtar", "--one-file-system", "--exclude-caches", "--exclude-caches-under"] + \
+                    __tar_cmd = ["tar", "--one-file-system", "--exclude-caches", "--exclude-caches-under"] + \
                                  self.compressArgs + \
                                  ["-cf", self.rootCacheFile,
                                   "-C", self.buildroot.make_chroot_path()] + \
--- mock-3.0.orig/py/mockbuild/scm.py
+++ mock-3.0/py/mockbuild/scm.py
@@ -182,7 +182,7 @@ class scmWorker(object):
             if self.config["tar"] == "bsdtar":
                 __tar_cmd = "bsdtar"
             else:
-                __tar_cmd = "gtar"
+                __tar_cmd = "tar"
             # Always exclude vcs data from tarball unless told not to
             if str(self.exclude_vcs).lower() == "true" and __tar_cmd == 'gtar':
                 proc = subprocess.Popen(['tar', '--help'], shell=False, stdout=subprocess.PIPE)

```

praiskup:
Now, `/bin/tar` will be used by default, and we expect it is a GNU tar.
Anyone can tell Mock that the `/bin/tar` binary is actually BSD tar by `config_opts["tar"] = "bsdtar"`, or a completely different tar binary path can be specified by `config_opts["tar_binary"] = '/bin/bsdtar'`.

Can anyone PTAL?

mock/docs/site-defaults.cfg
```diff
@@ -164,9 +164,16 @@
 # config_opts['createrepo_on_rpms'] = False
 # config_opts['createrepo_command'] = '/usr/bin/createrepo_c -d -q -x *.src.rpm'
 
-# You can configure which tar is used (for root cache and SCM plugin)
-# valid options are: "gnutar" or "bsdtar"
-# config_opts['tar'] = "gnutar"
+
+# What tar binary should be used by Mock.
+#config_opts['tar_binary'] = "/bin/tar"
```
xsuchy, line 169: Should the default be /usr/bin/tar ?

xsuchy:
+1 LGTM

mock/docs/site-defaults.cfg
praiskup: I'm not sure.  On Fedora, we have the /bin -> /usr/bin link, on Debian there used to be a different approach and I recall that using just /bin is safer?  Try e.g. on `podman run --rm -ti debian:11`.

PR #931, opened 2022-07-09
let rpmbuild know that it should not clean up after itself

xsuchy:
adressing: RHBZ#2105393

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -668,8 +668,12 @@ class Commands(object):
 
     @traceLog()
     def rebuild_installed_srpm(self, spec_path, timeout):
-        command = ['{command} -bs --target {0} --nodeps {1}'.format(
-            self.rpmbuild_arch, spec_path,
+        if self.config['cleanup_on_success']:
+            noclean = ''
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/backend.py
```diff
@@ -668,7 +668,7 @@ def install_srpm(self, srpm_path):
 
     @traceLog()
     def rebuild_installed_srpm(self, spec_path, timeout):
-        command = ['{command} -bs --target {0} --nodeps {1}'.format(
+        command = ['{command} -bs --noclean --target {0} --nodeps {1}'.format(
```
praiskup, line 671: This will fail for EL6 targets.

praiskup: I would prefer not to use this option as long as "cleanup_on_success" is True.

xsuchy:
Updated.

praiskup:
Looks fine, thank you.

mock/py/mockbuild/backend.py
praiskup: A few years later, I think I was wrong: #1809

PR #944, opened 2022-07-19
config: sync epel-8 and epel-9 templates

praiskup:
Namely use metalinks, as these tend to be more reliable:
https://pagure.io/fedora-infrastructure/issue/10767


xsuchy:
Original repo files from epel uses metalink as well. So +1 from me.

PR #945, opened 2022-07-20
config: Fedora ELN fixes

praiskup:
This package doesn't appear to be available in ELN repos.

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-eln.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-eln.tpl
@@ -3,6 +3,11 @@ config_opts['eln_rawhide_releasever'] = '37'
 
 config_opts['root'] = 'fedora-eln-{{ target_arch }}'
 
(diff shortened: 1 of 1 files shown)


praiskup:
@sgallagher can you PTAL?

xsuchy:
So what's there? It was added there because several packages can satisfy the dependency and the other ones are huge.

praiskup:
Seems like `glibc-all-langpacks` is installed (18M) as a dependency.

xsuchy:
@sgallagher do you have any insight into why glibc-minimal-langpack is not present in the compose?

sgallagher:
Yes, it's a bug that I'm working on, but I'm traveling for the remainder of the week, so I'll get to it on Monday.

praiskup:
> Yes, it's a bug that I'm working on,

So I will drop the first patch, and we'll ship non-working ELN configs tomorrow.  The second patch
is still needed to make the non-composed i386 work.

Issue #907, opened 2022-06-01
Typo in error message


qhaas:
### Short description of the problem
[There is a typo in an error message](https://github.com/rpm-software-management/mock/blob/mock-3.0-1/mock/py/mockbuild/util.py#L1005) that manifest when subscription-manager is not installed.  Those of us who blindly copy/paste error messages for missing files into `dnf provides` will find that `/etc/pki/entitlment` does not exist, but `/etc/pki/entitlement` does.

### Output of `rpm -q mock`
$ rpm -q mock
mock-3.0-1.el8.noarch

### Steps to reproduce issue

1. On an EL8 system without subscription-manager installed, run `mock -r /etc/mock/rhel+epel-8-x86_64.cfg --init`
2. Observe the resulting error message has a typo (`entitlment` vs `entitlement`): `ERROR: /etc/pki/entitlment is not a directory, is subscription-manager installed?`

praiskup:
Thank you for the report, and sorry for delay.

PR #946, opened 2022-07-21
mock: typo in the subscription error message

praiskup:
Fixes: #907

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -1001,8 +1001,9 @@ def subscription_redhat_init(opts):
     if 'redhat_subscription_key_id' in opts:
         return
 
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/util.py
```diff
@@ -1001,8 +1001,9 @@ def subscription_redhat_init(opts):
     if 'redhat_subscription_key_id' in opts:
         return
 
-    if not os.path.isdir('/etc/pki/entitlement'):
-        raise exception.ConfigError("/etc/pki/entitlment is not a directory, "
+    ent_path = '/etc/pki/entitlement'
+    if not os.path.isdir(ent_path):
+        raise exception.ConfigError(ent_path + " is not a directory "
```
xsuchy, line 1006: It is better to use format() rather than concating. But +1 anyway.

PR #955, opened 2022-08-10
Branch Fedora 37 configs

hroncok:


hroncok:
The keys: https://github.com/xsuchy/distribution-gpg-keys/pull/70

praiskup:
Fixed in 799d7886c79255ccda497da8aa31e9f8b466c475

PR #965, opened 2022-09-01
Add openEuler 20.03 LTS support

Yikun:
This patch adds the mock template and config for openEuler 20.03 LTS.

### DIFF
--- a/mock-core-configs/etc/mock/openeuler-20.03-aarch64.cfg
+++ b/mock-core-configs/etc/mock/openeuler-20.03-aarch64.cfg
@@ -1,0 +1,5 @@
+include('templates/openeuler-20.03.tpl')
+
+config_opts['root'] = 'openeuler-20.03-aarch64'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/openeuler-20.03-sp3-aarch64.cfg
+++ b/mock-core-configs/etc/mock/openeuler-20.03-sp3-aarch64.cfg
@@ -1,0 +1,5 @@
+include('templates/openeuler-20.03-sp3.tpl')
+
+config_opts['root'] = 'openeuler-20.03-sp3-aarch64'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/openeuler-20.03-sp3-aarch64.cfg
+++ b/mock-core-configs/etc/mock/openeuler-20.03-sp3-aarch64.cfg
@@ -1,5 +1,0 @@
-include('templates/openeuler-20.03-sp3.tpl')
-
-config_opts['root'] = 'openeuler-20.03-sp3-aarch64'
-config_opts['target_arch'] = 'aarch64'
-config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/openeuler-20.03-sp3-x86_64.cfg
+++ b/mock-core-configs/etc/mock/openeuler-20.03-sp3-x86_64.cfg
@@ -1,0 +1,5 @@
+include('templates/openeuler-20.03-sp3.tpl')
+
+config_opts['root'] = 'openeuler-20.03-sp3-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
--- a/mock-core-configs/etc/mock/openeuler-20.03-sp3-x86_64.cfg
+++ b/mock-core-configs/etc/mock/openeuler-20.03-sp3-x86_64.cfg
@@ -1,5 +1,0 @@
-include('templates/openeuler-20.03-sp3.tpl')
-
-config_opts['root'] = 'openeuler-20.03-sp3-x86_64'
-config_opts['target_arch'] = 'x86_64'
-config_opts['legal_host_arches'] = ('x86_64',)
--- a/mock-core-configs/etc/mock/openeuler-20.03-x86_64.cfg
+++ b/mock-core-configs/etc/mock/openeuler-20.03-x86_64.cfg
@@ -1,0 +1,5 @@
+include('templates/openeuler-20.03.tpl')
+
+config_opts['root'] = 'openeuler-20.03-x86_64'
+config_opts['target_arch'] = 'x86_64'
+config_opts['legal_host_arches'] = ('x86_64',)
--- a/mock-core-configs/etc/mock/templates/openeuler-20.03-sp3.tpl
+++ b/mock-core-configs/etc/mock/templates/openeuler-20.03-sp3.tpl
@@ -1,0 +1,71 @@
+config_opts['chroot_setup_cmd'] = 'install tar gcc-c++ openEuler-rpm-config openEuler-release which xz sed make bzip2 gzip gcc coreutils unzip shadow-utils diffutils cpio bash gawk rpm-build info patch util-linux findutils grep'
+config_opts['dist'] = 'oe2003sp3'  # only useful for --resultdir variable subst
+config_opts['releasever'] = '20.03-LTS-SP3'
+config_opts['package_manager'] = 'dnf'
+config_opts['description'] = 'openEuler 20.03 LTS SP3'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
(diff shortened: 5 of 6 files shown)


Yikun:
cc @TommyLike @xsuchy

Yikun:
looks like failure is unrelated?

xsuchy:
What is the difference between regular openeuler 20.03 and the sp3? 
We see only the latest major version for OS. E.g., for RHEL we have RHEL 8 and not 8.1, 8.2, 8.3...

Yikun:
@xsuchy IIUC, just like suse leap has 15.3 and 15.4.

The openEuler has different repo url and update url for each sp version. Currently, openEuler has:

- 20.03, 20.03 sp1, 20.03sp2, 20.03sp3
- 20.09
- 22.03(will have sp version soon)
- 22.09

For each LTS version(20.03, 22.03 currently), will have SP version. For each SP version, has their own and separate repo that means thier package base are different not just a patch update.

praiskup:
The way we do this for OpenSUSE is a bit unfortunate, though.  Every minor version needs a new set of configuration files, and we also need to manually remove the outdated configurations.  That said - very inconvenient from the Mock upstream perspective (we don't build for OpenSUSE or openEuler ourselves).

Isn't it a better idea to just keep the configs for major versions like `openEuler-20-*` and `openEuler-22`.  And then maintain the set of repositories?  (I.e. move minors, or add SP repositories when available...)

Who would be interested in building against e.g. 20.03 when there's already 20.09?

Conan-Kudo:
I personally am unhappy that openSUSE works this way, it's annoying for everyone involved. Unfortunately the openSUSE project has decided that each point release has an independent lifecycle, so it is necessary.

If you're a member of the openEuler project, I would strongly suggest you ask them to reconsider their choices here, because it's extremely painful for ISVs and downstreams to leverage when you do this.

Yikun:
@Conan-Kudo @praiskup Thanks for comments!

I raised [a mail](https://mailweb.openeuler.org/hyperkitty/list/dev@openeuler.org/thread/Q2ENDRFBTEGHW3KMGLCLTRIO3RRETANM/) on openEuler mail list for further discussion.

> Who would be interested in building against e.g. 20.03 when there's already 20.09?

Users who are using openEuler in production env, choose the LTS version (20.03, 22.03), openEuler also supports upgrade to sp version, but each SP version has their own independent lifecycle.

I have to say that the differet independent sp version might casue some users confusing, but according current release policy each sp version is the major version has own lifecycle rather than a patch version.

praiskup:
What I meant by the question is that it matters who is going to use those new `mock-core-configs` configs.
Typically those configs are for package maintainers who randomly fix things in a distribution, and they just want to test the build.  They want to pick the "latest" configuration for the distro major version - and build.  Won't they most likely expect all available SPs are enabled by default...?

If those configs are supposed to be used by the openEuler project itself (say for production builds), I'm afraid `mock-core-configs` is not an appropriate place.  We can't promise so frequent release cadence anytime it is needed.

Note that we can also specify all the repositories in one config file, and let the user pick the appropriate set by `--enablerepo` and `--disablerepo` options.

Yikun:
> is that it matters who is going to use those new `mock-core-configs` configs.

The currently known scenarios are:
1. [OpenHPC](https://github.com/openhpc/ohpc) are using `mock-core-configs` to mock env to have test on openEuler
2. openEuler community dev are using mock to validate the software pkg on each specific version.

it's not for production builds, just for test.

> Note that we can also specify all the repositories in one config file, and let the user pick the appropriate set by --enablerepo and --disablerepo options.

This might an alternative choice, IIUC, we could introduce 20.03 and 22.03, and switch sp version by  --enablerepo and --disablerepo?

praiskup:
Makes sense.

> we could introduce 20.03 and 22.03(add all sp verison) and make the latest sp version as enabled, and switch sp version by --enablerepo and --disablerepo?

Sure, that could be done.  NB similarly you can do `--disablerepo EPOL` for the repo you just defined:
```
[EPOL]
name=EPOL
baseurl=http://repo.openeuler.org/openEuler-20.03-LTS-SP3/EPOL/main/$basearch/
enabled=1
gpgcheck=1
gpgkey=file:///usr/share/distribution-gpg-keys/openeuler/RPM-GPG-KEY-openEuler
```

Note the `enabled` config option, using that you can decide what repositories will be enabled by default.

praiskup:
@Yikun what do you think?  Would you re-work the PR towards a single config per major version?

Yikun:
@praiskup Oh, I missed the reply before!

Yes, I will re-work soon! Thanks for your suggestion.

Yikun:
@praiskup done, we first rework this to 20.03 (no SP repos provided), if we still think sp version is useful in future, we will append disabled sp3 repo url then, otherwise just keep main version in here.

also cc @TommyLike

praiskup:
No strong opinion, but wouldn't it be better to have just `openeuler-20-x86_64.cfg`, instead of `openeuler-20.03-x86_64.cfg`?
Otherwise LGTM

praiskup:
Actually, `mock -r openeuler-20.03-x86_64 --shell` fails for me on chroot installation by dnf:
```
...
OS                                                                                                                                                                                                          330 kB/s | 2.7 MB     00:08    
everything                                                                                                                                                                                                  782 kB/s | 9.1 MB     00:11    
EPOL                                                                                                                                                                                                         96  B/s | 146  B     00:01    
Errors during downloading metadata for repository 'EPOL':
  - Status code: 404 for http://repo.openeuler.org/openEuler-20.03-LTS/EPOL/main/x86_64/repodata/repomd.xml (IP: 49.0.230.196)
Error: Failed to download metadata for repo 'EPOL': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
ERROR: Command failed: 
 # /usr/bin/dnf --installroot /var/lib/mock/openeuler-20.03-x86_64-bootstrap/root/ --releasever 20.03-LTS --setopt=deltarpm=False --allowerasing --disableplugin=local --disableplugin=spacewalk --disableplugin=versionlock install dnf dnf-plugins-core --setopt=tsflags=nocontexts

```

Yikun:
> No strong opinion, but wouldn't it be better to have just openeuler-20-x86_64.cfg, instead of openeuler-20.03-x86_64.cfg?

because openEuler still have 20.09 version, this is for beta test, so 20.03 is more suitable name.

> fails for me on chroot installation by dnf:

Addressed

praiskup:
Now it fails differently:
```Importing GPG key 0xB25E7F66:
 Userid     : "private OBS (key without passphrase) <defaultkey@localobs>"
 Fingerprint: 12EA 74AC 9DF4 8D46 C69C A0BE D557 065E B25E 7F66
 From       : /usr/share/distribution-gpg-keys/openeuler/RPM-GPG-KEY-openEuler
Key imported successfully
update                                                                                                                                                                                                      2.0 MB/s | 2.1 kB     00:00    
GPG key at file:///usr/share/distribution-gpg-keys/openeuler/RPM-GPG-KEY-openEuler (0xB25E7F66) is already installed
The GPG keys listed for the "update" repository are already installed but they are not correct for this package.
Check that the correct key URLs are configured for this repository.. Failing package is: ima-evm-utils-libs-1.3.1-4.oe1.x86_64
 GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/openeuler/RPM-GPG-KEY-openEuler
Error: GPG check FAILED
```

TommyLike:
> 
@Yikun @praiskup 

I think there would some issue when packaging this package at openEuler side, command below:
```
❯ rpm -Kv ima-evm-utils-libs-1.3.1-4.oe1.x86_64.rpm
ima-evm-utils-libs-1.3.1-4.oe1.x86_64.rpm:
    Header V3 RSA/SHA1 Signature, key ID 911a4386: NOKEY
    Header SHA256 digest: OK
    Header SHA1 digest: OK
    Payload SHA256 digest: OK
    V3 RSA/SHA1 Signature, key ID 911a4386: NOKEY
    MD5 digest: OK
```
which shows this package use another gpg key (911a4386) that absolutely differs from the official key , We will report this to openEuler community and try to fix it.

praiskup:
> We will report this to openEuler community and try to fix it.

Thank you!  Can you link us to the report, or perhaps suggest what will be the fix? (gpg key change, or re-signing packages)

TommyLike:
> 

Yeah, we will resign these packages soon, FYI [Issue](https://gitee.com/openeuler/community/issues/I5SWOX?from=project-issue)

praiskup:
Ok, in that case the configs should be OK.  I'll plan to release it tomorrow, so let's merge them...
Please keep them maintained, and thank you for your contribution!

Yikun:
@praiskup @TommyLike Thanks for your help!

PR #966, opened 2022-09-01
Update document for 'use_bootstrap_image'

TommyLike:
boostrap image is used when `use_bootstrap_image` configured true, update document for easy understanding.

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -148,7 +148,7 @@
 # case, you may need to have "bootstrap" chroot pre-populated from a container
 # image first (where the package manager stack is already pre-installed, so
 # mock doesn't have to).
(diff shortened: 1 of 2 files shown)


mock/docs/site-defaults.cfg
```diff
@@ -148,7 +148,7 @@
 # case, you may need to have "bootstrap" chroot pre-populated from a container
 # image first (where the package manager stack is already pre-installed, so
 # mock doesn't have to).
-#config_opts['use_bootstrap_image'] = False
+#config_opts['use_bootstrap_image'] = True
```
xsuchy, line 151: This file documents the defaults. And the default is False.

TommyLike: reasonable

PR #967, opened 2022-09-05
EL7 yum can't even install the EL9 chroot

praiskup:
Therefore we need to enable the bootstrap image feature by default (same
as we already do for Fedora 31+).

Resolves: rhbz#2122171

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -71,6 +71,24 @@ case $HOST in
             echo "config_opts['use_bootstrap_image'] = True" >> "$config"
         fi
     done
+
+    for config in etc/mock/*cfg; do
+        case $config in
+            *alma+epel-9*) ;&
+            *almalinux-9*) ;&
+            *centos-stream-9*) ;&
+            *centos-stream+epel-9*) ;&
+            *centos-stream+epel-next-9*) ;&
+            *eurolinux-9*) ;&
+            *oraclelinux-9*) ;&
+            *oraclelinux+epel-9*) ;&
+            *rhel-9*) ;&
+            *rhel+epel-9*) ;&
+            *rocky-9*) ;&
+            *rocky+epel-9*)
+                echo "config_opts['use_bootstrap_image'] = True" >> "$config" ;;
+        esac
+    done
     ;;
 esac
 


mock-core-configs/mock-core-configs.spec
```diff
@@ -68,9 +74,28 @@ case $HOST in
...
+            *centos-stream+epel-9*) ;&
+            *centos-stream+epel-next-9*) ;&
+            *eurolinux-9*) ;&
+            *oraclelinux-9*) ;&
+            *oraclelinux+epel-9*) ;&
+            *rhel-9*) ;&
+            *rhel+epel-9*) ;&
+            *rocky-9*) ;&
+            *rocky+epel-9*) ;&
+            FIX_IMAGE)
```
xsuchy, line 95: Why this line?

mock-core-configs/mock-core-configs.spec
```diff
@@ -59,6 +59,12 @@ HOST="fedora-%{fedora}"
 HOST="rhel-%{rhel}"
 %endif
 
+bootstrap_image_ON()
+{
+    echo >&2 "Turning bootstrap image ON for $1"
+    echo "config_opts['use_bootstrap_image'] = True" >> "$1"
```
xsuchy, line 65: why not simple STDOUT?

mock-core-configs/mock-core-configs.spec
praiskup: I'll clean this up. I wanted to have the lines alphabetically ordered and prepared for updates in the future.

mock-core-configs/mock-core-configs.spec
praiskup: I later realized `set -x` is in action, there's no need to create additional traffic on stdout (I'll drop the method).

xsuchy (approved):

anandb-ripencc:
Use of the bootstrap image also requires podman on the host system. Should the `mock` package also have a dependency on `podman`?

praiskup:
> Should the mock package also have a dependency on podman?

We can not add `Suggests` or `Recommends` for RHEL 7 package.  And adding just `Requires` seems like too much.  Mock works well even without Podman for many use-cases (and if podman is needed, pretty obvious error message tells the user what to do).

Also, it is discouraged (we should somehow state this in docs) to build packages for newer distributions on RHEL 7 systems where YUM/RPM stack, kernel, etc. is too old.

praiskup:
> Should the mock package also have a dependency on podman?

We can not add `Suggests` or `Recommends` for RHEL 7 package.  And adding just `Requires` seems like too much.  Mock works well even without Podman for many use-cases (and if podman is needed, pretty obvious error message tells the user what to do).

Also, it is discouraged (we should somehow state this in docs) to build packages for newer distributions on RHEL 7 systems where YUM/RPM stack, kernel, etc. is too old.

PR #970, opened 2022-09-11
Update to AlmaLinux Quay.io repo

srbala:
A new organization named `almalinuxorg` created since `almalinux` admin account is locked-down and not able to create new branches for publishing

Signed-off-by: Bala Raman <srbala@gmail.com>


Conan-Kudo:
Hey, @jaboutboul, is this right?

jaboutboul:
Yes this is correct

xsuchy (approved):

friesoft:
Maybe this should also be announced/updated, like for example here: https://wiki.almalinux.org/containers/docker-images.html#almalinux-docker-images-variants
Or is this new org only temporarily?

PR #975, opened 2022-09-27
minimize fork()ing during config loading

praiskup:
This PR should eventually

- fix https://bugzilla.redhat.com/show_bug.cgi?id=2080735
- make the --list-configs much faster
- make the configuration easier to read
- make the configuration logic easier for security enhancement (see below)

I wish we could invent a new way of Mock configuration in the future, which doesn't require `exec()` call.  I'd suggest using `yaml` for this, and for the time being (one or two major releases) allow both variants).

### DIFF
--- a/mock/py/mockbuild/exception.py
+++ b/mock/py/mockbuild/exception.py
@@ -11,17 +11,20 @@
 
 
 class Error(Exception):
-    "base class for our errors."
-    def __init__(self, msg, status=None):
-        Exception.__init__(self)
-        self.msg = msg
-        self.resultcode = 1
-        if status is not None:
-            self.resultcode = status
+    resultcode = 1
+
+    def __init__(self, *args):
+        """
+        A base class for our errors.  The exit code can be specified as
+        self.resultcode.  If multiple ARGS are specified, the second one is
+        used as the resultcode.
+        """
+        super().__init__(*args)
+        if len(args) > 1:
+            self.resultcode = args[1]
 
     def __str__(self):
-        return self.msg
-
+        return str(self.args[0])
 
 # result/exit codes
 # 0 = yay!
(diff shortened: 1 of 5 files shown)


praiskup:
See commit messages for more info.

mock/py/mockbuild/exception.py
```diff
@@ -11,17 +11,16 @@
...
-            self.resultcode = status
+    def __init__(self, *args):
+        super().__init__(*args)
+        if len(args) > 1:
+            self.resultcode = args[1]
 
     def __str__(self):
-        return self.msg
-
+        return str(self.args[0])
```
xsuchy, line 23: Hmmm, technically ... yes. But the readeability of the code is much worse. You have to scroll up in the code to read what is args[0] when you will be debugging the code.

mock/py/mockbuild/exception.py
```diff
@@ -55,7 +54,7 @@ def get_class_by_code(exit_code):
     elif exit_code == 1:
         return Error("Unknow error happened.")
     elif exit_code == 2:
-        return Error("Run without setuid wrapper.", 2)
+        return Error("Run without setuid wrapper.")
```
xsuchy, line 57: Why? Isn't better way to add the exit code in other parts of this if/else?

praiskup: Tihs is a typo.

mock/py/mockbuild/exception.py
praiskup: I agree it is hard to read (even though I'm not sure the old code was better, the fact Exception takes named argument is very weird).

The purpose of this change is that we need to keep `__init__()` without arguments (as Exceptions normally are) otherwise we can not pickle => unpickle (automatically done by the multiprocessing module).

mock/py/mockbuild/exception.py
```diff
@@ -11,17 +11,16 @@
...
+    resultcode = 1
+
     "base class for our errors."
-    def __init__(self, msg, status=None):
-        Exception.__init__(self)
-        self.msg = msg
-        self.resultcode = 1
-        if status is not None:
-            self.resultcode = status
+    def __init__(self, *args):
```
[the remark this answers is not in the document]
praiskup: +1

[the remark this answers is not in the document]
praiskup: Fixed.

xsuchy (approved):

hroncok:
FTR this has regressed on Fedora CI. It now tells me:

> Can't parse Mock configuration under root

Will report a full issue next week.

hroncok:
https://github.com/rpm-software-management/mock/issues/990

Issue #953, opened 2022-08-05
The --noclean option is not on RHEL6, so cleanup_on_success=False breaks EL6


praiskup:
Forward from: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2022-4e4f7442dc

```
Update unfortunately breaks building packages for RHEL 6…

Mock Version: 3.1
ENTER ['do_with_status'](['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --noclean --target x86_64 --nodeps /builddir/build/SPECS/example.spec'], chrootPath='/var/lib/mock/centos+epel-6-x86_64-example/root'env={'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'}shell=Falselogger=<mockbuild.trace_decorator.getLog object at 0x7f714f793ba8>timeout=0uid=1000gid=135user='mockbuild'nspawn_args=[]unshare_net=TrueprintOutput=False)
Executing command: ['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --noclean --target x86_64 --nodeps /builddir/build/SPECS/example.spec'] with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'} and shell False
--noclean: unknown option
Child return code was: 1
EXCEPTION: [Error()]
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/mockbuild/[trace_decorator.py](http://trace_decorator.py/)", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.6/site-packages/mockbuild/[util.py](http://util.py/)", line 598, in do_with_status
    raise exception.Error("Command failed: \n # %s\n%s" % (command, output), child.returncode)
mockbuild.exception.Error: Command failed: 
[...]

PR #978, opened 2022-10-04
mock: don't use rpmbuild --noclean option if not supported

praiskup:
Fixes: #953


mock/py/mockbuild/backend.py
```diff
@@ -666,14 +667,27 @@ def install_srpm(self, srpm_path):
         if return_code:
             raise PkgError("Source RPM is not installable:\n{0}".format(output))
 
+
+    @property
+    def _rpmbuild_noclean_option(self):
```
xsuchy, line 672: Can you add FIXME comment here that this method and its use can be removed after some date?

xsuchy:
otherwise +1

Issue #977, opened 2022-10-04
Failure to import yum in CentOS7 bootstrap root, when /var/lib/mock/ is a bind-mount or symlink


ydirson:
### Short description of the problem

I can run `mock --init` to create a custom CentOS7-based rootfs, with the default layout where `/var/lib/mock/` is a normal dir (which happen to live in the root partition).

Now I want to move this tree (as well as `/var/cache/mock`) on another partition.  But when I use symlinks or bind mounts, `mock --init` fails with:

```
ERROR: Command failed: 
 # /usr/bin/yum --installroot /var/lib/mock/centos-7-x86_64/root/ --releasever 7 install bash bzip2 coreutils cpio diffutils findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz
error: Failed to initialize NSS library
There was a problem importing one of the Python modules
required to run yum. The error leading to this problem was:

   cannot import name ts
```

This is the error we get when manually asking python to `import yum` in the bootstrap chroot, when `/dev/` is not bind-mounted, and which disappears as soon as `/dev/` is bind-mounted.  So I guess `mock` for some reason would not setup this mountpoint  `/var/lib/mock/centos-7-x86_64/result/root.log` does in fact not show it mount `/dev`, whereas it does mount `/proc`, but then it does not show it either when the run succeeds.

### Output of `rpm -q mock`

```
mock-3.1-1.fc36.noarch
```

### Steps to reproduce issue

1. `sudo mkdir $HOME/mock/lib`
2. `sudo chgrp mock $HOME/mock/lib`
3. `sudo mount -o bind $HOME/mock/lib /var/lib/mock`
4. `mock -r centos-7-x86_64 --init`

### Any additional notes

<details>
  <summary>
    <p>Output of `mock --debug-config`</p>
  </summary>

INFO: mock.py version 3.1 starting (python version = 3.10.7, NVR = mock-3.1-1.fc36)...
Start(bootstrap): init plugins
INFO: selinux disabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux disabled
Finish: init plugins
INFO: Signal handler active
Start: run
config_opts['additional_packages'] = []
config_opts['bootstrap_image'] = 'registry.fedoraproject.org/fedora:{{ releasever }}'
config_opts['build_log_fmt_str'] = '%(message)s'
config_opts['cache_alterations'] = False
[...]

PR #979, opened 2022-10-05
mock: stuff in /var/{lib,cache}/mock doesn't need SGID

praiskup:
We should only need the g+rw permissions so mock can write to those directories.  But we shouldn't need to re-chown everything to 'mock' chroot.  I tried to search through the history, and haven't found a real need for this.  According my testing, everything continues working.

Relates: #977


xsuchy:
I cannot remember the reason either. Lets do it and see if anything breaks. :)

xsuchy (approved):

PR #982, opened 2022-10-10
releng: run PyLint for each pull-request

praiskup:
This workflow is going to fail only if there are any newly created PyLint issues.  The old issues are silently ignored, but we try to not bring any new issues from now on.


TomasTomecek (approved):
Seems to work well :) you could introduce an issue and see if it's gonna fail :D

```
Run cd mock && vcs-diff-lint --print-fixed-errors
  cd mock && vcs-diff-lint --print-fixed-errors
  shell: sh -e {0}
```

xsuchy (approved):

praiskup:
> Seems to work well :) you could introduce an issue and see if it's gonna fail :D

Re-tested here: https://github.com/rpm-software-management/mock/actions/runs/3220216319/jobs/5266633167

PR #985, opened 2022-10-11
mock: re-introduce Error.msg attribute

praiskup:
The exceptions are expected to have that attribute, see c6cfaa3be. Some of our code also relies on that attribute.

Complements: ca069b44ed5954c40fd504ce2adb95eabc9dd660

### DIFF
--- a/mock/py/mockbuild/exception.py
+++ b/mock/py/mockbuild/exception.py
@@ -20,11 +20,12 @@ class Error(Exception):
         used as the resultcode.
         """
         super().__init__(*args)
+        self.msg = args[0]
         if len(args) > 1:
             self.resultcode = args[1]
 
     def __str__(self):
-        return str(self.args[0])
+        return self.msg
 
 # result/exit codes
 # 0 = yay!


xsuchy:
Then why not re-introduce it as an attribute as in https://github.com/rpm-software-management/mock/commit/ca069b44ed5954c40fd504ce2adb95eabc9dd660#diff-1d486697ab7f21e408df664c23daef3339d4cee783609d84330b76c9cc096c24L17

praiskup:
> Then why not re-introduce it as an attribute

Good point.  I just dislike these additional Exception attributes that I couldn't concentrate :-).  Fixed, PTAL.

Long term I'd prefer
1.to have `exception_to_exit_code(exc) -> int` internal mapper for our own sys.exit() calls,
2. deprecating the `get_class_by_code()` mapper and
3. introducing `mock_exit_code_interpretation() -> str` mapper which is what end users want (e.g. Copr to present the Mock error to user)

xsuchy (approved):

Issue #990, opened 2022-10-16
Can no longer run mock as root


AdamWill:
With the new release of mock - 3.2 - it seems that one can no longer run mock as root. If you try to do so, you get an error `Can't parse Mock configuration under root`.

This has always been possible before, and openQA tests rely on doing it. With this change in Rawhide, two openQA tests fail on every Rawhide update.

I wouldn't be surprised if others also run mock as root. If this is meant to be a 'security measure' against the use of `exec` in the code, well, that seems questionable. `exec` is a security risk and a code smell whether run as a regular user or as root; regular users have lots of important data and secrets too. The comment even acknowledges this...

praiskup:
Thank you for the early report, right - running Mock as root is probably a widely used pattern.  Relaxing this in #991

hroncok:
This is from Fedora CI:

```
mock.noarch                                   3.2-1.fc38                                  @distro-build
mock-core-configs.noarch                      37.8-1.fc38                                 @distro-build
mock-filesystem.noarch                        3.2-1.fc38                                  @distro-build
...
+ mock --isolation=simple -r /tmp/fedora-38-x86_64-ci.cfg --enablerepo=local init
ERROR: Can't parse Mock configuration under root
```

praiskup:
> exec is a security risk and a code smell whether run as a regular user
> or as root; regular users have lots of important data and secrets too.
> The comment even acknowledges this...

Yes, long-term we should migrate out from the format of configuration
files have now (python code with exec()), but that will be a huge change,
and a lot of stuff will need to be changed in all the infrastructure using
Mock.

> This has always been possible before, and openQA tests rely on doing it.

Still, many parts of Mock logic are done non-privileged normally even now,
for good.  If not for security reasons, it is good to run the QA tasks the
same way they are normally done in production.  E.g. in Copr, we also
use disposable VMs, but we build non-root.

hroncok:
Due to unfortunate design choices in the Fedora CI, running tests under non-root user is very tedious. We've done it before to test a regression/fix involving unwanted PermissionErrors, but  with mock it would require manually creating users and adding them o the mock group in every package we test. I understand that this would be nicer, so I'll look into it, but if mock wants to drop support for running under root, it should probably deprecate it first.

praiskup:
Understood.  Running Mock-as-root continues to be a supported thing.  Once my time permits, I'll try to take a look at #992 (unless somebody is faster).

praiskup:
[...]

PR #991, opened 2022-10-17
mock: re-allow running mock as root

praiskup:
Turns out many users are running mock directly under root.  So don't limit them; moving the non-root check into simple_load_config.

Fixes: https://bugzilla.redhat.com/2135203
Fixes: #990

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -15,6 +15,7 @@ import pwd
 import re
 import socket
 import sys
+import warnings
 
 from templated_dictionary import TemplatedDictionary
 from . import exception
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/config.py
```diff
@@ -764,6 +759,12 @@ def list_configs(config_opts):
 @traceLog()
 def simple_load_config(name, config_path=None):
     """ wrapper around load_config() intended by use 3rd party SW """
+
+    for the_id in getresuid() + getresgid():
+        if the_id != 0:
+            continue
+        raise exception.ConfigError("Can't parse Mock configuration as root")
```
xsuchy, line 766: I would argue, that this should be just a warning. If someone wants to do that. We should allow them to shoot in their leg.

xsuchy (approved):
+1

AdamWill:
:+1: . Are you planning to do a release and build today, or should I backport this? I need a build for Rawhide at least quite urgently to fix openQA tests (and I assume Fedora CI folks would like the same to fix CI tests).

praiskup:
I'm going to release it right now.  Thank you for the review!

PR #995, opened 2022-10-24
log.info print Config with config.chroot_name variable instead buildr…

sergiomb2:
…oot.shared_root_name

Because config.chroot_name is the real name of mock --root buildroot.shared_root_name is just the mock build directory

for example for: mock -r epel+rpmfusion_free-7-x86_64 config.chroot_name is epel+rpmfusion_free-7-x86_64 and buildroot.shared_root_name is centos+epel-7-x86_64

### DIFF
--- a/mock/py/mockbuild/rebuild.py
+++ b/mock/py/mockbuild/rebuild.py
@@ -18,7 +18,7 @@ def rebuild_generic(items, commands, buildroot, config_opts, cmd, post=None, cle
     start = time.time()
     try:
         for item in items:
-            log.info("Start(%s)  Config(%s)", item, buildroot.shared_root_name)
+            log.info("Start(%s)  Config(%s)", item, buildroot.config['chroot_name'])
             if clean:
                 commands.clean()
             commands.init(prebuild=not config_opts.get('short_circuit'))


praiskup:
+1

sergiomb2:
@praiskup , that is the next topic :smile:  ,  we in RPMFusion do this way intentionally 
for example rocky , rocky+epel, rocky+epel+rpmfusion_free and rocky+epel+rocky+epel+rpmfusion_free+rpmfusion_nonfree in our or in my point of view should have the same config_opts['root']   

I started devel it on https://github.com/sergiomb2/mock/tree/devel , the main goal for me is save disk space on local builds

praiskup:
>  in my point of view should have the same config_opts['root']

They can not have the same root.  Mock uses the string internally for separating caches, etc.  This is indeed a wrong pattern.

sergiomb2:
The point is rpmfusion_free use Fedora as buildroot and rpmfusion_nonfree use Fedora + rpmfusion_free as buildroot , so the buildroot is the same with more or less some subsets of packages, that is my point. The bootstrap image is the same and IMO is a waste of disk space have different directories in /var/cache/mock/ and in /var/lib/mock/  when we use mock for local builds with ccache .

praiskup:
Sharing the space in /var/cache (not sure about /var/lib) is something we could think about in the future, in some robust form though.  This must be properly designed though.  The `config["root"]` option is a brittle hack that sooner or later will break users' expectations.

The PR would just hide the symptom of a potentially larger problem, and thus I think it is a bad idea to merge.  Thank you for the proposal anyways!

A contra RFE proposal would be, unless explicitly defined, to deterministically change/generate the `config_opts.root` option at the top-level configuration to avoid these configuration issues.

sergiomb2:
@praiskup , ok we will need an RFE but it is another topic. this PR is about just print config.chroot_name variable instead buildroot directory

Issue #999, opened 2022-11-01
_rpmbuild_noclean_option


tpgxyz:
### Short description of the problem
Looks like every build a `rpmbuild --help` is called just to check if rpmbuild does have an option called `--noclean`. This was introduced here https://github.com/rpm-software-management/mock/commit/cab12758da074e8af87189bd8f4a11c73a1f1ebf

Accoding to comments in that new code it seems this `_rpmbuild_noclean_option` is only needed for RHEL 6. Nice overkill thb.

### Output of `rpm -q mock`

[tpg@omv-rockpro64 ~/OpenMandriva/gettext]$ rpm -q mock
mock-3.3-4.noarch

Build logs snippet.
```
Mock Version: 3.3
INFO: Mock Version: 3.3
INFO: Mock Version: 3.3
DEBUG: ensuring that dir exists: /var/lib/mock//cooker-aarch64/root/dev/pts
DEBUG: creating dir: /var/lib/mock//cooker-aarch64/root/dev/pts
DEBUG: ensuring that dir exists: /var/lib/mock//cooker-aarch64/root/dev/shm
DEBUG: creating dir: /var/lib/mock//cooker-aarch64/root/dev/shm
DEBUG: kernel version == 5.19.0
DEBUG: touching file: /var/lib/mock//cooker-aarch64/root/etc/fstab
DEBUG: touching file: /var/lib/mock//cooker-aarch64/root/etc/yum/yum.conf
DEBUG: touching file: /var/lib/mock//cooker-aarch64/root/etc/dnf/dnf.conf
DEBUG: touching file: /var/lib/mock//cooker-aarch64/root/var/log/yum.log
DEBUG: ensuring that dir exists: /var/lib/mock//cooker-aarch64/root/proc
DEBUG: child environment: None
DEBUG: Executing command: ['/bin/mount', '-n', '-t', 'tmpfs', '-o', 'rprivate', 'tmpfs', '/var/lib/mock//cooker-aarch64/root/proc'] with env {'TERM': 'vt100', 'SHELL': '/bin/sh', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'LANG': 'C.UTF-8'} and shell False
DEBUG: Child return code was: 0
DEBUG: ensuring that dir exists: /var/lib/mock//cooker-aarch64/root/proc
DEBUG: child environment: None
[...]

PR #1000, opened 2022-11-07
mock: more quiet rpmbuild --noclean check

praiskup:
The "DEBUG: " output might be annoying (printing every single command in the rpmbuild --help output), so let's minimize the stderr pollution with 'grep -- --noclean'.  This should be acceptable, note that this only appears in the --verbose stderr, and one needs to set "cleanup_on_success=False" explicitly.

Fixes: #999

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -684,9 +684,10 @@ class Commands(object):
             return self.rpmbuild_noclean_option
 
         self.rpmbuild_noclean_option = ""
-        out, status = self.buildroot.doChroot(["rpmbuild", "--help"],
-                                              returnOutput=True)
-        if not status and "--noclean" in out:
+        out, _ = self.buildroot.doChroot("rpmbuild --help | grep -- --noclean",
+                                         returnOutput=True, shell=True,
+                                         raiseExc=False)
+        if "--noclean" in out:
             self.rpmbuild_noclean_option = "--noclean"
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/backend.py
```diff
@@ -684,9 +684,10 @@ def _rpmbuild_noclean_option(self):
             return self.rpmbuild_noclean_option
 
         self.rpmbuild_noclean_option = ""
-        out, status = self.buildroot.doChroot(["rpmbuild", "--help"],
-                                              returnOutput=True)
-        if not status and "--noclean" in out:
+        out, _ = self.buildroot.doChroot("rpmbuild --help | grep -- --noclean",
```
xsuchy, line 687: why the grep? Isnt starting another process more expensive than parsing multiline output?

praiskup:
The [real default](https://github.com/rpm-software-management/mock/blob/0cbd5d40319b5d1db3f6bc822c8fbfe9e323c4cd/mock/py/mockbuild/config.py#L526-L531) though bothers me.  I thought it is "cleanup_on_success=True", but it is not as long as users don't set a custom resultdir explicitly (which most they don't).  Therefore, we practically always do this check ... :-(

xsuchy:
+1

xsuchy (approved):

praiskup:
Going to close this for now, see #999 for more info.

PR #998, opened 2022-10-30
check for target specific qemu-user-static

xsuchy:
resolves: #997

PR #1001, opened 2022-11-07
mock: drop the redundant qemu-user-static warning

praiskup:
In case we are on a RHEL-like distro, we will raise a self-explanatory RuntimeException anyway.

Complements: #998


praiskup:
Sorry I did not spot this during the review.

Issue #952, opened 2022-08-04
Can't build container images using mock


davidcassany:
### Short description of the problem

Container toolchain is not fully functional inside a mock environment. More specific I can't manage to run `buildah commit` command without errors inside a mock environment.

### Output of `rpm -q mock`

host os: Fedora Linux 36
mock: mock-3.0-1.fc36.noarch

### Steps to reproduce issue

As root

1. `mock -r fedora-36-x86_64 --init`
2. `mock -r fedora-36-x86_64 --install buildah`
3. `mock -r fedora-36-x86_64 --isolation simple --enable-network --shell`

Inisde the mock env:
```bash
# just dummy file to add something inside the container
dd if=/dev/random of=data bs=1024 count=10240

# Create an empty container from scratch
buildah from --name test-container scratch

# Mount the created container
mnt=$(buildah mount test-container)

# Add dummy data in it
cp data ${mnt}

# Umount the container
buildah umount test-container

# Add some metadata to the container
buildah config --author="David Cassany <dcassany@suse.com>" test-container

# Commit the container into a container image, here it fails
buildah commit --rm --format oci test-container test-image:latest
```

The error message is:
```
Getting image source signatures
Copying blob ae378299c037 done  
error committing container "test-container" to "test-image:latest": error copying layers and metadata for container "2c48e86798dc1dc86e9e20752793cb3986f3814b4fcf3b9f49fe01ff7a54eb28": writing blob: adding layer with blob "sha256:ae378299c037f42b2190ccdc9201bff4e341ffe42d187ca44c63e3ba3b2c1a6f": Error processing tar file(exit status 1): remount /, flags: 0x44000: invalid argument
```

Feels like the mock environment is not properly set to handle the container storage, I tried creating an `/etc/containers/storage.confg` file and apply different drivers but I always got the same exact result.

Feels like this could be something related about how the chroot is set and configured...

### Any additional notes

This issues has been created as a result of attempting to integrate mock and [KIWI](https://github.com/osinside/kiwi) together. In fact this is the last rock block we have to consider KIWI functional within mock. Issue tracker here https://github.com/OSInside/kiwi/issues/2020

xsuchy:
For the record - it works (for me) when I run it with systemd-nspawn:

```
mock -r fedora-36-x86_64 --isolation nspawn --enable-network --shell
```

Conan-Kudo:
cc: @ignatenkobrain

Conan-Kudo:
[...]

PR #1005, opened 2022-11-14
mockbuild/buildroot: Make DM control device available if supported

Conan-Kudo:
In order to be able to successfully build images properly inside of mock when device mapper is used (e.g. LVM), we need the device-mapper control device to exist.

However, we also need mock to gracefully ignore this if the host does not support DM/LVM and does not provide the device-mapper control device.

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -117,6 +117,7 @@ class Buildroot(object):
 
         self._homedir_bindmounts = {}
         self._setup_nspawn_btrfs_device()
+        self._setup_nspawn_devicemapper_device()
         self._setup_nspawn_loop_devices()
 
     @traceLog()
@@ -635,6 +636,13 @@ class Buildroot(object):
             self.config['nspawn_args'].append('--bind=/dev/btrfs-control')
 
     @traceLog()
+    def _setup_nspawn_devicemapper_device(self):
+        if not util.USE_NSPAWN or self.is_bootstrap:
+            return
+        if os.path.exists('/dev/mapper/control'):
+            self.config['nspawn_args'].append('--bind=/dev/mapper/control')
+
+    @traceLog()
     def _setup_nspawn_loop_devices(self):
         if not util.USE_NSPAWN or self.is_bootstrap:
             return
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
cc: @ignatenkobrain @schaefi @davidcassany @tkopecek

Conan-Kudo:
cc: @ignatenkobrain @schaefi @davidcassany @tkopecek @davide125

Conan-Kudo:
I think this will help with #952 too.

praiskup:
It looks fine to me.

schaefi (approved):
Looks good to me :+1:

mock/py/mockbuild/buildroot.py
```diff
@@ -656,6 +664,7 @@ def _setup_devices(self):
             file_util.rmtree(self.make_chroot_path("dev"), selinux=self.selinux, exclude=self.mounts.get_mountpoints())
             file_util.mkdirIfAbsent(self.make_chroot_path("dev", "pts"))
             file_util.mkdirIfAbsent(self.make_chroot_path("dev", "shm"))
+            file_util.mkdirIfAbsent(self.make_chroot_path("dev", "mapper", "control"))
```
tkopecek, line 667: ``file_util.mkdirIfAbsent(self.make_chroot_path("dev", "mapper"))``

tkopecek:
I needed to add simple change, but it works. I'm a little bit worried about security implications with exposing mapper - but it probably should  be ok as /dev/dm-* are not exposed?

mock/py/mockbuild/buildroot.py
Conan-Kudo: Fixed.

Conan-Kudo:
> I needed to add simple change, but it works. I'm a little bit worried about security implications with exposing mapper - but it probably should be ok as /dev/dm-* are not exposed?

Yeah, you won't have access to existing dm devices, only new ones created in the chroot environment. This change is basically the DM version of what we did for Btrfs: [220905541637c77ac9b09493dc24957a5dfd3475](https://github.com/rpm-software-management/mock/commit/220905541637c77ac9b09493dc24957a5dfd3475)

PR #1007, opened 2022-11-16
Better checking for qemu user static binaries

praiskup:
Relates:
https://bugzilla.redhat.com/show_bug.cgi?id=2143069
https://bodhi.fedoraproject.org/updates/FEDORA-2022-389622bd06

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -466,6 +466,7 @@
 ## mapping from target_arch (or forcearch) to arch in /usr/bin/qemu-*-static
 # config_opts['qemu_user_static_mapping'] = {
 #     'aarch64': 'aarch64',
+#     'armv7hl': 'arm',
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
@praiskup You might want to add `riscv64` as well, since we have that architecture being bootstrapped.

praiskup:
Not sure how to test riscv64... we have no such configs yet.  I'd have to blindly add those. Therefore I'd probably let someone else to contribute this config change?

Anyway, when we merge this PR, even builds for riscv64 wouldn't be fatal...  mock will just warn and sleep(5).

Conan-Kudo:
@davidlt has mock configs for Fedora RISC-V. The package name is `qemu-user-static-riscv` and the binary name is `qemu-riscv64-static`.

praiskup:
I'm OK to review patch like that.  For now, I'm merging this PR

PR #1010, opened 2022-11-23
Add some documentation for enabling modules on the CLI

bdunne:
I had trouble getting modules working in mock, but after a lot of experimentation I found that this worked for me and thought it would help others to have it documented.

My use case: https://github.com/ManageIQ/manageiq-rpm_build/pull/341

### DIFF
--- a/docs/Feature-modularity.md
+++ b/docs/Feature-modularity.md
@@ -26,6 +26,11 @@ config_opts['module_setup_commands'] = [
     ]
 ```
 
+Modules can also be specified on the command line using:
+```
+--config-opts=module_setup_commands.module_install=postgresql:13
+```
+
 The obsolete, less flexible, but still available  modularity syntax was added in Mock 1.4.2.
 
 ```


praiskup:
Thank you for the patch, but the command doesn't seem to work for me:

```
$ mock -r fedora-37-x86_64 --install postgresql-server-devel --config-opts=module_setup_commands.module_install=postgresql:13
...
postgresql-server-devel    x86_64   14.3-8.fc37 fedora    1.4 M
...
```

The structure of 'module_setup_commands' _configuration option_ is
different, see the documentation above your change.  It is not designed
for command-line options;  e.g. if you change the set of enabled modules
on command-line, it has no effect unless you first cleanup the chroot
caches.  The only safe option is to dump it into the config file ATM.

praiskup:
I believe the command in the docs you propose is not valid; not sure if you plan to continue with the PR so closing for now. Feel free to continue with the PR or with discussing (we can reopen the PR).  Anyway, thank you again for your proposal!

PR #1032, opened 2023-01-31
mock: don't use distutils copy_tree()

praiskup:
Distutils are not a part of standard library, and being moved to setuptools where copy_tree() isn't exposed.  Using shutils.copytree() seems more standard.

Resolves: rhbz#2165548


xsuchy (approved):
+1

PR #1042, opened 2023-02-20
Drop includepkgs=devtoolset* - fixes #1041

### LINKED ISSUE #1041: Incorrect `includepkgs=devtoolset*` in centos-7.tpl

```
Error: Package: devtoolset-7-rust-7.0-5.el7.x86_64 (centos-sclo-rh)
           Requires: rust-toolset-7
```

### Output of `rpm -q mock`
mock-3.5-1.el8.noarch

### Steps to reproduce issue

1. mock -r centos+epel-7-x86_64 --install devtoolset-7-rust

### Any additional notes

If I remove the line:
```
Installing:
 devtoolset-7-rust                                                      x86_64                                         7.0-5.el7                                               centos-sclo-rh                                         2.8 k
Installing for dependencies:
 audit-libs-python                                                      x86_64                                         2.8.5-4.el7                                             base                                                    76 k
 checkpolicy                                                            x86_64                                         2.5-8.el7                                               base                                                   295 k
 devtoolset-7-runtime                                                   x86_64                                         7.1-4.el7                                               centos-sclo-rh                                          20 k
 libcgroup                                                              x86_64                                         0.41-21.el7                                             base                                                    66 k
 libselinux-python                                                      x86_64                                         2.5-15.el7                                              base                                                   236 k
 libselinux-utils                                                       x86_64                                         2.5-15.el7                                              base                                                   151 k
 libsemanage-python                                                     x86_64                                         2.5-14.el7                                              base                                                   113 k
 llvm-toolset-7-llvm-libs                                               x86_64                                         5.0.1-8.el7                                             centos-sclo-rh                                          13 M
 llvm-toolset-7-runtime                                                 x86_64                                         5.0.1-4.el7                                             centos-sclo-rh                                         1.1 M
 policycoreutils                                                        x86_64                                         2.5-34.el7                                              base                                                   917 k
 policycoreutils-python                                                 x86_64                                         2.5-34.el7                                              base                                                   457 k
 python                                                                 x86_64                                         2.7.5-92.el7_9                                          updates                                                 96 k
 python-IPy                                                             noarch                                         0.75-6.el7                                              base                                                    32 k
 rust-toolset-7                                                         x86_64                                         1.26.2-1.el7                                            centos-sclo-rh                                         3.9 k
 rust-toolset-7-cargo                                                   x86_64                                         1.26.0-3.el7                                            centos-sclo-rh                                         2.3 M
 rust-toolset-7-runtime                                                 x86_64                                         1.26.2-1.el7                                            centos-sclo-rh                                          18 k
 rust-toolset-7-rust                                                    x86_64                                         1.26.2-3.el7                                            centos-sclo-rh                                          34 M
 rust-toolset-7-rust-std-static                                         x86_64                                         1.26.2-3.el7                                            centos-sclo-rh                                          28 M
 scl-utils                                                              x86_64                                         20130529-19.el7                                         base                                                    24 k
 setools-libs                                                           x86_64                                         3.3.8-4.el7                                             base                                                   620 k
```

opoplawski:


praiskup:
Can you please document why this is not needed anymore in the commit message?  Note the original patch fd78b04112ec95164c8118ace4a7925b093ec99e

opoplawski:
The commit you mention disables the sclo* repos during install - which is not what this patch does.  This patch removes the `includepkgs` option - which as shown in #1041 is needed to install `llvm-toolset-*` and `rust-toolset-*` from the `centos-sclo-rh` repository.

praiskup:
Ok, it makes sense.  I think the filter was originally added to mimic the situation in koji buildroots.
Are all packages from the sclo repos available in koji normal buidroots?

opoplawski:
I really have no idea what is available in the koji buildroots.

praiskup:
Ne try to mimic the Koji chroot configuration here, i.e. having additional
packages from scl* repositories available could cause unnecessary
surprises - users would successfully built with plain Mock, but then in
later release cycle they wouldn't be able to build in Koji.
If that was the case, I'd prefer having those separated - and enabled only
"on demand" (--enablerepo option).

@nirik, what is the Koji policy related to these scl* repos?

nirik:
So, epel7 wants to allow maintainers to build with select devtools scls. But just globally enabling them actually breaks the epel7 koji buildroot.

So, the repo is enabled, but mock's yum command is set with: 
config_opts['yum_install_command'] += " -x devtoolset*"
So, bootstrap and initial chroot are made without devtoolset, but then packages can BuildRequire them and get them. 

So, for upstream mock... I don't know that it makes sense to do anything with them?

praiskup:
Sorry for the long delay on this.

> So, epel7 wants to allow maintainers to build with select devtools
> scls.  But just globally enabling them actually breaks the epel7 koji
> buildroot.

Hmm, and then the question is what happens for "pure" rhel-7.  Because we
don't configure scl repos in the epel-specific configuration. :-(

Has this already happened, or is it currently just discussed somewhere?
I don't see SCL in the `kooji mock-config --target epel7-candidate --arch x86_64`.

> So, the repo is enabled, but mock's yum command is set with:
> config_opts['yum_install_command'] += " -x devtoolset*"

This only affects the bootstrap installation.  To affect the "minimal
buildroot", we would have to touch the `chroot_setup_cmd` option as well.

> So, for upstream mock... I don't know that it makes sense to do anything
> with them?

I think it makes sense to adjust Mock's configuration to what is happening
in Koji, yes.  I'm not sure about all the consequences till we try that,
e.g. for repeated invocations Mock does `yum update`.  Would be the
previously excluded packages updated?

praiskup:
```
$ /bin/dnf repoquery --repofrompath el7,https://kojipkgs.fedoraproject.org/repos/epel7-build/5184146/x86_64/ --disablerepo='*' --enablerepo=el7 | grep devtoolset- | wc -l
Added el7 repo from https://kojipkgs.fedoraproject.org/repos/epel7-build/5184146/x86_64/
Last metadata expiration check: 0:02:37 ago on Po 22. května 2023, 08:07:01.
1086
```
While the packages are not available to end-users, DTS packages are accessible in Koji.

praiskup:
RHEL 7 buildroots don't have these packages, so it is EPEL 7 only.

praiskup:
So technically, we already use `config_opts['yum_install_command'] += "{% if target_arch in ['x86_64', 'ppc64le', 'aarch64'] %} --disablerepo=centos-sclo*{% endif %}"` for the centos-7 configs.  So we don't need the same hack with ` -x devtoolset*` as Koji has.

I don't really know why, but `chroot_setup_cmd` isn't affected.  Perhaps YUM from the bootstrap (EL7) does different things?  Anyway, let's test this patch then.

praiskup:
Thank you for the PR!

Issue #1050, opened 2023-04-06
/usr/sbin/useradd: invalid option -- 'n'


kkzxak47:
### Short description of the problem
koji 1.32.0
mock 3.5
shadow 4.7

"invalid option" error interrupted rpm build process in koji.
I guess the option should be '-N' ?

DEBUG util.py:540:  Executing command: ['/usr/bin/systemd-nspawn', '-q', '-M', '82f34119fbf049ca8870178aa418431f', '-D', '/home/tmp_mock/testdrive-build-17-16/root', '-a', '--console=pipe', '--setenv=TERM=vt100', '--setenv=SHELL=/bin/bash', '--setenv=HOME=/builddir', '--setenv=HOSTNAME=mock', '--setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin', '--setenv=PROMPT_COMMAND=printf "\\033]0;<mock-chroot>\\007"', '--setenv=PS1=<mock-chroot> \\s-\\v\\$ ', '--setenv=LANG=C.UTF-8', '--resolv-conf=off', '/usr/sbin/useradd', '-o', '-m', '-u', '974', '-g', '135', '-d', '/builddir', '-n', 'mockbuild'] with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8', 'SYSTEMD_NSPAWN_TMPFS_TMP': '0', 'SYSTEMD_SECCOMP': '0'} and shell False
DEBUG util.py:443:  /usr/sbin/useradd: invalid option -- 'n'

### Output of `rpm -q mock`
mock-3.5-1.ky10.noarch

OUTPUT HERE

### Steps to reproduce issue

1.
2.
3.

Do not forget to mention full commandline with the mock command you executed.

### Any additional notes

<details>
  <summary>
    <p>Output of `mock --debug-config`</p>
  </summary>

  PUT THE OUTPUT HERE
</details>

xsuchy:
[...]

PR #1052, opened 2023-04-06
use -N instead of -n for useradd

xsuchy:
-n is downstream patch in Fedora and RHEL now. It was used in past. Since RHEL6 the option -N is available.
Let use it.

But that means if you want to build for RHEL5 you have to alter config_opts['useradd'] and return back -n option.

Resolves: #1050


Conan-Kudo:
Does this allow us to drop the `useradd` overrides we've been using for all the non-RH/Fedora chroots now?

Conan-Kudo:
And in that case, we should put the override in the RHEL 5 configs as part of this PR and drop the overrides everywhere else.

xsuchy:
I really hope that no one uses rhel5 config today. 
I know of Brew people.

praiskup:
> Does this allow us to drop the useradd overrides we've been using for all the non-RH/Fedora chroots now?

I'm not sure.  The difference would be:
- '/usr/sbin/useradd -o -m -u {{chrootuid}} -g {{chrootgid}} -d {{chroothome}} {{chrootuser}}'
+ '/usr/sbin/useradd -o -m -u {{chrootuid}} -g {{chrootgid}} -d {{chroothome}} -N {{chrootuser}}

Is it safe to use `-N` (capital) in all of those chroots that currently
override?  In the new RH distributions, `-n` is just an equivalent to `-N` (so
the change Mirek is doing is no-op).

> I know of Brew people.

They maintain their own configs, and typically also run an older Mock version
anyway.

Conan-Kudo:
Well, if `-N` has been the upstream one since RHEL 6 (2010), then everyone should have that option already. I didn't know what `-n` did when I wrote those stanzas and just dropped it. The correct thing to do is to put `-N` in and use it properly.

praiskup:
@Conan-Kudo we discussed with @xsuchy today - and it seems that removing the `useradd` overrides from config would make the `mock-core-configs` code (released asynchronously, e.g. we plan to release ASAP because of the new Amazon config fiules) incompatible with the old `mock` releases.  So when removing the overrides, we should at the same time bump the `mock` version requirement.

Therefore we should either merge as-is (and do the config change later on) or wait.  We are not yet ready to wrap a new `mock.rpm` release.

praiskup:
@Conan-Kudo what do you think?

Aemn567:
n

Conan-Kudo:
We should wait on this, then. It's not urgent to land, and I'd rather this be a complete change.

Conan-Kudo:
I would say you should probably fix the EL5 configs, but I don't care that much.

Issue #691, opened 2021-01-27
Feature --use-bootstrap-image doesn't work in container


praiskup:
Running mock in rootless container works fine by default, but the problem comes in el7 container, e.g. `podman run --rm -ti --privileged centos:7`.  EL7 RPM doesn't have the zstd compression, so we can not initialize bootstrap easily - we have to download the image.

But currently it fails like:
```
INFO: Using bootstrap image: fedora:rawhide
INFO: Pulling image: fedora:rawhide
time="2021-01-27T07:16:26Z" level=error msg="'overlay' is not supported over <unknown> at \"/var/lib/containers/storage/overlay\""
Error: could not get runtime: kernel does not support overlay fs: 'overlay' is not supported over <unknown> at "/var/lib/containers/storage/overlay": backing file system is unsupported for this graph driver
ERROR: Command failed: 
 # podman pull fedora:rawhide
```

Perhaps we should use something which doesn't need overlayfs support;  we in Mock anyway only want to download tarball and extract it.

praiskup:
Everything though could, in principle, be done without using Podman. We could download the
image manually, extract it, and install dnf with normal systemd-nspawn or man (2) chroot
abstraction (aka --isolation=nspawn|simple as with the other tasks we do).

praiskup:
Hm, seems like Podman can now be run in Podman, per https://www.redhat.com/sysadmin/podman-inside-container (thanks @kdudka for the reference).

praiskup:
Seems like this could help: https://github.com/containers/podman/issues/12657

IOW: Mount the image, copy (through tar??) the contents to a final bootstrap root directory the contents and install the package manager (util.do()).

praiskup:
Said that, while podman container can be run "nested", it still requires some additional configuration, see https://github.com/containers/podman/blob/main/contrib/podmanimage/stable/Containerfile

All these steps are kind off too much for us to require for running Mock in container.  For mounting
the image - we already have a documented that `podman run --privileged` is needed.

FrostyX:
Triage time: Wanted by covscan

Issue #627, opened 2020-09-11
Concurrent mock runs don't share the mountpoints


praiskup:
Doing `mock -r <chroot> --shell` in one terminal, and e.g. `mock -r <chroot> --install vim` in another doesn't work when bootstrap is enabled.
I mean, yeah - it does something - but the vim package isn't installed to the same <basedir>/root mountpoint as one would expect.

praiskup:
This is a pretty complicated problem if I interpret it correctly.  At this moment, we have very problematic
design of executing nspawn'ed package manager in bootstrap "on demand".
The bootstrap nspawn containers start/end several times during the mock run, when bootstrapped
package manager is needed to be run.

So when one mock is running nspawn with `mock --shell`, another mock can do `--install`, but it
re-initializes the bootstrap (it's own) mountpoints.

It would be ideal to run all the requested commands in one container, not to try to run multiple
containers (with systemd-nspawn -D) at the same time.  That would simplify the chroot preparation
a lot, and the overall code.

Btw., `systemd-nspawn -D` can not be run multiple times against the same directory (unless
we disable root locking by an ugly/hackish environment variable).  That's why we should start
the container once using systemd-nspawn, and then use `nsenter` to execute the commands inside
the container.  `Systemd-nspawn` + `nsenter` is equivalent pair as `podman run` + `podman exec`.
One might argue that `systemd-nspawn shell` can be used, but it can not because there's
no systemd process running inside our chroots.

Changing the abstraction of command calls according to previous paragraph would open doors
to implement other --isolation mechanisms (docker/podman).

xsuchy:
Running two mock on the same chroot was never supported. Honestly, I would expect that the second instance will print `Build root is locked by another process.` which we already in a code on some place.

praiskup:
The locking mechanism with "readers/writers like" behavior though really look like these
things should work.  It is there just to protect against accidental concurrent initialization.

I mean, there's IMO no sane reason to block users from doing this, except that we don't know
how to implement this correctly in all cases.

praiskup:
> I mean, there's IMO no sane reason to block users from doing this,

And not even saying the fact that having this working is extremely useful when debugging in `--shell`.
Work-around is to do `mock --isolation=simple --install` or disable bootstrap I believe.

PR #1073, opened 2023-05-13
Make --use-bootstrap-image work in containers

praiskup:
Fixes: #691

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -96,7 +96,7 @@ class Buildroot(object):
         self.use_bootstrap_image = self.config['use_bootstrap_image']
         self.bootstrap_image = self.config['bootstrap_image']
 
-        self.pkg_manager = package_manager(config, self, plugins, bootstrap_buildroot)
+        self.pkg_manager = None
         self.mounts = mounts.Mounts(self)
 
         self.root_log = getLog("mockbuild")
@@ -120,6 +120,26 @@ class Buildroot(object):
         self._setup_nspawn_devicemapper_device()
         self._setup_nspawn_loop_devices()
 
+
+    def set_package_manager(self):
+        """
+        (Re)Set the 'self.pkg_manager' object.  This might be a preliminary
+        choice for bootstrap chroot if 'self.uses_bootstrap_image' because we
+        the image might contain a different package manager by default.
+        Typically when 'dnf5' is desired, but 'dnf' only is baked into the
+        image.
+        """
+
+        old_name = None if not self.pkg_manager else self.pkg_manager.name
+        self.pkg_manager = package_manager(self.config, self, self.plugins,
+                                           self.bootstrap_buildroot)
+        name = self.pkg_manager.name
+        if not old_name:
+            getLog().info("Package manager %s detected and used", name)
+        elif old_name != name:
+            getLog().info("Switching package manager from %s to the %s",
+                          old_name, name)
+
     @traceLog()
     def make_chroot_path(self, *paths):
         new_path = self.rootdir
@@ -140,7 +160,7 @@ class Buildroot(object):
             self._lock_buildroot(exclusive=True)
             self._init(prebuild=prebuild, do_log=do_log)
         except BuildRootLocked:
-            pass
+            self._init_locked()
         finally:
             self._lock_buildroot(exclusive=False)
         if do_log:
@@ -157,6 +177,16 @@ class Buildroot(object):
                 file_util.mkdirIfAbsent(self.resultdir)
             except Error:
                 raise ResultDirNotAccessible(ResultDirNotAccessible.__doc__ % self.resultdir)
+
+
+    @traceLog()
+    def _init_locked(self):
+        """
+        Things to do if other mock process initialized the Buidlroot.  Stuff
+        which would otherwise happen in self._init().
+        """
+        # Detect what package manager to use.
+        self.set_package_manager()
 
     @traceLog()
     def _init(self, prebuild, do_log):
@@ -195,6 +225,7 @@ class Buildroot(object):
         self.root_log.debug('rootdir = %s', self.make_chroot_path())
         self.root_log.debug('resultdir = %s', self.resultdir)
 
+        self.set_package_manager()
         self.pkg_manager.initialize()
         self._setup_resolver_config()
         self._setup_katello_ca()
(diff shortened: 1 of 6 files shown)


praiskup:
Mock prints with `--use-bootstrap-image --config-opts=package_manager=dnf5` something like this:
```
INFO: Using bootstrap image: registry.fedoraproject.org/fedora:rawhide
INFO: Pulling image: registry.fedoraproject.org/fedora:rawhide
...
INFO: Copy content of container registry.fedoraproject.org/fedora:rawhide to /var/lib/mock/fedora-rawhide-x86_64-bootstrap/root
INFO: mounting registry.fedoraproject.org/fedora:rawhide with podman image mount
INFO: image registry.fedoraproject.org/fedora:rawhide as /var/lib/containers/storage/overlay/7821387bce6888d9804d3a725dde054835057205456e45466ef9e6c5b873d62c/merged
INFO: umounting image registry.fedoraproject.org/fedora:rawhide (/var/lib/containers/storage/overlay/7821387bce6888d9804d3a725dde054835057205456e45466ef9e6c5b873d62c/merged) with podman image umount
...
INFO: Using package manager from the bootstrap image
INFO: Searching for /bin/dnf5 in image
INFO: Searching for /bin/dnf in image
INFO: Package manager dnf detected and used
Start(bootstrap): installing dnf5 tooling
...
Installing:
 dnf5                                                        x86_64                                                5.0.9-3.fc39                                                 fedora                                                438 k
 dnf5-plugins                                                x86_64                                                5.0.9-3.fc39                                                 fedora                                                 80 k
...
INFO: Using package manager from the bootstrap image
INFO: Searching for /bin/dnf5 in image
INFO: Switching package manager from dnf to the dnf5
... from now on, DNF5 in bootstrap is used to install buildroot ...
Start: chroot init
...
INFO: Package manager dnf5 detected and used
Start: installing minimal buildroot with dnf5
...
```

Tested inside a RHEL-8-based container.

praiskup:
@xsuchy can you please take a look at this one?

mock/py/mockbuild/package_manager.py
```diff
@@ -91,12 +91,27 @@ def package_manager_class_fallback(desired, config_opts, bootstrap):
     raise Exception("No package from {} found".format(fallbacks[desired]))
 
 
+def package_manager_class_fallback_from_image(desired, bootstrap):
+    if desired not in fallbacks:
+        raise RuntimeError(f'Unexpected package manager "{desired}"')
+    path = bootstrap.make_chroot_path()
+    for manager in fallbacks[desired]:
+        getLog().info("Searching for /bin/%s in image", manager)
+        if os.path.exists(os.path.join(path, "usr", "bin", manager)):
```
xsuchy, line 100: /bin in Log() vs /usr/bin in os.path

mock/py/mockbuild/buildroot.py
```diff
@@ -140,7 +160,7 @@ def initialize(self, prebuild=False, do_log=True):
             self._lock_buildroot(exclusive=True)
             self._init(prebuild=prebuild, do_log=do_log)
         except BuildRootLocked:
-            pass
+            self._init_locked()
```
praiskup, line 163: This was a missing bit for making #627 to work (22c8fdcbd0f1de50b942a5792a0fa6e88c620946 is the first needed patch).

mock/py/mockbuild/podman.py
```diff
@@ -52,35 +35,37 @@ def install_pkgmgmt_packages(self):
...
-        podman = subprocess.Popen(cmd_podman, stdout=subprocess.PIPE)
-        cache_file = open(cache_file_name, "w")
-        cmd_compressor = [compress_program, "--stdout"]
-        compressor = subprocess.Popen(cmd_compressor, stdin=podman.stdout, stdout=cache_file)
-        compressor.communicate()
-        podman.communicate()
-        cache_file.close()
+    def exec(self, command):
+        """ exec command in container """
+        util.do(command, chrootPath=self.buildroot.make_chroot_path(),
```
xsuchy, line 40: Either the docstring is false or the body of this function. I fail to perceive how this exec command in container.

mock/py/mockbuild/podman.py
```diff
@@ -52,35 +35,37 @@ def install_pkgmgmt_packages(self):
...
-        podman.communicate()
-
-    @traceLog()
-    def remove(self):
-        """ remove the container """
-        cmd = ["podman", "rm", "-f", self.container_id]
-        util.do(cmd)
-        self.container_id = None
+        with self.mounted_image() as mount_path:
+            cmd_podman = [tar_cmd, "-C", mount_path, "-c", "."]
```
xsuchy, line 63: I would put a comment above this line. When you replaced "podman export" with tar_cmd it is less obvious what you are doing here. 
```
# create a tarball from mounted podman image and send it to PIPE
```

mock/py/mockbuild/package_manager.py
```diff
@@ -91,12 +91,29 @@ def package_manager_class_fallback(desired, config_opts, bootstrap):
     raise Exception("No package from {} found".format(fallbacks[desired]))
 
 
+def package_manager_class_fallback_from_image(desired, bootstrap):
+    if desired not in fallbacks:
+        raise RuntimeError(f'Unexpected package manager "{desired}"')
+    path = bootstrap.make_chroot_path()
+    for manager in fallbacks[desired]:
+        binary = os.path.join(path, "usr", "bin", manager)
```
xsuchy, line 99: ??? Why not self.config[f'{pm}_command'] ?  I know that we have no config nor self here, but that can be done.

xsuchy:
Otherwise ook.
This was really hard to review :)

mock/py/mockbuild/package_manager.py
praiskup: This is the `package_manager_class_fallback_from_image()` method - we try to find a pre-existing package manager (any) to install the target package manager.  This is to mimic what we do *on host* when installing bootstrap.  I'll dump there a docstring.

praiskup:
> This was really hard to review :)

/me nods, thank you very much for looking!

mock/py/mockbuild/podman.py
xsuchy: Hmm, the docstring is good now, but... I thought that point of Podman() class is to execute inside container using a podman. But this change the logic of Podman to execute the command using nspawn inside of the directory that was extracted from podman image.

praiskup: Nah, I missed this comment - that is right, and actually the point of this change. As long as we support Mock-in-containers/toolbox, it is not yet an easy thing to unconditionally spawn (nested, either Podman or Nspawn) containers by Mock.  It is though OK to let the automatic image download and mount logic on Podman (works fine in any rootless --privileged container).

PR #1074, opened 2023-05-14
mock: use shlex instead of pipes

praiskup:
Pipes is a deprecated:
DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13


xsuchy (approved):
Works on RHEL8. 
+1

PR #1076, opened 2023-05-14
releng: drop f36 from releasers

praiskup:
Fedora 36 goes EOL on 2023-05-16, per:
https://fedorapeople.org/groups/schedule/f-38/f-38-key-tasks.html


xsuchy (approved):
+1 and merging as it is trivial.

PR #1083, opened 2023-05-23
Fix a race condition with concurrent mock commands try to rebuild RMPs

cquike:
This MR fixes issue #1071.
It mainly creates the lock file *outside* of the buildroot to avoid race conditions when two mock --rebuild commands are running concurrently. Without that, some exceptions appear as shown in #1071.

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -198,13 +198,13 @@ class Buildroot(object):
         os.chmod(self.basedir, 0o775)
         file_util.mkdirIfAbsent(self.make_chroot_path())
         self.plugins.call_hooks('mount_root')
-        # intentionally we do not call bootstrap hook here - it does not have sense
+        # intentionally we do not call bootstrap hook here - it does not make sense
         self._setup_nosync()
         self.chroot_was_initialized = self.chroot_is_initialized()
         self._setup_result_dir()
         getLog().info("calling preinit hooks")
         self.plugins.call_hooks('preinit')
-        # intentionally we do not call bootstrap hook here - it does not have sense
+        # intentionally we do not call bootstrap hook here - it does not make sense
         self.chroot_was_initialized = self.chroot_is_initialized()
         if self.uses_bootstrap_image and not self.chroot_was_initialized:
             podman = Podman(self, self.bootstrap_image)
@@ -273,7 +273,7 @@ class Buildroot(object):
 
         # done with init
         self.plugins.call_hooks('postinit')
-        # intentionally we do not call bootstrap hook here - it does not have sense
+        # intentionally we do not call bootstrap hook here - it does not make sense
 
         self.mounts.mountall_user()
 
@@ -565,8 +565,13 @@ class Buildroot(object):
 
     @traceLog()
     def _open_lock(self):
+        """
+        Creates and opens the lock file.
+        The lock file is created in the parent directory of basedir.
+        If basedir = "/path/to/basedir" => /path/to/basedir.lock
+        """
         file_util.mkdirIfAbsent(self.basedir)
-        self._lock_file = open(os.path.join(self.basedir, "buildroot.lock"), "a+")
+        self._lock_file = open(os.path.join(os.path.abspath(self.basedir)+ ".lock"), "a+")
 
     @traceLog()
     def _lock_buildroot(self, exclusive):
@@ -576,11 +581,19 @@ class Buildroot(object):
         try:
             fcntl.lockf(self._lock_file.fileno(), lock_type | fcntl.LOCK_NB)
         except IOError:
+            # If the lock file cannot be locked then this process does not own it.
+            # Setting it to None to signal that no lock could be acquired
+            self._lock_file = None
             raise BuildRootLocked("Build root is locked by another process.")
 
     @traceLog()
     def _unlock_buildroot(self):
         if self._lock_file:
+            # We acquired a lock. Hence we can delete the lock file.
+            # Deleting the file must happen before closing it.
+            # Otherwise other process can acquire the lock after this process
+            # closes the lock file but before it deletes it, causing a race condition.
+            os.unlink(os.path.join(os.path.abspath(self.basedir)+ ".lock"))
             self._lock_file.close()
         self._lock_file = None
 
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/buildroot.py
```diff
@@ -566,7 +566,7 @@ def nuke_rpm_db(self):
     @traceLog()
     def _open_lock(self):
         file_util.mkdirIfAbsent(self.basedir)
-        self._lock_file = open(os.path.join(self.basedir, "buildroot.lock"), "a+")
+        self._lock_file = open(os.path.join(os.path.abspath(self.basedir)+ ".lock"), "a+")
```
praiskup, line 569: This `os.path.abspath(self.basedir)` seems like a no-op. What you did is `s/buildroot.lock/.lock/`.

mock/py/mockbuild/buildroot.py
```diff
@@ -923,6 +924,7 @@ def delete(self):
             file_util.rmtree(self.basedir, selinux=self.selinux)
         self.chroot_was_initialized = False
         self.plugins.call_hooks('postclean')
+        self._unlock_buildroot()
```
praiskup, line 927: This move doesn't look bad, but I need an explanation of why you need to do this.  What is triggering the misbehavior, concrete example?

mock/py/mockbuild/buildroot.py
```diff
@@ -576,11 +576,13 @@ def _lock_buildroot(self, exclusive):
         try:
             fcntl.lockf(self._lock_file.fileno(), lock_type | fcntl.LOCK_NB)
         except IOError:
+            self._lock_file = None
```
praiskup, line 579: What's the point of this reset?

mock/py/mockbuild/buildroot.py
cquike: Yes, I see it looks like a no-op but it is not exactly. It is probably a expensive way to get the absolute path of the basedir (irrespective of end backslashes or relative paths) and create a new file in the same parent dir of basedir:
```
self.basedir='/tmp/mock'
os.path.abspath(self.basedir)+".lock"
'/tmp/mock.lock'
```
```
self.basedir='/tmp/mock/'
os.path.abspath(self.basedir)+".lock"
'/tmp/mock.lock'
```
```
os.chdir('/tmp')
self.basedir='./mock'
os.path.abspath(self.basedir)+".lock"
'/tmp/mock.lock'
```

There are maybe more elegant solutions to it...any suggestion?

mock/py/mockbuild/buildroot.py
cquike: If it is unlock _before_ the chroot has been completely cleaned (with rmtree), then other process might kick in and will see a chroot half in the way of being cleaned.  If the two mock processes are doing a clean first then they will start two rmtree at the same time, which will cause exceptions as seen in !1071. For instance, I just tried again moving it where it was and running two process in parallel gives this exception in one of them:
```
  File "/usr/lib/python3.11/site-packages/mockbuild/file_util.py", line 68, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/var/lib/mock/fedora-36-x86_64/root/usr/share/man/ja'
```
Moving the unlock after the chroot is completely removed eliminates this kind of race condition.

Whether the unlock can happen _before_ or _after_ the call_hooks('postclean') is not clear to me. I thought that it might be safer to do it after.

mock/py/mockbuild/buildroot.py
cquike: The idea is that the mock process deletes the lock file _only_ when it owns the lock. Otherwise, a mock process that wasn't able to lock the lockfile would still delete it in line 585 (see change below).

mock/py/mockbuild/buildroot.py
praiskup: Ah, this is the effect
```diff
- /basedir/buildroot.lock
+ /basedir.lock
```

mock/py/mockbuild/buildroot.py
praiskup: Aha, can you please comment in-code?

praiskup:
Can you please summary all the details we discussed here in the commit message, or in the code?

mock/py/mockbuild/buildroot.py
cquike: Exactly

mock/py/mockbuild/buildroot.py
cquike: Sure!

cquike:
I will modify the PR with the comments of the discussion. Thank you for the review!

cquike:
I have added some comments to the code which I hope clarify the details on the locking. Please let know if those are enough or whether they should be more verbose.
Thanks!

mock/py/mockbuild/buildroot.py
```diff
@@ -576,11 +581,19 @@ def _lock_buildroot(self, exclusive):
...
             raise BuildRootLocked("Build root is locked by another process.")
 
     @traceLog()
     def _unlock_buildroot(self):
         if self._lock_file:
+            # We acquired a lock. Hence we can delete the lock file.
+            # Deleting the file must happen before closing it.
+            # Otherwise other process can acquire the lock after this process
+            # closes the lock file but before it deletes it, causing a race condition.
+            os.unlink(os.path.join(os.path.abspath(self.basedir)+ ".lock"))
```
praiskup, line 596: This doesn't feel right, I'm not sure.  What if some mock process still holds a non-exclusive lock?  Another Mock process can come then and take the exclusive/write lock, and modify the chroot contents under the hands of the first process.  WDYT?

praiskup:
Thank you very much for the update!

praiskup:
Ping

cquike:
Sorry, I did not have much free time in the last weeks due to family business. I will try to have and update in the next days.

praiskup:
Sure, no problem. In the meantime, I am curious how locking is related to `--rootdir` and `--uniqueext` (maybe a note for myself for the future research, because I seem to face some weird issues right now while testing Mock v5.0).

praiskup:
I'm closing for inactivity. Please reopen (or let me know) once you are back on this one.

Issue #1080, opened 2023-05-22
openeuler repositories are terribly slow


praiskup:
openeuler builds are very slow with mock becuase of the connectivity to the repo.openeuler.org (tested from us-east AWS):
```
DEBUG: OS                                               49 kB/s | 3.5 MB     01:12
DEBUG: everything                                       20 kB/s |  15 MB     12:21    
DEBUG: EPOL                                             15 kB/s | 4.0 MB     04:29    
```
@pkking is this expected or is it fixable?

pkking:
Hi @praiskup, currently `repo.openeuler.org` was located in Hong Kong with bandwidth about 1000M, from our lab and monitor it work as expected.

And yes the speed is not fluent from global, we are planning to setup a `metalink` service to speedup, before that for user in North American we have to manually change repo to [OCF mirror](https://mirrors.ocf.berkeley.edu/openeuler/) such as https://github.com/lxc/lxc-ci/pull/744.

pkking:
@Conan-Kudo multi baseurls will work as round-robin or priority? For most our use cases, the `repo.openeuler.org` should be pick as a prefer.

Conan-Kudo:
In order priority (first one will always be tried first, etc.)

zengchen1024:
@praiskup Could you try again by refering to the #1085 . Thanks.

zengchen1024:
@praiskup Sorry the site of [ocf](https://mirrors.ocf.berkeley.edu/openeuler/) has not been scanned yet. It may still be slow if that site is not available. I will test it first, if it works, I will leave a message. Thanks.

zengchen1024:
@praiskup You can try now again. Thanks.

PR #1085, opened 2023-05-24
update the dnf config for openeuler

zengchen1024:
This pr wants to replace the baseurl to metalink and expects to fix the issue #1080

### DIFF
--- a/mock-core-configs/etc/mock/templates/openeuler-20.03.tpl
+++ b/mock-core-configs/etc/mock/templates/openeuler-20.03.tpl
@@ -29,7 +29,6 @@ user_agent={{ user_agent }}
 [OS]
 name=OS
 metalink=https://mirrors.openeuler.org/metalink?repo=openEuler-20.03-LTS-SP3/OS&arch=$basearch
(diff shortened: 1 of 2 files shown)


mock-core-configs/etc/mock/templates/openeuler-20.03.tpl
```diff
@@ -28,42 +28,48 @@ user_agent={{ user_agent }}
 
 [OS]
 name=OS
-baseurl=http://repo.openeuler.org/openEuler-20.03-LTS-SP3/OS/$basearch/
+metalink=https://mirrors.openeuler.org/metalink?repo=openEuler-20.03-LTS-SP3/OS&arch=$basearch
+metadata_expire=7d
```
praiskup, line 32: Why 7d?  Seems like too much.

zengchen1024: Got it. I will fix it. Thanks!

praiskup:
Thank you!

PR #1090, opened 2023-05-29
Dnf 5 fixes

praiskup:


praiskup:
See also the issue https://pagure.io/releng/issue/11446 against Fedora RelEng

mock/mock.spec
```diff
@@ -63,9 +63,18 @@ BuildRequires: python%{python3_pkgversion}-devel
 %if %{with lint}
 BuildRequires: python%{python3_pkgversion}-pylint
 %endif
-Requires: dnf
+
+%if 0%{?fedora} >= 38
```
xsuchy, line 67: I would not change this for F38. Lets change it to > 38

xsuchy: Hmm, ignore me, this is is plain suggest. That is fine, and it will automagically start working in F39+

xsuchy (approved):

Issue #563, opened 2020-04-21
/etc/config is not updated after Fedora upgrade.


az-z:
### Short description of the problem

As per discussion on the buildysys:
Pavel :
I was corrected by @msuchy now, this is not %config(noreplace) because it
basically brings arch-specific file in noarch package.

So there's pretty complicated %post scriptlet for this, which doesn't ever
replace the old symlink, and only creates `rpmnew` file.

### Output of `rpm -q mock`
mock-2.1-1.fc31.noarch

### Steps to reproduce issue

1.upgrade from FC30 to FC31

### Any additional notes

<details>
 I just upgraded from FC30 to FC 31.
 it seems that mock's "default.cfg" is still pointing to the FC30 config.

cd /etc/mock/
[az@dell5000 mock]$ ls -l ./default.cfg
lrwxrwxrwx 1 root root 20 Nov 24 12:52 ./default.cfg -> fedora-30-x86_64.cfg

[az@dell5000 mock]$ uname -a
Linux dell5000 5.5.16-200.fc31.x86_64 #1 ...
...
[root@dell5000 ~]# dnf whatprovides /etc/mock/default.cfg
Last metadata expiration check: 1:41:15 ago on Sat 18 Apr 2020 02:52:52 PM EDT.
mock-core-configs-31.6-1.fc31.noarch : Mock core config files basic chroots
Repo        : fedora
Matched from:
Filename    : /etc/mock/default.cfg

mock-core-configs-32.4-1.fc31.noarch : Mock core config files basic chroots
Repo        : @System
Matched from:
Filename    : /etc/mock/default.cfg

mock-core-configs-32.4-1.fc31.noarch : Mock core config files basic chroots
Repo        : updates
Matched from:
Filename    : /etc/mock/default.cfg
</details>

xsuchy:
The file `/etc/mock/default.cfg` is symlink to e.g. fedora-29-x86_64.cfg. The problem is that after upgrade to F31 we can hardly guess whether the symlink is the default one (and user upgrading from F29->F31) or whether it is custom one. E.g. user is upgrading from F30->F31 but they wants to have default config pointed to `fedora-29-x86_64.cfg`

Note that `default.cfg` is the config used if you run mock without `-r` option.

So my original intention was that Mock should not play a smarty and simply create .rpmnew (exactly what %config(noreplace) would does) and let user handle it - simplest ways is `rpmconf -o mock-core-configs`

praiskup:
[...]

PR #1098, opened 2023-06-02
Use python3 macro for post scriptlet

omenos:
Use the %python3 macro for determining which Python interpreter to use. The interpreter selection of python3 on RHEL 8 is not guaranteed as the distro allows for any Python 3.x module stream to be used as an alternative for the `python3` symlink. However, only the `platform-python` (`python3` symlink provided by python36) has the DNF Python interfaces available to it.

On one of my RHEL 8 systems where I have Python 3.11 configured as the global `python3`, the following output was logged when upgrading:

```text
  Upgrading        : mock-core-configs-38.5-1.el8.noarch         2/8 
  Running scriptlet: mock-core-configs-38.5-1.el8.noarch         2/8 
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'dnf'
Warning: file /etc/mock/rhel+epel-8-.cfg does not exist.
         unable to update /etc/mock/default.cfg
```

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -1,3 +1,7 @@
+%if 0%{?el8}
+%global python3 /usr/libexec/platform-python
+%endif
+
 Name:       mock-core-configs
 Version:    38.5
 Release:    1%{?dist}
@@ -98,7 +102,7 @@ fi
 if [ -s /etc/mageia-release ]; then
     mock_arch=$(sed -n '/^$/!{$ s/.* \(\w*\)$/\1/p}' /etc/mageia-release)
 else
-    mock_arch=$(python3 -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
+    mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
 fi
 
 cfg=unknown-distro


mock-core-configs/mock-core-configs.spec
```diff
@@ -1,6 +1,10 @@
+%if 0%{?epel} == 8
+%global python3 /usr/libexec/platform-python
```
praiskup, line 2: Really `%epel`, not `%rhel`?  Also, what if `%rhel == 9` or higher?

mock-core-configs/mock-core-configs.spec
```diff
@@ -1,6 +1,10 @@
...
+%endif
+
 Name:       mock-core-configs
 Version:    38.5
-Release:    1%{?dist}
+Release:    2%{?dist}
```
praiskup, line 7: Please don't update `Release` and `%changelog` upstream, it is going to be automatically updated during the release (via `tito tag`).

mock-core-configs/mock-core-configs.spec
omenos: Good catch, I was just thinking of the EPEL branch when putting this together as it's an epel only package. Slip of the mind.

mock-core-configs/mock-core-configs.spec
omenos: Sounds good, will fix (wasn't aware of the process for this).

mock-core-configs/mock-core-configs.spec
```diff
@@ -98,7 +102,7 @@ fi
 if [ -s /etc/mageia-release ]; then
     mock_arch=$(sed -n '/^$/!{$ s/.* \(\w*\)$/\1/p}' /etc/mageia-release)
 else
-    mock_arch=$(python3 -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
+    mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
```
praiskup, line 105: Ok, I had to test this;  the `%python3` macro is baked into the binary RPM scriptlet expanded (at build time).  Therefore, at installation time, there's no need to have python-srpm-macros installed.

mock-core-configs/mock-core-configs.spec
omenos: @praiskup Another alternative is to use `%el8` instead (thanks @carlwgeorge) as to prevent any issues with rebuild projects that may clear out `rhel` macro.

mock-core-configs/mock-core-configs.spec
omenos: So there shouldn't be any problems with this solution?

omenos: Just verified for myself, looks good to me.

mock-core-configs/mock-core-configs.spec
omenos: Modified to `0%{?el8}`, this should be all set for merging unless there are any other areas that seem to pose a potential problem.

Conan-Kudo (approved):

praiskup (approved):
Thank you.

praiskup:
Note though a related issue #563

Issue #1088, opened 2023-05-26
Mock 4+ no longer uses the DNF config shipped in bootstrap_image


kdudka:
### Short description of the problem
After upgrading to `mock-4.0`, `mock --init` fails for a configuration that worked fine with `mock-2.16` on the same host.

### Output of `rpm -q mock`
mock-4.0-1.fc37.noarch

### Steps to reproduce issue

1. mock -r rhel-8.7.0.z-x86_64 --scrub=bootstrap
2. mock -r rhel-8.7.0.z-x86_64 --init

The second command fails with :
```
% mock -r rhel-8.7.0.z-x86_64 --init
INFO: mock.py version 4.0 starting (python version = 3.11.3, NVR = mock-4.0-1.fc37)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start: clean chroot
Finish: clean chroot
Start(bootstrap): chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: enabled package manager cache
Start(bootstrap): cleaning package manager metadata
Finish(bootstrap): cleaning package manager metadata
INFO: enabled HW Info plugin
INFO: Using bootstrap image: fedora:32
INFO: Pulling image: fedora:32
Trying to pull registry.fedoraproject.org/fedora:32...
Getting image source signatures
Copying blob sha256:480478cefb5d287032080377942da2c2ccb15cdcc5c45f3beb06aef1267bc8c4
Copying config sha256:27a9790209529e14132d05f032228f339a4fe1dd1e93a588f5af2c06c0234969
Writing manifest to image destination
Storing signatures
27a9790209529e14132d05f032228f339a4fe1dd1e93a588f5af2c06c0234969
INFO: Copy content of container fedora:32 to /var/lib/mock/rhel-8.7.0.z-x86_64-bootstrap/root
INFO: mounting fedora:32 with podman image mount
INFO: image fedora:32 as /var/lib/containers/storage/overlay/48df8b28c24ba6e7356c15f0cbc05a48b6a23241bb93a7a6e72f742a2979111e/merged
INFO: umounting image fedora:32 (/var/lib/containers/storage/overlay/48df8b28c24ba6e7356c15f0cbc05a48b6a23241bb93a7a6e72f742a2979111e/merged) with podman image umount
Mock Version: 4.0
[...]

PR #1101, opened 2023-06-06
enable use_bootstrap_image by default

xsuchy:
This feature exists for long time. I think it is time to enable it by default. It will allows us to run Mock without package manager on host.

### DIFF
--- a/docs/Feature-container-for-bootstrap.md
+++ b/docs/Feature-container-for-bootstrap.md
@@ -11,9 +11,9 @@ We come up with a nice feature. Mock will not install bootstrap chroot itself. I
 
 Using this feature, **any** incompatible feature in either RPM or DNF can be used in the target chroot. Now or in future. And you will be able to install the final chroot. You do not even need to have RPM on a host. So this should work on any system. Even Debian based. The only requirement for this feature is [Podman](https://podman.io/). Do not forget to install the `podman` package.
 
-This feature is now disabled by default. You can enable it using:
+This feature is available since 1.4.20 and enabled by default from 5.0. You can disable it using:
 
-    config_opts['use_bootstrap_image'] = True
+    config_opts['use_bootstrap_image'] = False
 
 It can be enabled or disabled on the command line using `--use-bootstrap-image` or `--no-bootstrap-image` options.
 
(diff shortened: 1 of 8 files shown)


praiskup:
This really can make building RPMs trivial on non-RPM systems, e.g., I was hesitant to underline this too much as I haven't really tested this feature on other distributions (just tested in podman `fedora:rawhide` container after removing `*dnf*` packages).  IMO we can consider Mock package existence in PyPI meaningful now.

While the feature and idea are relatively old, it has been recently reworked in 4.0+, and I'm curious whether we need some planning or blogging in advance before we do the hard switch.  E.g. #1088 raises some important questions related to potential fixes/enhancements in this feature.

praiskup:
Cross arch is a problem, too, see #1101.

xsuchy:
Updated.

docs/Feature-container-for-bootstrap.md
```diff
@@ -11,9 +11,9 @@ We come up with a nice feature. Mock will not install bootstrap chroot itself. I
 
 Using this feature, **any** incompatible feature in either RPM or DNF can be used in the target chroot. Now or in future. And you will be able to install the final chroot. You do not even need to have RPM on a host. So this should work on any system. Even Debian based. The only requirement for this feature is [Podman](https://podman.io/). Do not forget to install the `podman` package.
 
-This feature is now disabled by default. You can enable it using:
+This feature is available since 1.4.20 and enabled by default from 4.2. You can disable it using:
```
praiskup, line 14: So are we going to ship 5.0, eventually?

xsuchy: Updated.

praiskup:
JFTR, Copr already has a knob for enabling/disabling the bootstrap image feature, with Mock 5.0 in Copr we'll jump to the --use-bootstrap-image default for most of the projects.

tkopecek:
Created https://pagure.io/koji/issue/3872 - but it is not blocked for anything (builders' site-defaults can still be used for default behaviour override)

xsuchy:
Updated once again. :)

praiskup (approved):
lgtm

praiskup:
This has been open for a long time, and there are no objections (except for my own, and those have been more or less resolved).  With #1152 and e7139e1ce, seems like a pretty nice mock-buildling evolution :-)

praiskup:
Thank you!

Issue #1091, opened 2023-05-29
Self-signing Fedora kernels with pesign - socket permission problems


belegdol:
### Short description of the problem
Upon upgrading from mock-3.5-2.fc38 to mock-4.0-1.fc38 signing Fedora kernel RPMs using own certificate no longer works

### Output of `rpm -q mock`

mock-4.0-1.fc38.noarch

### Steps to reproduce issue

This is not really trivial unfortunately:
1.  Follow instructions from here https://gist.github.com/chenxiaolong/520914b191f17194a0acdc0e03122e63 with the following exception: add yourserf to pesign group: `sudo usermod -a -G pesign julas`
2.  Attempt to build a kernel: `mock -r fedora-38-x86_64 --disable-plugin=tmpfs --enable-plugin=yum_cache  --isolation=simple redhat/rpm/SRPMS/kernel-6.3.4-201.s0ix01.fc38.src.rpm --with baseonly --define='pe_signing_token NSS Certificate DB' --define='pe_signing_cert Julians Secure Boot signing key - Julian Sikorski'`
3. wait about 20 minutes (on an Ryzen 5 5600x)

The following happens:
```
+ /usr/bin/pesign --certdir /etc/pki/pesign -t 'NSS Certificate DB' -c 'Julians Secure Boot signing key - Julian Sikorski' -s -i arch/x86/boot/bzImage -o vmlinuz.tmp
pesign: Could not open NSS database ("security library: bad database."): Permission denied
error: Bad exit status from /var/tmp/rpm-tmp.tNH11n (%build)
RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.tNH11n (%build)
Child return code was: 1
```

### Any additional notes
In order to fix the problem, rolling back to mock-3.5 was not enough. I had to do:
```
$ mock -r fedora-38-x86_64 --scrub=all clean
```
for the signing to work again.

praiskup:
> pesign: Could not open NSS database ("security library: bad database."): Permission denied

Are you sure this is Mock's fault?  Seems like the directory is correctly mounted by Mock -> but 
Mock itself doesn't touch the permissions of mounted files.  Can you help a bit to diagnose what
is going on?

praiskup:
> Can you help a bit to diagnose what is going on?

By that I mean -> what permissions are expected? What mock should do to make it work?
You seem to be doing some tweaks to the `"${KEY_DIR}/nss_db"` database in your nice how-to.

[...]

PR #1103, opened 2023-06-09
mock: allow pre-creating users in chroot

praiskup:


mock/py/mockbuild/buildroot.py
```diff
@@ -466,6 +468,9 @@ def _fixup_build_user(self):
     @traceLog()
     @noop_in_bootstrap
     def _make_build_user(self):
+        self.shadow_utils.create_user("pesign", copy_from_host=True)
+        self.shadow_utils.create_user("random-user")
```
praiskup, line 472: This part likely needs to be done elsewhere and needs to be configurable!  This is just a dummy preview for experiments.

belegdol:
This fixes #1091 without having to set up ACLs. `/run/pesign` on the host remains owned by `pesign:pesign` even after installing `pesign` in the chroot.

xsuchy:
@praiskup please check buildroot._make_build_user() for historical knowledge. Like deleting the user first, because it may exists from previous run. And fixing the uid.

praiskup:
Yes, I've seen the history - in this case, I think we can afford to keep it simple, at least at the beginning;  if this logic fails, it fails for misconfiguration (which is not very likely because we do this only when the chroot is first initialized).

praiskup:
> @praiskup please check buildroot._make_build_user() for historical knowledge. Like deleting the user first, because it may exists from previous run. And fixing the uid.

I changed my opinion about this, and I agree that removing the user first is better (makes me more confident that UID/GIDs are aligned).

While doing this quite complex feature, I also propose changing the default way of handling `mockbuilder` (user) and `mock` (group) in-chroot -> so we use the same principle in both cases.

Ready for the review.

belegdol:
I tested the updated code and it works as expected, pesign works correctly.

mock/mock.spec
```diff
@@ -9,7 +9,7 @@
 
 Summary: Builds packages inside chroots
 Name: mock
-Version: 4.1
+Version: 4.1.post
```
xsuchy, line 12: ???

mock/py/mockbuild/buildroot.py
```diff
@@ -452,42 +454,26 @@ def _init_pkg_management(self):
...
+        # Resolving the 'mockbuilder' user
         excluded = [self.make_chroot_path(self.homedir, path)
                     for path in self.config['exclude_from_homedir_cleanup']] + \
             self.mounts.get_mountpoints()
         file_util.rmtree(self.make_chroot_path(self.homedir),
                     selinux=self.selinux, exclude=excluded)
 
-        # ok for these two to fail
-        if self.config['clean']:
-            self.doChroot(['/usr/sbin/userdel', '-r', '-f', self.chrootuser],
```
xsuchy, line 477: The -r is lost with the new code.

xsuchy:
> There's one different thing.  We we don't do `userdel -r` anymore for
mockbuilder.  This shouldn't be needed, because when user is really
re-created (prebuild == True), we already did `_cleanup_homedir()`.
The other thing is that before we `useradd`, we already have `/builddir`
pre-created -> so useradd doesn't create `.bashrc`, etc.  We can iterate
on this, but it doesn't seem to be Mock's responsibility.

I do not  get this. When you call mock with `--clean` you want to clean /builddir and with `-noclean` you want to preserve the content. Not just on the first run, but on consecutive runs.

mock/mock.spec
praiskup: This is an experiment (likely I should use `post1`) that I don't insist on,
but seems really attractive.

Simply, this way I can signalize that the Version is higher than the one
released in Fedora, always beating the NVR there (even if Release goes up due
to rebuilds) and in mock-core-configs I still can require the version newer
then the one from this PR.

praiskup:
> I do not get this. When you call mock with --clean you want to clean /builddir
> and with -noclean you want to preserve the content. Not just on the first run,
> but on consecutive runs.

I don't think [--no-clean](https://github.com/rpm-software-management/mock/blob/05cfc1185eda51650956b95d34b288cb30665f31/mock/py/mockbuild/buildroot.py#L469-L473) works this way.  That's executed unconditionally (note that `_make_build_user()` is always called for `prebuild == True`).  Which IMO implies that playing with `-r` option is just redundant (those files are already removed).

praiskup:
But honestly, I'm a bit confused, don't ask me what is the effect of `--no-clean` actually.

mock/mock.spec
praiskup: I moved this experiment into #1138, it was rather disturbing change in this PR.

praiskup:
So, the consequence here is that e.g. manual `/etc/passwd` modifications
are going to be dropped between `mock --shell` calls, because instead of
using `usermod` calls (old behavior), we are now newly re-creating the
from scratch (new behavior).  The new behavior can drop potentially
user-changed `/etc/{passwd,group}` entries.

Per documentation

        -n, --no-clean
               Do not clean chroot before building a package.
        ...
        --buildsrpm
               Build the specified SRPM either from  a  spec  file  and  source
               file/directory  or SCM. The chroot (including the results direc‐
               tory) is cleaned first, unless --no-clean is specified.

So keeping the old behavior for the "non build" use case seems like we
might be overcautious.  That said, the `--no-clean` option of course has
a meaning, but the thing users really want is to not clean up
the installed packages from the previous build (to potentially speed up
the subsequent build(s).

praiskup:
We know that @voxik and @mkoncek are heavy `--no-clean` users.  Folks, would you mind testing this not-yet-merged variant of mock?  Steps to install this:

    sudo dnf copr enable @mock/mock-pull-requests:pr:1103
    sudo dnf copr update mock mock-core-configs
    mock --scrub=all -r <your chroot>
    ... test the use-case you typically have ...

mkoncek:
Sure, can start tomorrow.
Just to be clear, my whole usage of `--no-clean` is:
* Build newer version of a dependency
* Install newer version of a dependency in mock
* Build newer version of a dependent package with `--no-clean`

I hope this is how it is intended to be used.

hobbes1069:
> Sure, can start tomorrow. Just to be clear, my whole usage of `--no-clean` is:
> 
> * Build newer version of a dependency
> * Install newer version of a dependency in mock
> * Build newer version of a dependent package with `--no-clean`

I use `--no-clean` via `fedpkg mockbuild` when I'm doing test builds of new versions of packages and the build fails for some reason. No point in reinstalling the whole buildroot.

praiskup:
Thank you for responding here!

@hobbes1069 wrote
> I use --no-clean via fedpkg mockbuild when I'm doing test builds of new versions of packages and the build fails for some reason. No point in reinstalling the whole buildroot.

Indeed, this behavior of --no-clean shouldn't be affected by this PR.

@fweimer-rh wrote
> The --shell option with the implied --no-clean still overwrites /etc/dnf/dnf.conf even before this change, which I found terribly confusing at one point.

Good point.  There already are other things that Mock sets up for every run no matter `--{,no-}clean`.  So this isn't precedent at least.

> But I think rewriting /etc/passwd is different because it reverts the
> the actions of postinst scripts of already-installed packages. I think
> this means that just re-running make in the otherwise unchanged
> buildroot re-entered via --shell can produce unexpected results. That
> can't be good?

Well, in theory at least (we may e.g. do `mock --shell` => `mock --install missinb-builddep` => `mock --shell` again).  But the original motivation was different dcad1667d16c1ee5c0be203f7253447fbe8dcdfe (it's not obvious, but the modified `else:` branch calling `_fixup_build_user` basically relates to `--shell`, not `--rebuild`, which btw did re-create the user even before, since 7814ebd4d074d5434b7984a28508d2a422aa156f).  Note mock always does other things like "chown -R <homedir>" so changing UID actually has some real effect.

xsuchy (approved):

Issue #539, opened 2020-03-15
mock-2.0+ regression mock logs everything twice.


bleve:
Piece of rust from epel7 build. Mock is epel7 version.

Building target platforms: x86_64
Building target platforms: x86_64
Building for target x86_64
Building for target x86_64
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.NhPvDw
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.NhPvDw
+ umask 022
+ umask 022
+ cd /builddir/build/BUILD
+ cd /builddir/build/BUILD
+ cd /builddir/build/BUILD
+ cd /builddir/build/BUILD
+ rm -rf rustc-1.42.0-src
+ rm -rf rustc-1.42.0-src
+ /usr/bin/xz -dc /builddir/build/SOURCES/rustc-1.42.0-src.tar.xz
+ /usr/bin/xz -dc /builddir/build/SOURCES/rustc-1.42.0-src.tar.xz
+ /usr/bin/tar -xf -
+ /usr/bin/tar -xf -
+ STATUS=0
+ STATUS=0
+ '[' 0 -ne 0 ']'
+ '[' 0 -ne 0 ']'
+ cd rustc-1.42.0-src
+ cd rustc-1.42.0-src

praiskup:
Similar report here https://bugzilla.redhat.com/show_bug.cgi?id=1805631

But I can not reproduce with default mock and default configs on centos7.
Is this specific to your package?  I tried `fedpkg clone rustc` and it failed.

Please upload your configuration, and minimal steps how to reproduce tihs.

praiskup:
Nevermind, reproduced.  I can't explain why I haven't seen it before - but
using `--resultdir /tmp/something` triggers the problem.

praiskup:
Note for myself, I have some progress on this in `fixup-logging` branch.

praiskup:
I re-flagged this to RFE, because the doubled-logs problem was
worked-around/fixed in master branch, but quite some work is needed to be
done on this issue.

Anyways, you shouldn't see doubled logs in mock v2.2.

praiskup:
See the bootstrap logging problems in #959, too.

PR #1106, opened 2023-06-09
mock: dump every package manager command to root log

praiskup:
Relates: https://pagure.io/releng/issue/11446

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -156,8 +156,11 @@ class Commands(object):
         return self.buildroot.make_chroot_path(*args)
 
     @traceLog()
-    def init(self, **kwargs):
+    def init(self, do_log=True, **kwargs):
         try:
+            if do_log:
+                self.buildroot.resetLogging()
+
             if self.bootstrap_buildroot is not None:
                 file_util.mkdirIfAbsent(self.buildroot.make_chroot_path())
                 self.bootstrap_buildroot.initialize(**kwargs)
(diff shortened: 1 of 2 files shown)


praiskup:
No, this doesn't help.  Bootstrap.pkg_manager commands are not dumped there.

mock/py/mockbuild/package_manager.py
```diff
@@ -245,16 +246,26 @@ def execute(self, *args, **kwargs):
...
 
             try:
+                str_invocation = " ".join(shlex.quote(arg) for arg in invocation)
+
                 # either it does not support --installroot (microdnf) or
                 # it is bootstrap image made by container with incomaptible dnf/rpm
+
+
                 if not self.support_installroot or self.is_bootstrap_image:
+                    getLog().info("Executing %s in chroot: %s", self.name,
```
xsuchy, line 256: try calling `self.buildroot.root_log.info()`

praiskup: Good idea!  That may lead to some success ... even though, in case of bootstrap this goes to `<chroot>-bootstrap/results/root.log` file, not `<chroot>/results/root.log`.

praiskup:
A new approach actually fixing #539; copy-pasting the git commit message:
```
    mock: revamp the logging logic
    
    Previously, when bootstrap was enabled, the buildroots were flipping the
    target log files all the time from '<root>-bootstrap/results' to
    '<root>/results' and back, anytime _resetLogging() was called.  This
    also means that the first log entry in the target chroot resultdir
    (<root>-bootstrap/results) was created too late, after the bootstrap
    chroot was fully initialized - loosing a lot of useful information.
    
    Now we never reset the logging for the bootstrap chroot.  This was
    always a wrong thing - both buildroot objects share the same
    self.root_log/self.build_log logging objects (same object id()).
    
    I renamed Buildroot._resetLogging() to Buildroot.resetLogging(), as it
    is actually a public method called from the outside.
    
    Now we call resetLogging(), and thus create the resultdir for logfiles
    much earlier — efore we even start initializing the bootstrap chroot.
    This makes the logs complete.
    
    Closes: #1106
    Relates: https://pagure.io/releng/issue/11446
    Fixes: #539
```

PR #1109, opened 2023-06-15
configs: fully qualified mageia cauldron image

praiskup:


xsuchy (approved):

praiskup:
The image is soo old, so actually `dnf update` fails there :-/ is there a better source?

praiskup:
There seems to be some controversy about the "official" state of the images: https://github.com/juanluisbaptiste/docker-brew-mageia/issues/30

PR #1112, opened 2023-06-15
mock: don't use dnf5 for bootstrap installation, yet

praiskup:
Seems like there's not yet the time, per: https://github.com/rpm-software-management/dnf5/issues/617
Reverts: 6c7cb9fcb248ef6ba7ab0559eefc02c4065e77f2.
Relates: #1089


praiskup:
Note that removing `dnf-3` binary from a DNF5-capable system is not a "trivial" task people would be doing routinely. Therefore, even if dnf5 is available on the system, dnf4 is likely around - so we can afford to revert this for a while.

xsuchy (approved):

Issue #1108, opened 2023-06-14
hw_log is run twice, once for buildroot, once for bootstrap


praiskup:
We should though run this only for buildroot.

xsuchy:
True. Plugins should have attribute whether they should be run in both roots or just in buildroot. For some of them it have sense to have them in both, but for others like hw log it have sense to run them only in buildroot.

PR #1114, opened 2023-06-15
hw_info: do nothing for bootstrap chroot

praiskup:
Fixes: #1108


xsuchy:
Hmm, it will work. But what about this one https://github.com/rpm-software-management/mock/pull/1116 ?

praiskup:
I wanted to delay the change similar to #1116 :) simply because some small risks discussed there, and this issue in particular is mostly cosmetics (since #1106 it becomes more apparent in `root.log`).  But let's move to #1116!

PR #1116, opened 2023-06-15
Rework plugins and do not run hw_info in bootstrap

xsuchy:


mock/py/mockbuild/plugin.py
```diff
@@ -62,7 +62,12 @@ def init_plugins(self, buildroot):
                     raise Error('Plugin version mismatch - requested = %s, current = %s'
                                 % (requested_api_version, current_api_version))
 
-                module.init(self, self.plugin_conf["{0}_opts".format(plugin)], buildroot)
+                if hasattr(module, 'run_in_bootstrap'):
+                    run_in_bootstrap = module.run_in_bootstrap
+                else:
+                    run_in_bootstrap = True
```
praiskup, line 68: `run_in_bootstrap = getattr(module, "run_in_bootstrap", True)`

mock/py/mockbuild/plugin.py
```diff
@@ -62,7 +62,12 @@ def init_plugins(self, buildroot):
...
+                if (buildroot.is_bootstrap and run_in_bootstrap) or not buildroot.is_bootstrap:
```
praiskup, line 69: This is equivalent:
```python
if not buildroot.is_bootstrap or run_in_bootstrap:
    ...
```

praiskup:
Thank you for the PR!  I like it, it would be good to merge soon so we can test heavily (I'm not 100% sure that not initializing the module is equivalent to what we were doing before, at least constructors of all plugins were always called).

Small nits in-code.  Plus I'd prefer changing the default to `run_in_boostrap = False` since the beginning, and **explicitly** set `run_in_bootstrap = True` for plugins.  I think it is much easier to spot that something **isn't executed in bootstrap**, than that **is** executed in bootstrap **without intentions and effects**.

xsuchy:
> Thank you for the PR! I like it, it would be good to merge soon so we can test heavily (I'm not 100% sure that not initializing the module is equivalent to what we were doing before, at least constructors of all plugins were always called).

If we want to be 100% compatible, then I can move the `if` check to `add_hook`. But I think this is better. No necessary initialization in constructor.

> 
> Small nits in-code. Plus I'd prefer changing the default to `run_in_boostrap = False` since the beginning, and **explicitly** set `run_in_bootstrap = True` for plugins. I think it is much easier to spot that something **isn't executed in bootstrap**, than that **is** executed in bootstrap **without intentions and effects**.

But that would be a possible breaking change if there is some 3rd party plugin we are not aware, that has to be loudly announced. I am fine with that though, if you are really sure.

praiskup (approved):
LGTM

PR #1121, opened 2023-06-21
Combine multiple if statements

zengwei2000:
Combine multiple if statements to reduce Duplicate code and improve readability

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -101,11 +101,11 @@ class Commands(object):
         try:
             self.plugins.call_hooks('clean')
             if self.bootstrap_buildroot is not None:
-                self.bootstrap_buildroot.plugins.call_hooks('clean')    
+                self.bootstrap_buildroot.plugins.call_hooks('clean')
             for scrub in scrub_opts:
                 self.plugins.call_hooks('scrub', scrub)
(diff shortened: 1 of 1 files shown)


praiskup:
Thank you for the PR, I appreciate your help here!

Ad first commit, I'm personally against merging white-space-only changes/re-formatting. The reason is simple (as in all old and conservative projects), such changes only complicate history searching with `git blame`.  For the formatting issues;  we try to concentrate on not creating new problems (i.e. having a CI check that would disallow us to create a line with a trailing whitespace would be a killing feature).

Ad second commit, I'm not quite sure what is the benefit.

zengwei2000:
> 

Personally, I feel that too many repetitive if statements affect the reading of code. These if statements can be combined into a more concise selection structure to reduce Duplicate code and improve readability

zengwei2000:
> 

Personally, I feel that too many repetitive if statements affect the reading of code. These if statements can be combined into a more concise selection structure to reduce Duplicate code and improve readability

xsuchy:
I am -1 as well.
This does not fix anything. While it removes repeating `if` it replaces it by storing functions into a hash and then calling a dictionary value as function. To me this obfuscation. And the new code is longer than the older one. :(

PR #1122, opened 2023-06-21
Adding a retry mechanism can increase the response to network request…

zengwei2000:
Adding a retry mechanism can increase the response to network request failures

PR #1123, opened 2023-06-21
Update file_downloader.py

zengwei2000:
I added two new parameters max_ Retries and retries_ The delay is 3 times and 10 seconds, respectively, representing the maximum number of retries and the waiting time between each retry. During the download process, a loop is used to retry, waiting for a period of time after each request fails before attempting the request until the maximum number of retries is reached or successful. If the maximum number of retries is reached and still fails, false is returned. In this way, we can increase the robustness of the code and cope with network instability or other abnormal situations.

### DIFF
--- a/mock/py/mockbuild/file_downloader.py
+++ b/mock/py/mockbuild/file_downloader.py
@@ -3,6 +3,7 @@
 
 import cgi
 import shutil
+import time
 import tempfile
 from urllib.parse import urlsplit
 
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/file_downloader.py
```diff
@@ -41,28 +42,31 @@ def get(cls, pkg_url_or_local_file):
...
+                pkg = cls.tmpdir + '/' + filename
+                with open(pkg, 'wb') as filed:
+                    for chunk in req.iter_content(4096):
+                        filed.write(chunk)
+                cls.backmap[pkg] = url
+                return pkg
+            except requests.exceptions.RequestException as err:
+                try_count += 1
+                log.error('Downloading error %s: %s (retrying...)', url, str(err))
+                time.sleep(2 ** try_count)
```
praiskup, line 68: Thank you for the update, but can you hire a library for this, e.g. `python3-backoff` (as discussed in #1122)?

zengwei2000: I tried using the Python 3 backoff package to solve this problem, but the Python 3 backoff package is not available and I don't know where to add it

praiskup: That is weird, `python3-backoff` is both available in EPEL 8 and Fedora.  Where do you miss it?

zengwei2000: https://github.com/rpm-software-management/mock/pull/1125   This is my submission using Python 3 backoff, but there will be errors during ci checking,  py/mockbuild/file_downloader.py:7: E0401[import-error]: Unable to import 'backoff'

PR #1130, opened 2023-06-23
Automatically handle Red Hat subscription in-container

praiskup:
When a RHEL container is started, the entitlement files are made available in-container and users can then 'dnf install' stuff inside. The entitlement directory though isn't /etc/pki/entitlement, but /etc/pki/entitlement-host.

Mock configuration currently expects that the files exist in /etc/pki/entitlement, so let's try to simplify the job to the end user, and copy the keys into the that directory.  Alternatively (is it needed?) we could create yet another entitlement directory specific to Mock, and change the PEM paths in mock-core-configs.

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -696,7 +696,7 @@ def main():
 
     sys.setrecursionlimit(config_opts["recursion_limit"])
 
-    util.subscription_redhat_init(config_opts)
+    util.subscription_redhat_init(config_opts, uidManager)
 
     # allow a different mock group to be specified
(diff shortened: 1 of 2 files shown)


xsuchy (approved):

Conan-Kudo:
Could we consider breaking out this RHSM integration code into its own Python module? I think it'd be useful for other tools that would need to do stuff like this (for example, KIWI, livecd-tools/appliance-tools, etc.)

cc: @schaefi

praiskup:
Why not, but moving to #1137.

PR #1133, opened 2023-06-26
mock: umount essential/bootstrap mountpoins on exception

praiskup:
Typically, when user pressed Ctrl+C during the long-running 'dnf install' command, KeyboardInterrupt was raised and unhandled, causing that the buildroot mounted in bootstrap stayed mounted (eventually breaking even subsequent `mock --scrub=all`).

While on it, stop using manual "mount_bootstrap" and use only the context manager.


xsuchy (approved):
+1

PR #1144, opened 2023-07-10
Use monorepo syntax for Packit builds

lachmanfrantisek:
Use monorepo syntax for Packit builds which should simplify the config.

More info:
* https://packit.dev/docs/configuration/#packages
* https://packit.dev/posts/monorepos/


xsuchy (approved):
+1

PR #1150, opened 2023-07-18
For now don't use --allowerasing with 'dnf5 remove'

praiskup:
Since https://github.com/rpm-software-management/dnf5/issues/461 is fixed, we can drop the very same work-around for 'builddep'.

Relates: #1149


xsuchy (approved):
+1

PR #1151, opened 2023-07-18
Provide <pkg_manager>_avoid_opts defaults as empty dict

praiskup:
Then, in real life configuration files we can assume it already is a dictionary and we can just:

    config_opts["dnf5_avoid_opts]["remove"] = ["--allowerasing"]

Instead of:

    config_opts["dnf5_avoid_opts] = {
        "remove": ["--allowerasing"],
    }


xsuchy (approved):
+1

Issue #1147, opened 2023-07-13
Set fedora-rawhide to use DNF5


xsuchy:
See
https://pagure.io/fesco/issue/3039

It is about time to set 

config_opts['package_manager'] = 'dnf5'

for fedora rawhide configs.

praiskup:
One of the issues tracked in #894 seems like blocker: https://github.com/rpm-software-management/dnf5/issues/617

praiskup:
Related [Fedora change proposal](https://pagure.io/fesco/issue/3096) and [discussion](https://discussion.fedoraproject.org/t/f40-change-proposal-build-with-dnf5-self-contained/94937).

PR #1152, opened 2023-07-18
Install Rawhide (39+) buildroot with dnf5

praiskup:
The Fedora 39 container images have 'dnf5' and 'dnf5-plugins' installed, so we don't actually have to install anything into the bootstrap now, we mark it "ready" to speedup the builds.

Fixes: #1147


praiskup:
TODO: after #1168 we should re-order the priority (to prefer dnf5 over dnf4).

praiskup:
Closing this till Rawhide is Fedora 41.

PR #1158, opened 2023-07-22
mockbuild/buildroot: Make the FUSE device available if supported

Conan-Kudo:
Without the FUSE device, we cannot use FUSE filesystems with image builds.

This is particularly important for RPM-OSTree builds that use FUSE as part of the image build process.

Reference: https://pagure.io/fedora-infrastructure/issue/11420

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -121,6 +121,7 @@ class Buildroot(object):
         self._homedir_bindmounts = {}
         self._setup_nspawn_btrfs_device()
         self._setup_nspawn_devicemapper_device()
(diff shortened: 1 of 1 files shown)


travier (approved):
Not tested and not a mock developer but I approve this change :)

This does it for nspawn. Should we do it for the old chroot method as well (or is that just one thing now)?

xsuchy (approved):
+1

Conan-Kudo:
> This does it for nspawn. Should we do it for the old chroot method as well (or is that just one thing now)?

Old chroot gets all of `/dev` mounted in already.

praiskup:
thank you

Issue #1155, opened 2023-07-20
Mock should use `dnf --setopt=allow_vendor_change=yes`


praiskup:
This config is going to be switched to `allow_vendor_change=no`  by default in DNF5 (as discussed e.g. in https://github.com/rpm-software-management/dnf5/issues/738), and I tend to think we should switch the default to OFF in Mock. The typical thing is to use `--addrepo` to provide replacement packages, or with `mock --chain` the local repo with results. Also Mock in Copr e.g. builds "override" packages with a different Vendor tag.

PR #1160, opened 2023-07-24
Use 'allow_vendor_change=yes' with DNF5

praiskup:
The `allow_vendor_change=no` option is going to be set in DNF5 as default: https://github.com/rpm-software-management/dnf5/issues/712

With Mock though, it is very common to override the distro default package with a custom one (newer or fixed one?).  And blindly trying to keep the Vendor tag unchanged for the overridden packages would be very complicated, without any real benefits.

Note `allow_vendor_change=no` only affects the cases when some package in the buildroot/bootstrap needs to be _updated_ while _typically_ all the packages are only _installed_ at the beginning and are already up2date.  This is therefore mainly related to the --chain option, or cases like `bootstrap_image_ready=False`, `update_before_build`, etc.

Fixes: #1155

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -487,7 +487,7 @@
 
 # DNF (or DNF4, Python) - https://github.com/rpm-software-management/dnf
 #config_opts['dnf_command'] = '/usr/bin/dnf-3'
-#config_opts['dnf_common_opts'] = ['--setopt=deltarpm=False', '--allowerasing']
+#config_opts['dnf_common_opts'] = ['--setopt=deltarpm=False', '--setopt=allow_vendor_change=yes', '--allowerasing']
 # DF5 sub-command 'builddep' doesn't support the '--allowerasing' option:
 # https://github.com/rpm-software-management/dnf5/issues/461
 #config_opts["dnf5_avoid_opts"] = {"builddep": ["--allowerasing"]}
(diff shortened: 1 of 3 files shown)


xsuchy (commented):
Why this way? This option exists in dnf.conf in DNF4 too. Can we rather put it in configs in mock-core-configs?

Conan-Kudo:
> > Why this way?
> 
> Because it is the simplest way how to do this (one-liner, instead of multiple configs in the future) and still quite safe to be set by default. I didn't want to affect DNF4 case as there's probably no existing distribution turning the feature ON.

It's on by default in openSUSE.

praiskup:
Interesting, it looks like this has rather a low-severity hit, because we haven't noticed
this so far even though openSUSE chroots are enabled in Fedora Copr.  Lemme ask
explicitly, should we have the option in `dnf_common_options`, too?

praiskup:
Ok, updated, why "blocked"?

mock/docs/site-defaults.cfg
```diff
@@ -499,7 +499,7 @@
 # Known issue with DNF5: --forcearch is not supported by DNF5
 # https://github.com/rpm-software-management/dnf5/issues/112
 #config_opts['dnf5_command'] = '/usr/bin/dnf5'
-#config_opts['dnf5_common_opts'] = ['--setopt=deltarpm=False', '--allowerasing']
+#config_opts['dnf5_common_opts'] = ['--setopt=deltarpm=False', '--setopt=allow_vendor_change=yes', '--allowerasing']
```
Conan-Kudo, line 502: Does `yes` work instead of `True`?

Conan-Kudo (changes requested):

Conan-Kudo (approved):

mock/docs/site-defaults.cfg
Conan-Kudo: Okay, then. Never seen that before. 😅

PR #1163, opened 2023-07-26
TOCTOU in mkdirIfAbsent

praiskup:
Do not check for directory existence before calling os.makedirs(), this brings no benefit (and because of the race, we anyway already ignore the EEXISTS)

### DIFF
--- a/mock/py/mockbuild/file_util.py
+++ b/mock/py/mockbuild/file_util.py
@@ -16,14 +16,13 @@ from .trace_decorator import getLog, traceLog
 def mkdirIfAbsent(*args):
     for dirName in args:
         getLog().debug("ensuring that dir exists: %s", dirName)
-        if not os.path.exists(dirName):
-            try:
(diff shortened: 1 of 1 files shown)


xsuchy:
-1 call me old-fashioned, but I like checking the state before doing the work. Even when try/except handle that.

praiskup:
Can you elaborate on the benefits? This is so obvious antipattern in all the languages around that I don't even know to point at.  OTOH, I won't fight too much.  I just faced some weird race situation while testing today, so I came across this one. But still failing even with this patch.

jafd:
All it takes is just one rogue process (with or without a bug) removing the path just after your `os.path.exists()` call has returned `True`.

At this point I have to note that in my practice, even if you have a "highly unlikely" possibility for a race, there _will_ come a system utilization high enough that it will be happening once every 30 minutes or so, while "virtually unlikely" races will be happening once every two hours.

xsuchy (approved):
ok then +1

PR #1166, opened 2023-07-27
Bootstrap inner mount orphan kill

praiskup:


FrostyX (approved):
I am not sure how to reproduce the issue and test the patch properly. But it looks reasonable to me, and it fixes tests, so I would go for it.

praiskup:
Reproducing this is relatively easy:
```
$ mock --postinstall --isolation=simple https://github.com/rpm-software-management/mock-test-data/raw/main/daemontest-1-0.src.rpm
... successful build, mock exists ...
$ pgrep daemontest -a
41652 daemontest PID=41652
```

PR #1167, opened 2023-07-27
bootstrap_image and the local repositories

praiskup:
We can not use user_mountpoints because these aren't (historically, not
    sure what are the reasons) mounted too late during the Buildroot
    preparation (after the self._init_pkg_management()).  This isn't
    actually an issue when 'use_bootstrap_image == False' because we use the
    package manager on host (so it actually sees the repository on host).
    
    So newly we place the local_repo mount points into the "managed"
    category, which seems even semantically better now.  We though need to
    call the 'self.mounts.mountall_managed()' method a little bit later than
    before, when the local repo BindMountPoints are already defined by
    self.pkg_manager.initialize().

    This problem has been triggered by the LVM test-case that creates a
    custom local_repo (on host) with `baseurl=/tmp/foo` that needs to be
    bind-mounted into the bootstrap chroot properly.

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -241,14 +241,18 @@ class Buildroot(object):
         self._setup_devices()
 
         self._setup_files()
-        self.mounts.mountall_managed()
 
         # write out config details
         self.root_log.debug('rootdir = %s', self.make_chroot_path())
         self.root_log.debug('resultdir = %s', self.resultdir)
 
         self.set_package_manager()
+
+        # this creates some managed mounts
         self.pkg_manager.initialize()
+
+        self.mounts.mountall_managed()
+
         self._setup_resolver_config()
         self._setup_katello_ca()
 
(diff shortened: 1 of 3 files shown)


praiskup:
> PYTHONPATH=/home/jkadlcik/git/mock/mock/py \

Wow, I was never able to run Mock like that, I always do `tito build --test --rpm --install` first.  Nice.

> So I am probably misunderstanding the issue?

I bet this is related to `--isolation=simple`, i.e. the mock-in-container case.  Can you please try to re-try?
The problem has been triggered by one of these cases:
https://github.com/rpm-software-management/mock/blob/6c7efb6535389a463ec7c0d2e2ac0529fc5caa58/mock/integration-tests/24-postinstall.tst#L18-L19

praiskup:
You might need to `--scrub=all` first actually.

praiskup:
I updated the commit message with the info on how this can be triggered.

FrostyX (approved):

PR #1168, opened 2023-07-27
Search for 'dnf5' in fedora:rawhide image

praiskup:
We still prefer doing the Rawhide builds with DNF4 over DNF5, at least till https://github.com/rpm-software-management/dnf5/issues/617 is fixed.

But DNF4 is not part of the 'fedora:rawhide' image nowadays, and to install it we actually have to use the 'dnf5' installer there.


praiskup:
Updated - when both DNF4 and DNF5 is available, we still prefer trying DNF4 first to install microdnf or yum.

FrostyX:
It is curious to me that we need this PR because AFAIK there is always an alias like this:

```
[root@5f9fc4eb5a86 /]# ls -l /usr/bin/dnf*
lrwxrwxrwx. 1 root root       4 Jun 30 00:00 /usr/bin/dnf -> dnf5
-rwxr-xr-x. 1 root root 1148216 Jun 30 00:00 /usr/bin/dnf5
```

That being said, explicitly listing `dnf5` is fine.

praiskup:
Mock always works either with /bin/dnf-3 or /bin/dnf5, because /bin/dnf is ambiguous across distributions.
The "fallbacks" dictionary is there for the cases when the desired package manager isn't found (yet).  We need this PR for three use-cases:

- user runs Mock on Fedora Rawhide system where's no DNF4 at all (mock needs to use dnf5 on-host to install dnf4 based bootstrap chroot)
- user runs Mock with "use-bootstrap-image" with `-r fedora-rawhide-x86_64`, the `fedora:rawhide` podman image by default contains `dnf5` now (this is being reverted right now for a while, but one day we'll be there anyway).  In this case, Mock is configured to use `dnf4` for Rawhide, in bootstrap is `dnf5` so we need to try `dnf5` command to install dnf4 
- when bootstrap is OFF, and we use host's DNF5 command to install YUM or DNF4 based chroots

praiskup:
In the context of this PR, string `dnf` represents just an "ID" of one of the classes in `package_manager.py` file (not a command name or alike).

Issue #1110, opened 2023-06-15
Always use a native "bootstrap" chroot, even for --forcearch builds


praiskup:
On x86_64, for the i{3,5}86 builds, we need to clone 32bit containers or fail early with a reasonable error if such image doesn't exist.  The same needs to be done for --forcearch.

That's because Mock by default installs `dnf.conf` with repositories designed for a different architecture.  IOW, `dnf.conf` copied into the bootstrap is always the same as for the target buildroot.  Doing `dnf install dnf-plugins-there`  (to install `dnf builddep` command) can easily fail (trying to install i386 only packages into the x86_64 bootstrap).

https://www.redhat.com/sysadmin/specify-architecture-pulling-podman-images

We need some reasonable mapping from Mock arch to Podman arch (e.g. `x86_64 => amd64`) or a config option like `bootstrap_image_architecture`.

praiskup:
We can implement this with `podman pull <image-fqdn> --arch <arch>`, though this overrides the local image tag that might be used later on by the end-user and it might lead to confusion, see https://github.com/containers/podman/issues/18896

praiskup:
The https://github.com/containers/podman/issues/18896 might take some time, and the new feature would require the latest-greates podman version anyway;  seems like we should go with skopeo as discussed there.

praiskup:
The bug we currently have is at least for "multilib" builds, which are not really cross-arch.

The image is x86_64, but the repository is i686 - bootstrap inherits from buildroot.  Trying to install `dnf3-plugins-core` etc. ends up with:
```
...

  file /usr/bin/gapplication from install of glib2-2.76.4-3.fc38.i686 conflicts with file from package glib2-2.76.3-1.fc38.x86_64
  file /usr/bin/gdbus from install of glib2-2.76.4-3.fc38.i686 conflicts with file from package glib2-2.76.3-1.fc38.x86_64
  file /usr/bin/gio from install of glib2-2.76.4-3.fc38.i686 conflicts with file from package glib2-2.76.3-1.fc38.x86_64
  file /usr/bin/glib-compile-schemas from install of glib2-2.76.4-3.fc38.i686 conflicts with file from package glib2-2.76.3-1.fc38.x86_64
  file /usr/bin/gsettings from install of glib2-2.76.4-3.fc38.i686 conflicts with file from package glib2-2.76.3-1.fc38.x86_64
[...]

PR #1172, opened 2023-07-28
Fix multilib builds with --use-bootstrap-image

praiskup:
When building an i?86 package (with i?86 chroot config) on an x86_64 host, we report to kernel that we work in 32 bit mode (using the '/bin/setarch'-like condPersonality() call).  This is completely OK and desired for the build (target) chroot, but not for the bootstrap chroot which is generated from a container image.

The container image has always a host-native arch (unless there's a configuration error).  So the pre-installed packages are x86_64. When we want to install additional packages there (typically to provide the 'dnf builddep' command in bootstrap), and we do that using the package manager already installed in bootstrap image (no package manager on host is required!), the package manager is confused by the 32-bit mode and resolves '$basearch' into 'i386' instead of 'x86_64'.  So we need to switch back to 64-bit mode (in a subprocess, preexec_fn).

Note that a late call to `os.uname()` would be confused by the condPersonality() call.  So newly we rather remember the 'host_arch' at the beginning in config_opts, before the initial call to condPersonality().

Fixes: #1110

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -521,7 +521,7 @@ def check_arch_combination(target_arch, config_opts):
         legal = config_opts['legal_host_arches']
     except KeyError:
         return
-    host_arch = os.uname()[-1]
+    host_arch = config_opts['host_arch']
     if (host_arch not in legal) and not config_opts['forcearch']:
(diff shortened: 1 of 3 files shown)


FrostyX (approved):

praiskup:
Thank you for the review!

PR #1173, opened 2023-07-28
Image configuration fixes for use_bootstrap_image = True

praiskup:


mock-core-configs/etc/mock/templates/navy-8.tpl
```diff
@@ -4,6 +4,7 @@ config_opts['releasever'] = '8'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['description'] = 'Navy Linux {{ releasever }}'
+config_opts['bootstrap_image'] = 'docker.io/navylinux/navylinux:latest'
```
FrostyX, line 7: This works for now

```
[root@8ffc9ef52bcf /]# cat /etc/os-release |grep ^VERSION*
VERSION="8.4 (Base)"
VERSION_ID="8"
```

but `latest` can easily change from 8 to 9 in the future. Maybe put a FIXME here, so that somebody asks the container maintainer to tag a specific version?

praiskup: Good point, fixed.

FrostyX (approved):

PR #1176, opened 2023-08-09
New [mock] release [5.0-1] and [mock-core-configs] release [39-1]

### LINKED ISSUE #1177: New version of mock-core-configs needs tagging for F39 branch
The changes for F39 branching have been made in git, but no release is tagged yet. We need a release tagged upstream and builds done downstream. Thanks!

praiskup:
Created by command (mock):
/usr/bin/tito tag --use-version=5.0

Created by command (mock-core-configs)
/usr/bin/tito tag --use-version=39

Fixes #1177


praiskup:
@Conan-Kudo I'd prefer doing this tomorrow, Fedora is being branched right now (already requested by @hroncok). I think we can have yet another release for mock-core-configs the very next day, if needed.

AdamWill:
thanks! next time it'd be great if this could land at the same time as branching (as we try to do for fedora-repos), because without it, I have to do [hacks in the openQA tests](https://pagure.io/fedora-qa/os-autoinst-distri-fedora/c/8483e7bd82f038cd669b164dd2cecc1ec0f7bf67?branch=main) to make the live image build tests work.

We could end up in a similar chicken/egg situation as we do with fedora-repos (we need the new build to go in the compose, but the tests for the build will fail until it's in the compose...), but hey, we already have to deal with that for fedora-repos, this would just be one more case. @nirik for info

PR #1192, opened 2023-08-23
Drop unused Podman methods

praiskup:


lachmanfrantisek:
@praiskup the `Response is not in JSON format, there is probably a bug in the API code.` comes from copr, right?

I am thinking about multiple things how to improve this:
* Try multiple times: not sure if this can't have any drawbacks
* Suggest rerunning (`/packit build`) if the person thinks it's a flake or resolved the issue.
* Try to get a reason from the API. (I have a deja vu we've been playing with this very message.)

xsuchy (approved):
+1

PR #1195, opened 2023-08-23
override ppc64le %_host_cpu

praiskup:
This is what Koji does for all architectures.  The macro is used by
rpmbuild while matching 'ExclusiveArch' patterns.  E.g. "%java_arches"
then contains 'ppc64le', but by default '_host_cpu' contains powerpc64le
on Fedora.

RPM checks %_target_cpu with %_build_cpu fallback:
https://github.com/rpm-software-management/rpm/blob/21457de886faf2415500a8bb7cc6c816d72939ef/build/parsePreamble.c#L442-L449

For 'BuildArch: noarch' packages, %_target_cpu is 'noarch', fallback is
used and '%_build_cpu' is defined as '%_host_cpu' in redhat-rpm-config.
This is set by 'config.guess' to powerpc64le.

TODO: submit a redhat-rpm-config issue
TODO: resolve other EPEL/Fedora targets
Relates: https://github.com/fedora-copr/copr/issues/2870
Relates: https://bugzilla.redhat.com/show_bug.cgi?id=1461288

### DIFF
--- a/mock-core-configs/etc/mock/centos-stream+epel-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-8-ppc64le.cfg
@@ -6,3 +6,5 @@ config_opts['root'] = 'centos-stream+epel-8-ppc64le'
 config_opts['description'] = 'CentOS Stream 8 + EPEL'
 config_opts['target_arch'] = 'ppc64le'
 config_opts['legal_host_arches'] = ('ppc64le',)
+# see pull-request #1195
+config_opts['macros']['%_host_cpu'] = 'ppc64le'
--- a/mock-core-configs/etc/mock/centos-stream+epel-9-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-9-ppc64le.cfg
@@ -6,3 +6,5 @@ config_opts['root'] = 'centos-stream+epel-9-ppc64le'
 config_opts['description'] = 'CentOS Stream 9 + EPEL'
 config_opts['target_arch'] = 'ppc64le'
 config_opts['legal_host_arches'] = ('ppc64le',)
+# see pull-request #1195
+config_opts['macros']['%_host_cpu'] = 'ppc64le'
--- a/mock-core-configs/etc/mock/centos-stream+epel-next-8-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-next-8-ppc64le.cfg
@@ -8,3 +8,5 @@ config_opts['root'] = 'centos-stream+epel-next-8-ppc64le'
 config_opts['description'] = 'CentOS Stream 8 + EPEL Next'
 config_opts['target_arch'] = 'ppc64le'
 config_opts['legal_host_arches'] = ('ppc64le',)
+# see pull-request #1195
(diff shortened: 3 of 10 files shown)


pmatilai:
Before trodding deeper into these dark woods, it would be good to check what the current state of things is in rpm >= 4.19 where config.guess has nothing to do with anything (because cmake is used).

praiskup:
Thank you for your comment, @pmatilai.

OK, but that's not even in Rawhide, right?  I'd like to fix builds for EPEL7+.

Otherwise yes, 4.19 [seems to](https://github.com/rpm-software-management/rpm/blob/21457de886faf2415500a8bb7cc6c816d72939ef/CMakeLists.txt#L140) use [CMAKE_HOST_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html), which is `uname -m`, which should be fine (`ppc64le` returned and set as %_host_cpu, both on our Power9 and Power8 machines).  The question is how realistic is to fix the macro in **old** buildchroots so Mock correctly builds them.

E.g. %{java_arches} expands to just `aarch64 ppc64le s390x x86_64`. And this clearly looks like a Koji-targeted list (%_cpu_host is very predictable in Koji).  Compare that with %go_arches `i386 i486 i586 i686 pentium3 pentium4 athlon geode x86_64 armv3l armv4b armv4l armv4tl armv5tl armv5tel armv5tejl armv6l armv6hl armv7l armv7hl armv7hnl armv8l armv8hl armv8hnl armv8hcnl aarch64 ppc64le s390x mips mipsel mipsr6 mipsr6el mips64 mips64el mips64r6 mips64r6el` which is much more interesting.  Sadly, `powerpc64le` is missed even there!

As noted above, Koji overrides `%_host_cpu`. Is it OK to do that in Mock by default?

praiskup:
@pmatilai do you think this patch is a wrong idea?

praiskup:
Per https://github.com/rpm-software-management/rpm/issues/791 it seems this is going to be fixed for Fedora 39+ with rpm 4.19.0.

Since we know the preferred fix (`%_host_cpu=ppc64le`, not `powerpc64le`), overriding this macro in Mock for older chroots seems quite OK to me (considering Koji does this).  @pmatilai wdyt?

Conan-Kudo:
I'm not @pmatilai, but I think that is reasonable. But does this behave reasonably when we do foreign arch mock runs (e.g. ppc64le on x86_64)?

praiskup:
> But does this behave reasonably when we do foreign arch mock runs (e.g. ppc64le on x86_64)?

I guess?  The `%_host_cpu` for target ppc64le would be unconditionally `ppc64le`, no matter on what system Mock is executed.  The current behavior is similar, no matter the host system (even if x86_64):

```
$ mock -r epel-8-ppc64le --shell
<mock-chroot> sh-5.1# rpm --eval %_host_cpu
powerpc64le
```

praiskup:
Trying to be brave, and marking this PR for review (see the commit message for the info).  I still eventually believe this is worth merging (but it only affects EL 9 and older, all actively developed Fedora versions are OK).

xsuchy (approved):

PR #1196, opened 2023-08-28
podman: better "podman pull / image mount" error reporting

praiskup:
Defining a new BootstrapError() exception covering the bootstrap chroot preparation problems.

Make sure that stderr output is dumped into logs when some failure happens.

When `podman pull` fails, we just throw a logger.error() message now, and delay the fatal failure till the `podman image mount` attempt. Users might have the image pre-downloaded.


xsuchy (approved):
+1

PR #1197, opened 2023-08-28
Add int_src_dir option to specify source directory

pastalian:
This allows users to use repositories with non-standard structures.

There is already a 'spec' scm option to specify a SPEC file in arbitrary directories. Why not allow arbitrary source directories to be specified as well?

### DIFF
--- a/mock/py/mockbuild/scm.py
+++ b/mock/py/mockbuild/scm.py
@@ -76,6 +76,7 @@ class scmWorker(object):
         self.spec = opts['spec']
         self.spec = self.spec.replace("SCM_PKG", self.pkg)
 
+        self.int_src_dir = opts['int_src_dir']
         self.ext_src_dir = opts['ext_src_dir']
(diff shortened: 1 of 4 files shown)


mock/py/mockbuild/scm.py
```diff
@@ -147,6 +148,11 @@ def prepare_sources(self):
             sys.exit(5)
         self.spec = sf
 
+        # Use specified sources directory
+        sources_dir = self.src_dir
+        if self.int_src_dir:
+            sources_dir += "/" + self.int_src_dir
```
praiskup, line 154: Can you use `sources_dir = os.path.join(sources_dir, self.int_src_dir)`?

praiskup:
Otherwise looks good, thank you.

mock/py/mockbuild/scm.py
pastalian: Fixed (and accidentally rebased).

PR #1205, opened 2023-08-31
podman: warn that 'docker run --privileged' may be needed

praiskup:
This complements 79c36e81be4535b8e45bd33ac95a5b98b2951e37 with useful info, Mock already falls-back to using "normal bootstrap" if in Docker without --privileged.

Fixes: #1184


Conan-Kudo (approved):

praiskup:
Thank you for looking!

Issue #1206, opened 2023-09-01
Specifying `module_enable` config option via the command line errors


yavor-atanasov:
### Short description of the problem
Specifying `module_enable` config option via the command line seems to cause an error.

### Output of `rpm -q mock`

```
mock-5.0-1.el8.noarch
```

### Steps to reproduce issue
This is the command I'm running. The contents of the specfile and the sources is irrelevant, because mock fails when it tries to setup the build environment

```
/usr/bin/mock --isolation=simple --config-opts module_enable=nodejs:18 --root /etc/mock/rocky+epel-8-aarch64.cfg  --no-bootstrap-image --buildsrpm --spec SPECS/hello-rpm.spec --sources SOURCES --resultdir SRPMS
```

The error this produces is:
```
ERROR: Exception(SPECS/hello-rpm.spec) Config(rocky+epel-8-aarch64) 1 minutes 3 seconds
INFO: Results and/or logs in: SRPMS
INFO: Cleaning up build root ('cleanup_on_failure=True')
Start: clean chroot
Finish: clean chroot
ERROR: can only concatenate list (not "str") to list
Traceback (most recent call last):
  File "/usr/libexec/mock/mock", line 1087, in <module>
    exitStatus = main()
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 847, in main
    result = run_command(options, args, config_opts, commands, buildroot, state)
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 966, in run_command
    mockbuild.rebuild.do_buildsrpm(config_opts, commands, buildroot, options, args)
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.6/site-packages/mockbuild/rebuild.py", line 105, in do_buildsrpm
    cmd=cmd, post=None, clean=clean)
  File "/usr/lib/python3.6/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.6/site-packages/mockbuild/rebuild.py", line 24, in rebuild_generic
    commands.init(prebuild=not config_opts.get('short_circuit'))
[...]

PR #1208, opened 2023-09-06
man: better --config-opts array specification

praiskup:
Closes: #1206


Conan-Kudo (approved):

Conan-Kudo:
We should consider changing the source format of this man page into something more maintainable (like rst or md)...

praiskup:
> We should consider changing the source format of this man page into something more maintainable (like rst or md)...

It should be moved into the `argparse` --help output, then we could use `argparse-manpage` for generating manual pages.

Issue #1183, opened 2023-08-13
Mock 5 broke --install from a clean chroot


QuLogic:
### Short description of the problem

Using `mock --install` fails because of a missing result directory.

```
$ mock -r fedora-rawhide-x86_64 --install python3-tables
INFO: mock.py version 5.0 starting (python version = 3.11.4, NVR = mock-5.0-1.fc38)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
ERROR: Could not create dir /var/container/mock//fedora-rawhide-x86_64/result. Error: [Errno 13] Permission denied: '/var/container/mock//fedora-rawhide-x86_64/result'
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/mockbuild/file_util.py", line 22, in mkdirIfAbsent
    os.makedirs(dirName)
  File "<frozen os>", line 225, in makedirs
PermissionError: [Errno 13] Permission denied: '/var/container/mock//fedora-rawhide-x86_64/result'
ERROR:
Could not create output directory for built rpms. The directory specified was:
    /var/container/mock//fedora-rawhide-x86_64/result

Try using the --resultdir= option to select another location. Recommended location is --resultdir=~/mock/.
```

It was not previously necessary to set a `--resultdir` in mock <5, and I don't have any use for one. (Note that I have configured mock to use a different storage path, but this has otherwise been fine before, and works fine with the usual `fedpkg mockbuild`.)

### Output of `rpm -q mock`

mock-5.0-1.fc38.noarch

### Steps to reproduce issue

1. mock -r fedora-rawhide-x86_64 --clean
2. mock -r fedora-rawhide-x86_64 --install python3-tables

### Any additional notes

<details>
  <summary>
    <p>Output of `mock --debug-config`</p>
  </summary>

```
INFO: mock.py version 5.0 starting (python version = 3.11.4, NVR = mock-5.0-1.fc38)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
config_opts['additional_packages'] = []
config_opts['basedir'] = '/var/container/mock/'
config_opts['bootstrap_image'] = 'registry.fedoraproject.org/fedora:rawhide'
[...]

PR #1209, opened 2023-09-08
Assure basedir exists before creating resultdir

praiskup:
Because resultdir is, by default, created below basedir.  This has been breaking use-cases started with commands.init(do_log=True).

The bug was caused by f7b0c9bf18c46046be922adc740d1c6616ae0cc6, where we started pre-creating resuldir sooner than before so we could start the logging early.

Fixes: #1183
Complements: f7b0c9bf18c46046be922adc740d1c6616ae0cc6

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -4,6 +4,7 @@
 from contextlib import contextmanager
 import errno
 import fcntl
+import functools
 import glob
 import grp
 import logging
@@ -37,6 +38,16 @@ def noop_in_bootstrap(f):
             getLog().debug("method {} skipped in bootstrap".format(f.__name__))
             return
         return f(self, *args, **kwargs)
+    return wrapper
+
+
+def call_just_once(f):
+    """ Assure that method returns None, and is called just once """
+    @functools.lru_cache(maxsize=None)
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        if f(*args, **kwargs) is not None:
+            raise RuntimeError(f"Method '{f.__name__}' can not return values")
     return wrapper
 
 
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

mock/py/mockbuild/buildroot.py
```diff
@@ -40,6 +40,18 @@ def wrapper(self, *args, **kwargs):
...
 
 
+def call_just_once(f):
+    """
+    Make sure that an object method is called just once.
+    """
+    def wrapper(self, *args, **kwargs):
+        if hasattr(self, "called_method_retval"):
+            return self.called_method_retval
+        self.called_method_retval = f(self, *args, **kwargs)
```
QuLogic, line 50: This is a bit odd; you're saving the return value on `self`, which means that there is only ever one return value saved regardless of how many functions are decorated. It's also given a public naming convention even though it's not public.

mock/py/mockbuild/buildroot.py
```diff
@@ -40,6 +40,18 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def call_just_once(f):
+    """
+    Make sure that an object method is called just once.
+    """
+    def wrapper(self, *args, **kwargs):
```
QuLogic, line 47: Best practice would be to decorate this with `functools.wraps`.

mock/py/mockbuild/buildroot.py
QuLogic: Also odd is that this passes through `*args, **kwargs`, but that is not used to differentiate return values.

praiskup: Good point. This would be very tricky in the future, thank you for the review! TIL

PR #1213, opened 2023-09-12
mounts: make the hide-from-host mountpoints root-only

praiskup:
We don't need to allow anyone writing there, these mountpoints are just hiding the complicated recursive bind-mounts in Mock's separate namespace.

### DIFF
--- a/mock/py/mockbuild/mounts.py
+++ b/mock/py/mockbuild/mounts.py
@@ -205,7 +205,7 @@ class Mounts(object):
                 FileSystemMountPoint(filetype='tmpfs',
                                      device=device,
                                      path=host_path,
-                                     options="rprivate"),
+                                     options="rprivate,mode=0755"),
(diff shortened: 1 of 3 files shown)


mock/py/mockbuild/mounts.py
```diff
@@ -205,7 +205,7 @@ def __init__(self, rootObj):
                 FileSystemMountPoint(filetype='tmpfs',
                                      device=device,
                                      path=host_path,
-                                     options="rprivate"),
+                                     options="rprivate,mode=0755"),
```
xsuchy, line 208: Did you meant 0700 ?

praiskup: I didn't want to disallow anyone from stepping into the directory, but not sure.

xsuchy (approved):
+1

Issue #42, opened 2017-02-16
postinstall option and mock return code if failure


morucci:
Hi,
I expected to have mock returns an error code when the install built packages step fails but it isn't the case apparently. I haven't found any option to make mock fails instead of returning a warning + exit 0. I ends up with the solution to grab logs from root.log but I don't really like that solution.
I'm wondering if this is something you want to change in the future or if the actual behavior will remains. Should I propose a patch ?

Thanks in advance for your response.

Frodox:
Do you mean 
when 
```
config_opts['chroot_setup_cmd'] = 'install bash coreutils gcc gcc-c++...'
```
fails to install smthing?

morucci:
No, when I'm building a package in mock:
mock --rebuild --postinstall --resultdir artifacts stuff-0.6-1.el7.centos.src.rpm

The postinstall option will attempts to install built packages inside the chroot. My concern is only a warning in log is reported by mock when it try and fails to install built packages.

xsuchy:
Yum is very bad in returing status code. You may try to persuade that in https://bugzilla.redhat.com/show_bug.cgi?id=1107742  I gave up.
Yum will return 0 even if scriplet fail, if anything in transaction fails, it still exit with 0. It must be error in yum itself to return something non zero.

Anyway. There is nothing I can do about this on Mock side. Sorry.

praiskup:
Reopening per suggestion in #1207.

frenzymadness:
Proposed solution (with my zero knowledge of the codebase):

If we re-raise or skip catching exceptions during the installation here:

https://github.com/rpm-software-management/mock/blob/49c9cb6014e47ead6ab764de7ccadec1bb2b4404/mock/py/mockbuild/backend.py#L829-L833

it should be properly handled in rebuild_generic where `install_build_results` is called as `post()`:

https://github.com/rpm-software-management/mock/blob/49c9cb6014e47ead6ab764de7ccadec1bb2b4404/mock/py/mockbuild/rebuild.py#L17-L48

PR #1216, opened 2023-09-12
Fail mock process if DNF in --postinstall fails

praiskup:
The Answer to the Ultimate Question of Life, the Universe, and Everything is #42.


xsuchy (approved):
+1

PR #1218, opened 2023-09-12
podman: redirect 'podman pull' stdout to stderr

praiskup:
This is to fix the 27-nspawn.tst which again detects an unwanted stdout output from Mock.

Follow-up-for: b626045d2bdde026e0278dd1971ed54b217cc7af

### DIFF
--- a/mock/py/mockbuild/podman.py
+++ b/mock/py/mockbuild/podman.py
@@ -101,7 +101,7 @@ class Podman:
         finally:
             logger.info("umounting image %s (%s) with podman image umount",
                         self.image, mountpoint)
-            subprocess.run(cmd_umount, check=True)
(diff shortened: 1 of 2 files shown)


praiskup:
The `do` and `do_with_status` methods are out of our control;  we should reorganize it.  (And perhaps unit-test it and create a separate library for it)

praiskup:
Still the same output, even with:
```
DEBUG: Executing command: /usr/bin/podman pull registry.fedoraproject.org/fedora:38 >&2 ....
```
What am I missing? :-(

xsuchy (approved):
+1

praiskup:
Self-merging.  Trivial enough and blocks the release.

PR #1220, opened 2023-09-13
Suggests DNF5 installation on 38+ again

praiskup:
Actually, having DNF5 on Fedora 38 Mock **host** is still a good idea for the future when we'll be building Fedora 41 chroots.

This reverts commit 8cca2d02aef81c8dbfb05ba46bb2e8bdf37baafb.

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -68,17 +68,19 @@ BuildRequires: python%{python3_pkgversion}-pylint
 
 %if 0%{?fedora} >= 38
 # DNF5 stack
-Suggests: dnf5
-Suggests: dnf5-plugins
+Recommends: dnf5
+Recommends: dnf5-plugins
 %endif
 
 # DNF4 stack
-Suggests: python3-dnf
-Suggests: dnf-plugins-core
+Recommends: python3-dnf
+Recommends: python3-dnf-plugins-core
 
-Suggests: yum
-Recommends: btrfs-progs
+# YUM stack, dnf-utils replace yum-utils
+Recommends: yum
 Recommends: dnf-utils
+
+Recommends: btrfs-progs
 Suggests: qemu-user-static
 Suggests: procenv
 Recommends: podman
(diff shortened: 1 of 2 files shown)


praiskup:
FYI @evan-goode @j-mracek  - mostly related to our today discussion that we should proopose building of Fedora 41 chroots with dnf5 rather soon.

Conan-Kudo:
Suggests doesn't do anything. You need to use Recommends instead if you want it to actually get installed.

praiskup:
With the bootstrap image feature ON by default, it really isn't worth more than just Suggests?

Conan-Kudo:
We don't have it on by default universally.

praiskup:
That's true, the question is what to do about it.  Do you have a proposal? :)

My understanding is that users have to decide whether they prefer (a) bootstrap image, (b) any package manager (that is actually able to install bootstrap, aka bootstrap image off) or (c) a particular package manager version on the host.  Pushing DNF5 installation on user's hosts doesn't seem **so necessary** considering that we already [kinda push for (a)](https://github.com/rpm-software-management/mock/pull/1101).  The image feature needs to settle down, but long term it is going to be the generally preferred variant, I believe.

Conan-Kudo:
I think we should use Recommends for the package managers. If they don't want it, they can uninstall it and it won't come back again.

Conan-Kudo (approved):

xsuchy (approved):

decathorpe:
Meh, this is really annoying. I'll need to add `exclude=dnf5,yum` to my dnf configuration because of this.

j-mracek:
> Meh, this is really annoying. I'll need to add `exclude=dnf5,yum` to my dnf configuration because of this.

May I ask you what kind of problem you have with `DNF5`? The problems we can resolve.

decathorpe:
I don't have a problem with dnf5, but I don't like the kind of dependency bloat changes like these cause. I already have `exclude=podman,toolbox` in my dnf.conf, and now it's `exclude=podman,toolbox,dnf5,yum` ...

PR #1233, opened 2023-09-23
capture_output=True is not supported on EL8

praiskup:
Per report in Bodhi update:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-45ace77fca


Conan-Kudo (approved):

Conan-Kudo (changes requested):
Can you make the commit message a little more descriptive? It's not very useful as it is...

xsuchy (approved):

PR #1236, opened 2023-09-29
Properly handle Fedora ELN

sandrobonazzola:
Properly handle Fedora ELN in `mock-core-configs/mock-core-configs.spec` by setting the default to `fedora-eln-x86_64.cfg` instead of `fedora+epel-rawhide-x86_64.cfg` which doesn't exists.


praiskup (approved):

praiskup:
Thank you.

PR #1237, opened 2023-10-03
Add flag to indicate system user to mockbuild

mhjacks:
To address https://bugzilla.redhat.com/show_bug.cgi?id=2242018

### DIFF
--- a/mock/py/mockbuild/shadow_utils.py
+++ b/mock/py/mockbuild/shadow_utils.py
@@ -53,7 +53,7 @@ class ShadowUtils:
         """
         command = ["useradd", username]
         if uid is not None:
-            command += ["-o", "-u", str(uid)]
+            command += ["-o", "-r", "-u", str(uid)]
         if gid is not None:
             command += ["-g", str(gid), "-N"]
         if home is not None:


praiskup:
Thank you for the patch, @mhjacks!  The patch looks appropriate to me.

Can you please drop in a `.bugfix` markdown file (see for example [this older one](https://github.com/rpm-software-management/mock/blob/c198b6c0d28d98b09ff08ea1e91ac2da5163bea5/releng/release-notes-next/expanduser-with-copyout.bugfix#L1-L4)), according to [docs](https://rpm-software-management.github.io/mock/Release-Notes-New-Entry)?

What are all the consequences?  Homedir preparation, UID selection, etc.?

I'm not sure if [user-copying](https://github.com/rpm-software-management/mock/blob/3f351f5c269880c561e8ee1a6be592d682da7220/mock/py/mockbuild/buildroot.py#L543-L544)  may be affected by this. If yes, it would be better to have a new `create_user(system=True|False|None)` argument and apply the option only for the `mockbuild` user.

mhjacks:
OK, I've had a little more time to look at this, and I'm inclined to think this might be a problem in useradd/userdel. Logs show. When

```text
DEBUG util.py:543:  Executing command: ['userdel', '-f', 'mockbuild', '--prefix', '/var/lib/mock/fedora-39-x86_64/root'] with env {'TERM': 'vt100', 'SHELL': '/bin/sh', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'LANG': 'C.UTF-8'} and shell False
DEBUG util.py:446:  userdel: cannot remove entry 0 from /var/lib/mock/fedora-39-x86_64/root//etc/subuid
DEBUG util.py:595:  Child return code was: 16
DEBUG util.py:621:  child environment: None
DEBUG util.py:543:  Executing command: ['groupdel', 'mock', '--prefix', '/var/lib/mock/fedora-39-x86_64/root'] with env {'TERM': 'vt100', 'SHELL': '/bin/sh', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'LANG': 'C.UTF-8'} and shell False
DEBUG util.py:446:  groupdel: cannot remove the primary group of user 'mockbuild'
DEBUG util.py:595:  Child return code was: 8
DEBUG util.py:621:  child environment: None
DEBUG util.py:543:  Executing command: ['groupadd', 'mock', '-g', '135', '--prefix', '/var/lib/mock/fedora-39-x86_64/root'] with env {'TERM': 'vt100', 'SHELL': '/bin/sh', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'LANG': 'C.UTF-8'} and shell False
DEBUG util.py:446:  groupadd: group 'mock' already exists
DEBUG util.py:595:  Child return code was: 9
```

I think userdel -f should not fail when there's not a subuid/subgid file entry, but the semantics of this are unusual, and getting a change into useradd/userdel for this (admittedly odd edge case) seems like it might be a stretch. userdel doesn't seem to have the corresponding "system" flag for special handing of subuid/gid that I can see ("-r" means "remove").

So it appears calling create_user with the system flag would not really be a solution here.  I can still build mock when my initial UID is 0 when I have subuid/subgid turned on, but that doesn't seem ideal. I'm not entirely sure where to go from here.

mhjacks:
I've submitted https://bugzilla.redhat.com/show_bug.cgi?id=2249524 on shadow-utils. (Userdel doesn't have an option to force user removal if subids are not present in the local files; they are not always guaranteed to be). This should indicate which direction this PR should go

Issue #1246, opened 2023-10-18
It is possible to generate a BuildRequires which is not used during the build


hroncok:
### Short description of the problem

It's possible to build an RPM package with `%generate_buildrequries` generated BuildRequires which are part of the SRPM metadata but not used during the build.

This is based on the fact that when repeated `%generate_buildrequries` loops happen, they always repeat `%prep`, except for the final round, which is executed redundantly.

The build happens more or less this way:

 1. %prep
 2. %generate_buildrequries
 3. dnf install the result of the above
 4. if something was installed, goto 1.
 5. %generate_buildrequires
 6. %build
 7. ...

Note explicitly the following:

 - the requirements generated in 5. end up in the SRPM metadata but are never dnf-installed
 - there is no %prep before 5.

There are various things that might be confusing for packagers. For instance, why is %prep repeatedly called with each round of %generate_buildrequries except the final one? And why is the final round of %generate_buildrequires executed in the first place?

Nevertheless, this is all more or less just wasteful but OK. Until something is terribly state-dependent or even random.

Take this package:

https://src.fedoraproject.org/rpms/python-setupmeta/blob/61c9a6adfcb67bf62ecd3b19ee27d3bbfce515fe/f/python-setupmeta.spec

It has:

```
%prep
%autosetup -n setupmeta-%{version}
# required to make it not try to self-refer
rm -rf setupmeta.egg-info

%generate_buildrequires
%pyproject_buildrequires
```

When setupmeta.egg-info exists, %pyproject_buildrequires generates a dependency on python3dist(setupmeta) (that's what the upstream setup.py script does, don't ask me why). 

Translated to the build order from above:

 1. %prep unpacks the sources and deletes setupmeta.egg-info
 2. %generate_buildrequries does not see setupmeta.egg-info, does not generate python3dist(setupmeta)
 3. dnf install the result of the above
 4. if something was installed, goto 1.
 5. %generate_buildrequires **does see** setupmeta.egg-info, **does generate** python3dist(setupmeta)
 6. %build
 7. ...

The python3dist(setupmeta) dependency is never installed, but the resulting SRPM (and repodata) claim that it exists.

Similarily, if the %generate_buildrequries  section echoed a random available dependency, the final SPRM metadata would not match what was installed during the build.

### Output of `rpm -q mock`

mock-5.2-1.fc37.noarch

### Steps to reproduce issue

[...]

PR #1247, opened 2023-10-18
do not skip %prep when doing next round of %dynamic_buildrequires

xsuchy:
Reverts: 3fbda88bfbeddb3afb782d1913a2a5ba82f1ee67
Fixes: #1246

When %prep deletes setupmeta.egg-info next round of build due %dynamic_buildrequires works in different environment and results can be different.


releng/release-notes-next/generate_buildrequires.bug
```diff
@@ -0,0 +1,5 @@
+Previously Mock skipped %prep section when doing the next round of build of `%generate_buildrequries` to save time.
+But we found that when the package deletes `setupmeta.egg-info` in `%prep`, then the next loop can work with incorrect data
+and generate incorrect build requires that may not be satisfied.
+Mock now calls `%prep` section in every loop. This may result in a few seconds of delay when spec file uses
+`%generate_buildrequries`, but generated build dependency is correct now.
```
hroncok, line 5:
```suggestion
To save time, Mock previously skipped the `%prep` section when building
an RPM package after completing the installation of all the requirements
generated by `%generate_buildrequries`.

However, when the `%generate_buildrequries` section has side effects altering
the results of a subsequent run of self,
then the final build may work with different data and generate incorrect
build requirements that may not be actually installed during the build.
In extreme case, it is possible to successfully build a package with
unsatisfiable BuildRequires in the built SRPM metadata.

Mock now calls `%prep` section even during the final build after completing
the installation of all the generated build requirements.
This may result in a few seconds of delay when spec file uses
`%generate_buildrequries` but the source used to generate the build requirements
is more likely to stay consistent during the build,
making it harder to accidentally generate unsatisfied BuildRequires.
```

xsuchy:
Thank you. Your wording is indeed much better.

hroncok:
Do not merge. I have a better solution.

praiskup:
Closing, the other PR was merged.

PR #1249, opened 2023-10-19
Dynamic BuildRequires: Prevent generation of unsatisfied dependency

hroncok:
When the %generate_buildrequires section behaved unpredictable, it was possible to successfully build a package with unsatisfied BuildRequires. Now, the build will fail.

Fixes https://github.com/rpm-software-management/mock/issues/1246

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -737,10 +737,11 @@ class Commands(object):
         if additional_opts == ['']:
             additional_opts = []
 
-        def get_command(mode):
+        def get_command(mode, checkdeps=False):
+            nodeps_opt = [] if checkdeps else ['--nodeps']
             command = [self.config['rpmbuild_command']] + mode + \
                       [self._rpmbuild_noclean_option] + \
-                      ['--target', self.rpmbuild_arch, '--nodeps'] + \
+                      ['--target', self.rpmbuild_arch] + nodeps_opt + \
                       check_opt + [spec_path] + additional_opts
             command = ["bash", "--login", "-c"] + [' '.join(command)]
             return command
@@ -748,7 +749,8 @@ class Commands(object):
         bd_out = self.make_chroot_path(self.buildroot.builddir)
         max_loops = int(self.config.get('dynamic_buildrequires_max_loops'))
         success = False
-        if dynamic_buildrequires and self.config.get('dynamic_buildrequires'):
+        dynamic_buildrequires = dynamic_buildrequires and self.config.get('dynamic_buildrequires')
+        if dynamic_buildrequires:
             while not success and max_loops > 0:
                 # run rpmbuild+installSrpmDeps until
                 # * it fails
@@ -787,7 +789,20 @@ class Commands(object):
                 # on re-doing it
                 mode += ['--noprep']
 
-        self.buildroot.doChroot(get_command(mode),
+        # When we used dynamic buildrequires, the rpmbuild call will
+        # execute the %generate_buildrequires section once again
+        # in order to actually add the BuildRequires to the SRPM metadata.
+        # Since the output of the %generate_buildrequires section isn't
+        # controlled by mock, it is possible (albeit unlikely) that the list
+        # of generated BuildRequires will differ now, as it is not guaranteed
+        # to remain stable.
+        # By instructing rpmbuild to check dependencies, we ensure a failure
+        # if a new unsatisfied dependency is generated.
+        # Unfortunately, we can only do this when using a bootstrap chroot,
+        # because the rpm in the chroot might not understand the rpmdb otherwise.
+        # See https://github.com/rpm-software-management/mock/issues/1246
+        checkdeps = dynamic_buildrequires and self.bootstrap_buildroot is not None
+        self.buildroot.doChroot(get_command(mode, checkdeps=checkdeps),
                                 shell=False, logger=self.buildroot.build_log, timeout=timeout,
                                 uid=self.buildroot.chrootuid, gid=self.buildroot.chrootgid,
                                 user=self.buildroot.chrootuser,
--- a/releng/release-notes-next/generate_buildrequires.bug
+++ b/releng/release-notes-next/generate_buildrequires.bug
@@ -1,0 +1,19 @@
+When Mock completes the installation of all the requirements generated
+by `%generate_buildrequries`, it calls `rpmbuild -ba` to perform a final build
+of the package.
+
+During the final build, `%generate_buildrequries` runs again in order to
+generate a list of `BuildRequires` to be added to the built SRPM metadata.
+An arbitrary `%generate_buildrequries` section may generate different
+requirements that may not have been installed.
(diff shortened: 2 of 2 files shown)


mock/py/mockbuild/backend.py
```diff
@@ -737,18 +737,20 @@ def rebuild_package(self, spec_path, timeout, check, dynamic_buildrequires):
...
+                      ['--target', self.rpmbuild_arch] + nodeps_opt + \
                       check_opt + [spec_path] + additional_opts
             command = ["bash", "--login", "-c"] + [' '.join(command)]
             return command
 
         bd_out = self.make_chroot_path(self.buildroot.builddir)
         max_loops = int(self.config.get('dynamic_buildrequires_max_loops'))
         success = False
-        if dynamic_buildrequires and self.config.get('dynamic_buildrequires'):
+        dynamic_buildrequires = dynamic_buildrequires and self.config.get('dynamic_buildrequires')
```
hroncok, line 752: I've repurposed this variable here to say whether dynamic_buildrequires were used. If that's not nice, I can create a new one, or decide later based on the value of `success`.

mock/py/mockbuild/backend.py
```diff
@@ -787,7 +789,15 @@ def get_command(mode):
...
 
-        self.buildroot.doChroot(get_command(mode),
+        # when we used dynamic_buildrequires, we must let rpmbuild check deps
+        # becasue rpmbuild will call %genearte_buildrequires once again
+        # to actually add the BuildRequires to the SRPM metadata
+        # we need to abort the build if %genearte_buildrequires generates a new
+        # unsatisfied dependency
+        # see https://github.com/rpm-software-management/mock/issues/1246
+        # not using --nodeps is safe in this case
+        # becasue the SRPM was generated by the rpm in the chroot
```
praiskup, line 799: Do you react on this?  https://github.com/rpm-software-management/mock/blob/c3a4bca61d1863cb8b59e8ba6d2feaff1e67ace7/mock/py/mockbuild/backend.py#L720-L721

hroncok: Yes

mock/py/mockbuild/backend.py
```diff
@@ -787,7 +789,15 @@ def get_command(mode):
                 # on re-doing it
                 mode += ['--noprep']
 
-        self.buildroot.doChroot(get_command(mode),
+        # when we used dynamic_buildrequires, we must let rpmbuild check deps
+        # becasue rpmbuild will call %genearte_buildrequires once again
+        # to actually add the BuildRequires to the SRPM metadata
```
praiskup, line 794: Can you separate sentences with capital words and periods?  That would help me to understand this and the following sentence.

mock/py/mockbuild/backend.py
```diff
@@ -787,7 +789,15 @@ def get_command(mode):
                 # on re-doing it
                 mode += ['--noprep']
 
-        self.buildroot.doChroot(get_command(mode),
+        # when we used dynamic_buildrequires, we must let rpmbuild check deps
+        # becasue rpmbuild will call %genearte_buildrequires once again
```
praiskup, line 793: s/genearte/generate/ (typo repeated below)

mock/py/mockbuild/backend.py
hroncok: Sure thing

mock/py/mockbuild/backend.py
```diff
@@ -787,7 +789,15 @@ def get_command(mode):
...
+        # to actually add the BuildRequires to the SRPM metadata
+        # we need to abort the build if %genearte_buildrequires generates a new
+        # unsatisfied dependency
```
praiskup, line 796: I fail to follow this part, can you elaborate on how this can happen?  We tried to resolve the last nosrc.rpm, nothing was installed (and if there was some unresolved build-dep, we would fail before).  What's the reason for generating a different dependency list?

hroncok: The reason is that rpmbuild MUST do it to be able to store the generated BuildRequires into the SRPM metadata. If rpmbuild skipped this step, mock would have already installed them, sot he build would succeed, but the BuildRequires would be lost.

praiskup: The question is not why the step needs to be repeated, to properly follow this paragraph I want to absolutely understand "How come the dep list is different" ... some unwanted side-effects?

mock/py/mockbuild/backend.py
praiskup: Well, are you sure that (a) the ability to generate the src.rpm implies that (b) the in-chroot RPM can read the RPM database?

hroncok: Oh. No, I am not sure. Without bootstrap chroot, this might blow up. I'll remove the comment. Whether we can safely do this or not, I ma not sure.

praiskup: I mean -> this is quite OK if bootstrap is turned ON, but we still support non-bootstrapped builds.  Consider that the in-chroot database might be generated by an outdated RPM version from host.

hroncok:
I amended the comment.

mock/py/mockbuild/backend.py
praiskup: This is fine for sure!

mock/py/mockbuild/backend.py
hroncok: I see. In that case, we cannot call `rpmbuild` without `--nodeps` here unless we are in a bootstrap mode.

A hacky solution for disabled bootstrap is to call installSrpmDeps on the final built SRPM and die if it fails to install something. That's not as nice.

Alternatively, we might force the `--nodeps`  when not in bootstrap and claim this problem is only fixed when bootstrap is on. What do you think?

hroncok: I pushed a fixup commit that only does this when using a bootstrap chroot. It's not perfect, but it is a good enough solution for Fedora.

releng/release-notes-next/generate_buildrequires.bug
```diff
@@ -0,0 +1,19 @@
...
+generate a list of `BuildRequires` to be added to the built SRPM metadata.
+An arbitrary `%generate_buildrequries` section may generate different
+requirements that may not have been installed.
+
+Previously, the `rpmbuild -ba` call used the `--nodeps` option,
+hence it was possible to successfully build a package with
+unsatisfiable BuildRequires in the built SRPM metadata.
+
+When a bootstrap chroot is used, the `--nodeps` option is no longer used
+in the final `rpmbuild -ba` call.
```
praiskup, line 15: Can you use `[no longer used][PR#1249]` instead to link this PR in release notes, or link `[issue#1246]` somehow?

praiskup:
Otherwise LGTM, nice work.

hroncok:
Amended, squashed.

praiskup (approved):
Thank you.

PR #1253, opened 2023-11-09
Add rpmautospec plugin

nphilipp:
This plugin preprocesses spec files using %autorelease and/or %autochangelog in the mock chroot prior to building SRPMs. The `rpmautospec` package will be installed into the chroot if necessary.

/cc @sgallagher

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -65,6 +65,7 @@ BuildRequires: python%{python3_pkgversion}-devel
 %if %{with lint}
 BuildRequires: python%{python3_pkgversion}-pylint
 %endif
+BuildRequires: python%{python3_pkgversion}-rpmautospec-core
 
 %if 0%{?fedora} >= 38
 # DNF5 stack
@@ -131,6 +132,15 @@ Requires: lvm2
 %description lvm
 Mock plugin that enables using LVM as a backend and support creating snapshots
 of the buildroot.
+
+%package rpmautospec
+Summary: Rpmautospec plugin for mock
+Requires: %{name} = %{version}-%{release}
+# This lets mock determine if a spec file needs to be processed with rpmautospec.
+Requires: python%{python3_pkgversion}-rpmautospec-core
+
+%description rpmautospec
+Mock plugin that preprocesses spec files using rpmautospec.
 
 %package filesystem
 Summary:  Mock filesystem layout
@@ -265,6 +275,10 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 %files lvm
 %{python_sitelib}/mockbuild/plugins/lvm_root.*
 %{python3_sitelib}/mockbuild/plugins/__pycache__/lvm_root.*.py*
+
+%files rpmautospec
+%{python_sitelib}/mockbuild/plugins/rpmautospec.*
+%{python3_sitelib}/mockbuild/plugins/__pycache__/rpmautospec.*.py*
 
 %files filesystem
 %license COPYING
--- a/mock/py/mockbuild/plugins/rpmautospec.py
+++ b/mock/py/mockbuild/plugins/rpmautospec.py
@@ -1,0 +1,128 @@
+# -*- coding: utf-8 -*-
+# vim:expandtab:autoindent:tabstop=4:shiftwidth=4:filetype=python:textwidth=0:
+# License: GPL2 or later see COPYING
+# Copyright (C) 2023 Stephen Gallagher <sgallagh@redhat.com>
+# Copyright (C) 2023 Nils Philippsen <nils@redhat.com>
+"""A mock plugin to pre-process spec files using rpmautospec."""
+
+from pathlib import Path
+from typing import Optional, Union
+
+from rpmautospec_core import specfile_uses_rpmautospec
+
+from mockbuild.exception import ConfigError, PkgError
+from mockbuild.trace_decorator import getLog, traceLog
+
+requires_api_version = "1.1"
+
+
+@traceLog()
+def init(plugins, conf, buildroot):
+    """Register the rpmautospec plugin with mock."""
+    RpmautospecPlugin(plugins, conf, buildroot)
+
+
+class RpmautospecPlugin:
+    """Fill in release and changelog from git history using rpmautospec"""
+
+    @traceLog()
+    def __init__(self, plugins, conf, buildroot):
+        self.buildroot = buildroot
+        self.config = buildroot.config
+        self.opts = conf
+        self.log = getLog()
+
+        if "cmd_base" not in self.opts:
+            raise ConfigError("The 'rpmautospec_opts.cmd_base' is unset")
+
+        plugins.add_hook("pre_srpm_build", self.attempt_process_distgit)
+        self.log.info("rpmautospec: initialized")
+
+    @traceLog()
+    def attempt_process_distgit(
+        self,
+        host_chroot_spec: Union[Path, str],
+        host_chroot_sources: Optional[Union[Path, str]],
+    ) -> None:
+        """Attempt to process a spec file with rpmautospec."""
+        # Set up variables and check prerequisites.
+        if not host_chroot_sources:
+            self.log.debug("Sources not specified, skipping rpmautospec preprocessing.")
+            return
+
+        host_chroot_spec = Path(host_chroot_spec)
+        host_chroot_sources = Path(host_chroot_sources)
+        if not host_chroot_sources.is_dir():
+            self.log.debug(
+                "Sources not a directory, skipping rpmautospec preprocessing."
+            )
+            return
+
+        distgit_git_dir = host_chroot_sources / ".git"
+        if not distgit_git_dir.is_dir():
+            self.log.debug(
+                "Sources is not a git repository, skipping rpmautospec preprocessing."
+            )
+            return
+
+        host_chroot_sources_spec = host_chroot_sources / host_chroot_spec.name
+        if not host_chroot_sources_spec.is_file():
+            self.log.debug(
+                "Sources doesn’t contain spec file, skipping rpmautospec preprocessing."
+            )
+            return
+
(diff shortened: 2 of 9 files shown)


nphilipp:
The `rpm-build:fedora-rawhide-x86_64:mock` check failed because the unit tests depend on the `rpmautospec-core` Python package which needs to be available as an RPM package in Fedora, see fedora-infra/rpmautospec-core#10.

praiskup:
Thank you for your contribution!  Nice.

Can you please document the plugin https://rpm-software-management.github.io/mock/#plugins ?

Some of the new PyLint warnings are too pedantic (I'm going to tweak the
rules), though some are worth fixing (plus I'd vote for having at least
some of those non-test doc strings written, not a hard requirement).

> The rpmautospec package will be installed into the chroot if necessary.

But we also require it also host because of
`from rpmautospec_core import specfile_uses_rpmautospec`. 

Have you considered calling the host tooling (on the sources stored in-chroot)?

praiskup:
Also, can you please write a [release notes entry](https://rpm-software-management.github.io/mock/Release-Notes-New-Entry), template:
```
$ cat releng/release-notes-next/autospec.feature 
New plugin ... [has been implemented][PR#1254] ...
...
```

praiskup:
Can you please also add a "commented out" plugin-config into the docs file `./mock/docs/site-defaults.cfg`?  We document all the available options there.

mock/tests/compat.py
```diff
@@ -0,0 +1,12 @@
+"""Compatibility code for old Python versions:"""
+
+try:
+    from contextlib import nullcontext  # noqa: F401
```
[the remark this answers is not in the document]
nphilipp: Yeah, that’s the happy path. If it fails to import, we implement it in the `compat` module and the tests using it import `nullcontext` from there.

mock/tests/plugins/test_rpmautospec.py
```diff
@@ -0,0 +1,185 @@
...
+from mockbuild.exception import ConfigError
+from mockbuild.plugins import rpmautospec
+
+from ..compat import nullcontext
+
+UNSET = object()
+
+
+@mock.patch("mockbuild.plugins.rpmautospec.RpmautospecPlugin")
+def test_init(RpmautospecPlugin):  # noqa: C0103
```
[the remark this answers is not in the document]
nphilipp: This is to match with the object that is being mocked in the test.

mock/tests/plugins/test_rpmautospec.py
```diff
@@ -0,0 +1,185 @@
...
+            "happy-path-no-requires",
+            "without-sources",
+            "sources-not-dir",
+            "sources-not-repo",
+            "sources-no-specfile",
+            "spec-files-different",
+            "specfile-no-rpmautospec",
+        ),
+    )
+    def test_attempt_process_distgit(self, testcase, tmp_path):  # noqa: R0912, R0915
```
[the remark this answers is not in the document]
nphilipp: And this is a side effect to setting up and verify things slightly differently for the parametrized variants of the test.

mock/mock.spec
```diff
@@ -275,6 +288,9 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 %dir  %{_datadir}/cheat
 
 %changelog
+* Tue Nov 21 2023 Nils Philippsen <nils@redhat.com>
+- Package rpmautospec plugin
+
```
praiskup, line 293: Please no %changelog upstream :-)

mock/tests/compat.py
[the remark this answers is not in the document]
praiskup: Can't you use the `mockbuild.util.nullcontext`?

mock/py/mockbuild/plugins/rpmautospec.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if not specfile_uses_rpmautospec(host_chroot_sources_spec):
+            self.log.debug(
+                "Spec file doesn’t use rpmautospec, skipping rpmautospec preprocessing."
+            )
+            return
+
+        # Install the `rpmautospec` command line tool into the build root.
+        if self.opts.get("requires", None):
+            self.buildroot.pkg_manager.install_as_root(*self.opts["requires"], check=True)
```
praiskup, line 93: The question still applies..  Depending on the packaged-in-chroot rpmautospec means that the behavior may vary depending on the Mock --chroot option.

nphilipp: Sorry missed this comment of yours so far. What do you mean? I’ve looked up the `--chroot` option but nothing jumps at me how this is related to using rpmautospec as packaged in the chroot.

praiskup: That's the point of `self.opts.get("requires", None)`, right?  You want to install some packages into the build root.  Or am I missing something?

sgallagher: OK, I think I can translate:

@praiskup Are you asking "What happens if the repos that the mock chroot is using don't contain the packages this is requesting?"

That will throw an exception here, if I'm reading things right. Maybe we should catch it and make sure there's a reasonable error message written, but if the specfile requires rpmautospec and rpmautospec isn't present in the targeted buildroot, failing is the right outcome.

nphilipp: The default assume rpmautospec and dependencies need to be installed, but we want to have the option available to just put it in the image (and then skip attempts to install it). Or maybe I’m still not understanding the question. 😉

praiskup: Thank you for the comments above.

The question I wanted to ask is whether we need to depend on both (a) Autospec on host, as the hard Requires dependency, and (b) in-chroot, installed using `self.opts.get("requires", None)`.

Can't we use host's `Autospec` for everything?

praiskup: Said that, I'd like to clear that question out -- but I agree that, if we have to fail for missing in-chroot package, a nice error message is worth it.

mock/py/mockbuild/plugins/rpmautospec.py
```diff
@@ -0,0 +1,117 @@
...
+        chroot_spec = Path("/") / host_chroot_spec.relative_to(chroot_dir)
+        chroot_sources = Path("/") / host_chroot_sources.relative_to(chroot_dir)
+        chroot_sources_spec = Path("/") / host_chroot_sources_spec.relative_to(chroot_dir)
+
+        # Call subprocess to perform the specfile rewrite
+        command = list(self.opts["cmd_base"])
+        command += [chroot_sources_spec]  # <input-spec>
+        command += [chroot_spec]  # <output-spec>
+
+        self.buildroot.doChroot(
```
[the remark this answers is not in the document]
sgallagher: Pretty much the entire reason for doing this in a mock plugin is because we _cannot_ do this on the host. We moved to mock so that we can rely on rpmautospec being available in the target buildroot.

mock/mock.spec
```diff
@@ -65,6 +65,7 @@ BuildRequires: python%{python3_pkgversion}-devel
 %if %{with lint}
 BuildRequires: python%{python3_pkgversion}-pylint
 %endif
+BuildRequires: python%{python3_pkgversion}-rpmautospec-core
```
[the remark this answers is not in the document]
praiskup: Ah, no, it's a BuildRequires.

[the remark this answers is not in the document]
nphilipp: Yes, it’s needed to execute the tests.

mock/py/mockbuild/plugins/rpmautospec.py
nphilipp: I assume, that another exception should be thrown so mock can signal a problem to the caller (e.g. Koji), right?

nphilipp:
> Hello @nphilipp! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
>     * In the file [`mock/py/mockbuild/plugins/rpmautospec.py`](https://github.com/rpm-software-management/mock/blob/0fd14985906a58701acc8f6697735195364b1a60/mock/py/mockbuild/plugins/rpmautospec.py):
> 
> 
> > [Line 98:21](https://github.com/rpm-software-management/mock/blob/0fd14985906a58701acc8f6697735195364b1a60/mock/py/mockbuild/plugins/rpmautospec.py#L98): [W503](https://duckduckgo.com/?q=pep8%20W503) line break before binary operator
> 
>     * In the file [`mock/tests/plugins/test_rpmautospec.py`](https://github.com/rpm-software-management/mock/blob/0fd14985906a58701acc8f6697735195364b1a60/mock/tests/plugins/test_rpmautospec.py):
> 
> 
> > [Line 199:21](https://github.com/rpm-software-management/mock/blob/0fd14985906a58701acc8f6697735195364b1a60/mock/tests/plugins/test_rpmautospec.py#L199): [W503](https://duckduckgo.com/?q=pep8%20W503) line break before binary operator

This actually conforms to what PEP8 says about breaking lines around binary operators: https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

nphilipp:
@praiskup the update should be available with the next EPEL8 compose and I’ve added the reasons for doing this in the chroot as a comment.

praiskup (approved):
LGTM, thank you.

praiskup:
> Have you considered calling the host tooling (on the sources stored in-chroot)?

Just getting back to this; I still think this is a bad idea.

The Mock's mission should be to abstract from the in-chroot nuances, and rely on "easily fixable" on-host tooling (by example, that way you can bootstrap a new distribution without having the autospec tooling built yet, or build fedora packages with autospec against distributions that do not support it).

Issue #1171, opened 2023-07-28
Particular distro-specific configuration should have their maintainers


praiskup:
I really appreciate the situation in mock-core-configs;  that (sub)project becomes a central place for cooperation for so many RPM GNU/Linux distributions.  It is awesome to have all the configurations maintained in one place.  It is just pretty difficult to get everything fixed (we test all the configs during the pre-release time, and there are too many failures).

We should have a way to identify a particular person or team that can provide timely responses to questions related to specific distribution configs, and ideally even patches.  I am in particular highly Fedora/RHEL/CentOS oriented, and it is relatively hard to search the web for questions about other distros like "Where is the docker image of Foo", "Where is the baseurl address", "Is there a metalink or mirrorlist address?", "Is it [dead](https://github.com/rpm-software-management/mock/issues/1170)"?

My proposal would be to have (a) either a documentation page with a list of distributions supported in `mock-core-configs` and corresponding responsible people or (b) a metadata section for every config file declaring the same info.  Is there any other/better option?

I'd like to use this issue as a place to gather the necessary info that we should have documented, too:

- [ ] AlmaLinux 
- [ ] Amazon Linux
- [ ] Anolis OS
- [x] CentOS - Copr Team
- [ ] CircleLinux
- [x] EPEL - Copr Team
- [x] Fedora - Copr Team
- [ ] Mageia
- [ ] Navy
- [ ] OpenEuler - @pkking
- [ ] OpenMandriva
- [ ] OpenSUSE
- [ ] OracleLinux
- [ ] RHEL - Copr Team
- [ ] Rocky Linux

Help here is very welcome!

nazunalika:
Hello! I happened to run into this issue from looking for something else mock related, and figured this is something I could comment on. I am one of the lead release engineers for Rocky Linux and submitted the configs for both 8 and 9 in the past. If you would like, you can mark myself and @NeilHanlon as the responsible parties for questions around Rocky Linux and the related mock configurations.

Please let us know if there are any questions you may have and we'll try our best to help!

praiskup:
@nazunalika awesome!  Thank you for the help.

praiskup:
From #963, @sharewax / @Mno-hime - would you mind taking care of oraclelinux configs, and signing yourself into [the docs](https://rpm-software-management.github.io/mock/Mock-Core-Configs)?

praiskup:
Per #947, @mskdxl can you assign yourself for the Anolis OS configuration?

[...]

PR #1259, opened 2023-11-27
Mock core configs - README.md with maintainers and issue trackers

FrostyX:
See also #1171

We have a lot of chroots, so it is not uncommon, that some of them break from time to time. Most of the reports that I've got were about Fedora ELN but there were others as well. People usually report them as a Copr issue, so I need to point them here (and make sure the maintainer knows) or point them to the distribution issue tracker, which I don't remember from the top of my head.

Hopefully, this will improve the situation.
What do you think?

### DIFF
--- a/mock-core-configs/README.md
+++ b/mock-core-configs/README.md
@@ -1,0 +1,44 @@
+# Mock core configs
+
+The mock configs in this package are maintained by the community.
+When encountering an issue please use your best judgement to decide
+whether a Mock config is broken, or the distribution is broken.
+
+
+## Mock config issues
+
+If a Mock config is broken (e.g. [#756][mock-756]), please
+[create a ticket for this repository][mock-issues]
+and tag the responsible maintainer from the table below.
+
+
+## Distribution or repository issues
+
+If a distribution or repository is broken (e.g. [#889][mock-889]),
+please report the issue to the appropriate issue tracker for the
+distribution.
+
+
+## Table
+
+| Distribution                                                                   | Chroots           | Maintainer                                                            | Distribution or repository issue tracker |
+| ------------------------------------------------------------------------------ | ----------------- | --------------------------------------------------------------------- | ------------- |
+| [AlmaLinux](https://almalinux.org/)                                            | `almalinux-*`     | [@Conan-Kudo](https://github.com/Conan-Kudo)                          | [Issues](https://bugs.almalinux.org/)  |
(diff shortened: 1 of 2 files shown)


xsuchy (approved):
I would prefer this data in configs itself and generate the table from the configs... but this is definitely improvement. So +1

praiskup:
Awesome, thank you for the PR!

> The table is missing some chroots

I think it is fair to state that we maintain CentOS & RHEL chroots, wdyt?  Simply add a link to our tracker.

praiskup:
I think it is worth underlining when users should contact 1) us to fix the config or 2) the distro tracker because the distribution is broken (I'd put **Distribution issues** header for the colum with links). I.e. clearly tell them that for configuration bugs they should open an issue against us, or even better, send a patch.

FrostyX:
> I think it is fair to state that we maintain CentOS & RHEL chroots, wdyt?

Sure

> I think it is worth underlining when users should contact 1) us to fix the config or 2) the distro tracker because the distribution is broken (I'd put Distribution issues header for the colum with links)

I tried to explain that in the text above the table. Should I re-word or highlight some of its parts?

FrostyX:
I added the missing CentOS and RHEL chroots and tried to improve the config issue vs distro issue readability.

> What if we created a proper user documentation page, e.g. referenced directly on [homepage](https://rpm-software-management.github.io/mock/)? We could keep README.md as you propose, but just link the new docs?

I don't have any preference. We can either reference this README from the documentation or we can move the page to the documentation and have only a README with the link to the documentation. It's up to you. I think this would be a good place  where to mention it https://rpm-software-management.github.io/mock/#chroot-config-files

praiskup:
> I added the missing CentOS and RHEL chroots and tried to improve the config
> issue vs distro issue readability.

Nice, I really like it now!  Thank you.

> I don't have any preference. We can either reference this README from the
> documentation or we can move the page to the documentation and have only a
> README with the link to the documentation. It's up to you. I think this would
> be a good place where to mention it
> https://rpm-software-management.github.io/mock/#chroot-config-files

I'd probably prefer a separate documentation page, referenced both by the
README file and the place you mention (index.md is quite long now to add
the table directly there).

The benefit is that we could "quickly" update the documentation page when needed.
The README.md is distributed with release tarballs/RPMs and old
RPMs would be prone to have outdated info..

Of course, we can merge as-is, and I can do this change later if you prefer.

xsuchy (approved):

PR #1282, opened 2024-01-09
dnf5: configure system_cachedir for dnf5

praiskup:
With DNF5, the cache directory for root is now configured via `system_cachedir`.  Dnf4 seems to silently ignore this config option.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2256945
Relates: https://github.com/rpm-software-management/dnf5/issues/1150


evan-goode (approved):
Yeah, if you want DNF 5 to store cache in `/var/cache/dnf` to be consistent with DNF 4, then that's fine. DNF 5 will not write to `/var/cache/dnf` by default so there should be no conflicts as long as DNF 4 and DNF 5 are not running on the same filesystem.

PR #1283, opened 2024-01-10
Run shadow-utils inside the chroot to avoid contradictory configurations

mhjacks:
This is the suggestion that @sgallagher offered; I tested this on a FreeIPA-enrolled system and it worked.
I'm a bit nervous about the comment above the line about old shadow-utils being problematic; but we definitely also have the case where shadow-utils config from the host "leaks" into the chroot. I'm not sure which problem is bigger.

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -653,3 +653,13 @@
 # 'BuildRequires: pesign' package which would overwrite the ownership of the
 # socket file.  See https://github.com/rpm-software-management/mock/issues/1091
 #config_opts["copy_host_users"] = []
+
+# Whether to use host's shadow-utils to provision users and groups in the
+# buildroot, which we normally want to do because host shadow-utils are
+# newer and more flexible than buildroot ones. However, there is an issue in shadow-utils
+# where even using the --prefix (or, even --root if we did it that way) option, the host
+# config will "leak" into the chroot. This is not an issue if the configs are
+# effectively the same between host and buildroot, but will cause problems if, for
+# example, the host is configured to use FreeIPA-provided subids.
+# See https://github.com/shadow-maint/shadow/issues/897
+# config_opts["use_host_shadow_utils"] = True
--- a/mock/py/mockbuild/shadow_utils.py
+++ b/mock/py/mockbuild/shadow_utils.py
@@ -14,15 +14,17 @@ class ShadowUtils:
     def __init__(self, root):
         self.root = root
 
-    @property
-    def _chroot_opts(self):
-        return ["--prefix", self.root.make_chroot_path()]
-
     def _execute_command(self, command, can_fail=False):
         with self.root.uid_manager.elevated_privileges():
-            # Execute the command _on host_, not in bootstrap (where we're not
-            # sure how old shadow-utils are).
-            do_with_status(command + self._chroot_opts, raiseExc=not can_fail)
+            # Ordinarily we do not want to depend on shadow-utils in the buildroot, but
+            # configuring certain options (such as FreeIPA-provided subids) can make it
+            # impossible to create users in the buildroot using host shadow-utils so we
+            # provide this workaround.
+            # Tracking upstream bug https://github.com/shadow-maint/shadow/issues/897
+            if self.root.config['use_host_shadow_utils']:
+                do_with_status(command + ['--prefix', self.root.make_chroot_path()], raiseExc=not can_fail)
+            else:
+                self.root.doChroot(command, raiseExc=not can_fail)
 
     def delete_user(self, username, can_fail=False):
         """
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -355,6 +355,13 @@ def setup_default_config_opts():
 
     config_opts["copy_host_users"] = []
 
+    # shadow-utils --prefix and --root options do not play well with
+    # FreeIPA-provided subids. Using the shadow-utils inside the
+    # chroot works around this but this is a niche situation so it is
+    # not the default.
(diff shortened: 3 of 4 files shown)


sgallagher:
FWIW, I think the comment there refers mostly to whether the version of shadow-utils in the container supported the `--root` option, which it probably didn't if we were building for an old RHEL release. I'm 99% sure that all non-EOL RHEL systems have that feature at this point.

mhjacks:
That's good to know. This approach specifically avoids using the `--root` option now because of the "leaking" problem discussed above. Thanks for the feedback!

mhjacks:
This is pursuant to our Slack discussion of how to handle this - I know the full patch would need documentation in the defaults file and probably tests as well. Since the buildroot object has the config dict, we can look it up there.

The way this is structured, the default case would be to use the host's shadow-utils, and then only if the explicit configuration is set to use the chroot's would the chroot's utils be used. I'm definitely open to more idiomatic ways to express it but that's the intention.

praiskup:
Thank you for your work on this, @mhjacks.

- can you please specify the default value in config.py?
- use_chroot_shadow_utils should be also documented in `./mock/docs/site-defaults.cfg`
- it would be nice to have a new [releng/release-notes-next/*.feature markdown entry](https://rpm-software-management.github.io/mock/Release-Notes-New-Entry) documenting the change.

Note that Mock no longer installs shadow-utils into the minimal buildroot as it used to, this has changed in a34d3ae954fb8dcc04e885b44f3b76bb72e94f6a.  Mock never installed shadow-utils into the bootstrap chroot explicitly.
bootstrap typically only contains `dnf` and its dependencies. But since this is opt-in, I think it is just fine to depend on an implicit set of packages.

mock/py/mockbuild/shadow_utils.py
```diff
@@ -15,14 +15,18 @@ def __init__(self, root):
...
 
     def _execute_command(self, command, can_fail=False):
         with self.root.uid_manager.elevated_privileges():
-            # Execute the command _on host_, not in bootstrap (where we're not
-            # sure how old shadow-utils are).
-            do_with_status(command + self._chroot_opts, raiseExc=not can_fail)
+            # Execute the command in bootstrap, since host configuration may not permit
+            # ordinary user addition, such as would happen with freeipa-provided subids
+            if ('use_chroot_shadow_utils' in self.root.config) and self.root.config['use_chroot_shadow_utils']:
+                do_with_status(command, chrootPath=self.chroot_dir, raiseExc=not can_fail)
```
praiskup, line 26: I think you are not executing in the bootstrap chroot, but in buildroot (build chroot) (see doOutChroot).

praiskup:
> FWIW, I think the comment there refers mostly to whether the version of shadow-utils in the container supported the `--root` option, which it probably didn't if we were building for an old RHEL release. I'm 99% sure that all non-EOL RHEL systems have that feature at this point.

There only used to be nuance between `-N` and `-n` option, see the [4.0 release notes](https://rpm-software-management.github.io/mock/Release-Notes-4.0). But indeed, this is no longer the case for non-EOL RHEL systems.

Mock never used `--prefix` with shadow-utils from bootstrap, and never used ShadoUtils' `--root` at all. That's because `--root` seems to be broken even on recent distributions.

The reasoning behind the "on host decision" is that Mock upstream expects that the machines that are used by users for Mock builds are "new enough".  Ideally Fedora builders, or the latest RHEL releases (users should start moving builders out from RHEL8 definitely).  Then, using the tooling "on host" is much more predictable (and fixable) for the builder machine administrator) then relying on various in-chroot tooling versions (one machine typically builds for different distributions, and builder admins are unlikely to have rights to fix the target buildroots).

mock/py/mockbuild/shadow_utils.py
mhjacks: Odd. This worked like I expected it to. I'll investigate it further before finalizing.

mhjacks:
Converting to draft to add the other items you requested. Thanks for the review!

mock/py/mockbuild/shadow_utils.py
```diff
@@ -15,14 +15,18 @@
         self.root = root
 
     @property
-    def _chroot_opts(self):
-        return ["--prefix", self.root.make_chroot_path()]
+    def chroot_dir(self):
```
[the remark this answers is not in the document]
mhjacks: Removed the method as it is only used once

mock/py/mockbuild/shadow_utils.py
```diff
@@ -14,15 +14,17 @@ class ShadowUtils:
...
-    @property
-    def _chroot_opts(self):
-        return ["--prefix", self.root.make_chroot_path()]
-
     def _execute_command(self, command, can_fail=False):
         with self.root.uid_manager.elevated_privileges():
-            # Execute the command _on host_, not in bootstrap (where we're not
-            # sure how old shadow-utils are).
-            do_with_status(command + self._chroot_opts, raiseExc=not can_fail)
+            # Ordinarily we do not want to depend on shadow-utils in the bootstrap, but
```
praiskup, line 19: s/bootstrap/buildroot/

mock/docs/site-defaults.cfg
```diff
@@ -653,3 +653,13 @@
 # 'BuildRequires: pesign' package which would overwrite the ownership of the
 # socket file.  See https://github.com/rpm-software-management/mock/issues/1091
 #config_opts["copy_host_users"] = []
+
+# Whether to use host's shadow-utils to provision users and groups in the
+# buildroot, which we normally want to do because host shadow-utils are
+# newer and more flexible than buildroot ones.However, there is an issue in shadow-utils
```
praiskup, line 659: s/ones.H/ones. H/

praiskup (approved):
Thank you!

PR #1291, opened 2024-01-19
configs: BuildWithDNF5 for ELN

praiskup:
Relates: https://pagure.io/releng/issue/11895

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-eln.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-eln.tpl
@@ -12,17 +12,22 @@ config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils f
 
 config_opts['dist'] = 'eln'  # only useful for --resultdir variable subst
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
-config_opts['package_manager'] = 'dnf'
+
+# https://fedoraproject.org/wiki/Changes/BuildWithDNF5
+# https://pagure.io/releng/issue/11895
+config_opts['package_manager'] = 'dnf5'
 
 # Per https://github.com/fedora-eln/eln/issues/164 updated up to 4 times a day.
 # Docs: https://docs.fedoraproject.org/en-US/eln/deliverables/#_container_image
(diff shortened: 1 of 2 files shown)


sgallagher:
Let's hold off on this for a few more days; I've got two related MRs out that we should land first:

1. https://github.com/minimization/content-resolver-input/pull/1105 - makes sure ELN has `dnf5` and `dnf5-plugins` explicitly included. They're present right now, but only as part of ELN Extras.
2. https://pagure.io/fedora-kickstarts/pull-request/1025 - Once the above is merged, we will make `dnf5[-plugins]` part of the bootstrap image.

Then we can avoid needing to turn off the `bootstrap_ready` option.

yselkowitz:
Can this proceed now?

praiskup:
Yeah, we have to...  this is breaking Copr builds.

mock-core-configs/etc/mock/templates/fedora-eln.tpl
```diff
@@ -12,17 +12,24 @@ config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils f
...
 # Per https://github.com/fedora-eln/eln/issues/164 updated up to 4 times a day.
 # Docs: https://docs.fedoraproject.org/en-US/eln/deliverables/#_container_image
 config_opts['bootstrap_image'] = 'quay.io/fedoraci/fedora:eln'
-# Per https://github.com/fedora-eln/eln/issues/166
-config_opts['bootstrap_image_ready'] = True
+
+# Used to be True per https://github.com/fedora-eln/eln/issues/166 but with
+# the 'dnf5' we are switching to False, reobserve after
+# https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5 applied to ELN!
+config_opts['bootstrap_image_ready'] = False
```
praiskup, line 27: Unfortunately, we made this change in ELN stack in the wrong order -> the ELN image now contains `dnf5` and not `dnf4`, so it stays "ready".  Going to update..

FrostyX:
I just tested the proposed config and it works well.
+1

Issue #1258, opened 2023-11-27
 Setup automatic beaker/behave testing after each merge


praiskup:
And semi/automatically before merging "on demand" -- the tests are pretty time-consuming and wasting all the resources for every "push" wouldn't make sense.

The benefit of this change will be a significant reduction of time spent on doing mock releases.

I assume we could and should use Fedora Testing Farm for this.

Issue #1300, opened 2024-02-01
The --spec option doesn't work with `-rw-------` files


praiskup:
```
INFO: EXCEPTION: [Error('Command failed: \n # /usr/bin/systemd-nspawn -q -M dcb76b3793154753873b620c85c143c0 -D /var/lib/mock/fedora-38-x86_64-5874-23614/root -a -u mockbuild --capability=cap_ipc_lock --bind=/tmp/mock-resolv.hlsh9u8n:/etc/resolv.conf --bind=/dev/btrfs-control --bind=/dev/mapper/control --bind=/dev/fuse --bind=/dev/loop-control --bind=/dev/loop0 --bind=/dev/loop1 --bind=/dev/loop2 --bind=/dev/loop3 --bind=/dev/loop4 --bind=/dev/loop5 --bind=/dev/loop6 --bind=/dev/loop7 --bind=/dev/loop8 --bind=/dev/loop9 --bind=/dev/loop10 --bind=/dev/loop11 --console=pipe --setenv=TERM=vt100 --setenv=SHELL=/bin/bash --setenv=HOME=/builddir --setenv=HOSTNAME=mock --setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin \'--setenv=PROMPT_COMMAND=printf "\\033]0;<mock-chroot>\\007"\' \'--setenv=PS1=<mock-chroot> \\s-\\v\\$ \' --setenv=LANG=C.UTF-8 --resolv-conf=off bash --login -c \'/usr/bin/rpmbuild -bs  --target x86_64 --nodeps /builddir/build/originals/test-C.spec\'\n', 1)]
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
             ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/mockbuild/util.py", line 612, in do_with_status
    raise exception.Error("Command failed: \n # %s\n%s" % (cmd_pretty(command, env), output), child.returncode)
mockbuild.exception.Error: Command failed:
[...]

PR #1301, opened 2024-02-01
Enable Packit + Fedora Testing Farm CI

praiskup:
- Behave tests now remove dnf5 by default first as the dnf5 packgae is installed by default with `dnf install mock` (brings in weak dependencies).

- Build with Packit also into the stable Fedora, so we can later test on stable fedora.

- When run with TMT (not only in Testing Farm), we don't generate the Mock package with Tito anymore, but we download it pre-built from Copr.

- Packit config uses 'merge_pr_in_ci: false' because we don't want to have "git sha" mismatch like in https://github.com/packit/packit-service/issues/2329

- Packit: move the `actions:` configuration top-level.

- Apply work-around for test failure #1300.

- Packit: start building also for pushes.

Relates: https://gitlab.com/testing-farm/tests/-/issues/2
Fixes: #1258


testing-farm/tests/old-testsuite/test.sh
```diff
@@ -0,0 +1,28 @@
...
+# Install the tested RPMs
+install-mock-packages-built-by-packit mock-core-configs mock
+
+# Download the tested SRPM
+SRPM_DOWNLOAD_DIR=/tmp/mock-test-srpms install-mock-packages-built-by-packit mock
+
+cd "$workdir/mock"
+
+# shellcheck disable=SC2024
+if (sudo -E -u mockbuild make check > >(tee the-log | grep -e FAILED: -e PASSED:) 2>&1) >&2; then
```
xsuchy, line 22: I believe that it is preferred to use `runuser` in the script over `sudo`. But that is nitpick.

xsuchy (approved):
LGTM - and I have learned several new things.

Issue #1310, opened 2024-02-06
Config parser error


praiskup:
`WARNING: '%' must be followed by '%' or '(', found: '%40fedora-llvm-team/llvm-compat-packages/fedora-rawhide-x86_64/'`

From https://github.com/fedora-copr/copr/issues/3132.

```
$ cat child.cfg 
include('/etc/mock/fedora-rawhide-x86_64.cfg')

config_opts.setdefault('plugin_conf', {})
config_opts['plugin_conf'].setdefault('tmpfs_opts', {})
config_opts['plugin_conf']['tmpfs_opts']['keep_mounted'] = True


config_opts['root'] = 'test-me'
config_opts['chroot_additional_packages'] = 'llvm-snapshot-builder'


config_opts['rpmbuild_networking'] = True
config_opts['use_host_resolv'] = True

config_opts['macros']['%copr_username'] = '@fedora-llvm-team'
config_opts['macros']['%copr_projectname'] = 'llvm-snapshots-incubator-20240206'
config_opts['macros']['%vendor'] = 'Fedora Copr - group @fedora-llvm-team'
config_opts['macros']['%buildtag'] = '.copr6987915'
config_opts['macros']['%__urlhelper_localopts'] = '--proto -all,+https,+ftps'


config_opts[f"{config_opts.package_manager}.conf"] += """

[copr_base]
name='Copr repository'
baseurl=https://download.copr.fedorainfracloud.org/results/@fedora-llvm-team/llvm-snapshots-incubator-20240206/fedora-rawhide-x86_64/
gpgcheck=0
enabled=1
skip_if_unavailable=0
metadata_expire=0
cost=1
best=1
[https_download_copr_fedorainfracloud_org_results_40fedora_llvm_team_llvm_compat_packages_fedora_rawhide_x86_64]
name='Additional repo https_download_copr_fedorainfracloud_org_results_40fedora_llvm_team_llvm_compat_packages_fedora_rawhide_x86_64'
baseurl=https://download.copr.fedorainfracloud.org/results/%40fedora-llvm-team/llvm-compat-packages/fedora-rawhide-x86_64/
gpgcheck=0
enabled=1
skip_if_unavailable=0
metadata_expire=0
cost=1
best=1
"""
```

PR #1311, opened 2024-02-06
package_manager: disable %-interpolation in dnf.conf parser

praiskup:
This fixes parsing of dnf.conf containing url-encoded links.

Fixes: #1310


xsuchy (approved):
+1

PR #1321, opened 2024-02-13
Fix broken attempt to handle --install completion

opoplawski:
Attempt at fixing #1279 

Very sorry about the mess, hopefully this fixes it.  Are there other `greedy` options that can take more than one arguement that we need to worry about?

### DIFF
--- a/mock/etc/bash_completion.d/mock
+++ b/mock/etc/bash_completion.d/mock
@@ -60,7 +60,7 @@ _mock()
         count=$((++count))
     done
 
-    case "$prevopt" in
+    case "$prev" in
         -h|--help|--version)
             # no further arguments are accepted after the above arguments
             return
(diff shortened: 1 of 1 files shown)


praiskup:
I submitted PR #1320 a few hours earlier; do you think this one is better?

praiskup:
> Are there other greedy options that can take more than one arguement that we need to worry about?

I think there are a few of them.  See #1320.

There's also a fix for:
`mock --install x86_64/foo-20240214_1237-1.fc39.x86_64.rpm -r fedora-rawhide-x86_64 x86_64/<tab>`

praiskup:
And the implicit `--rebuild` mode.

praiskup:
Closing, but thank you for the PR.

PR #1322, opened 2024-02-14
Fix root_cache invalidation triggered by config changes

praiskup:
The root_cache plugin assures that no new configuration file in the `config_opts['config_paths']` list is newer than the root_cache cache tarball, otherwise the caches are invalidated.

The bug is caused by use of improper set operation (union() doesn't modify the calling object).


xsuchy (approved):
+1

hroncok:
Does this invalidate the bootstrap chroot as well? I recall I needed to drop it when rawhide was updated to dnf5.

Issue #1234, opened 2023-09-26
Release Mock 5.2


praiskup:
- [ ] #1188
- [x] #1233
- [x] #1229 
- [x] https://github.com/rpm-software-management/mock/pull/1235

praiskup:
Done https://rpm-software-management.github.io/mock/Release-Notes-5.2

The ODCS fix is skipped for now, seems like a configuration error.

PR #1325, opened 2024-02-14
ChrootScan: don't fail for missing option

praiskup:
I just tried to comment-out the 'tarball' option, and Mock started to fail with KeyError traceback.  It's better to use config_opts.get().

While on it, fix documentation to suggest configuration that is not prone to this kind of errors (the default values stay defined).

### DIFF
--- a/mock/py/mockbuild/plugins/chroot_scan.py
+++ b/mock/py/mockbuild/plugins/chroot_scan.py
@@ -40,11 +40,11 @@ class ChrootScan(object):
 
     def _only_failed(self):
         """ Returns boolean value if option 'only_failed' is set. """
-        return str(self.scan_opts['only_failed']) == 'True'
+        return str(self.scan_opts.get('only_failed')) == 'True'
 
     def _tarball(self):
         """ Returns boolean value if option 'write_tar' is set. """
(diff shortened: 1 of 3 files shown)


mock/py/mockbuild/plugins/chroot_scan.py
```diff
@@ -40,11 +40,11 @@ def __init__(self, plugins, conf, buildroot):
...
     def _only_failed(self):
         """ Returns boolean value if option 'only_failed' is set. """
-        return str(self.scan_opts['only_failed']) == 'True'
+        return str(self.scan_opts.get('only_failed')) == 'True'
 
     def _tarball(self):
-        """ Returns boolean value if option 'write_tar' is set. """
-        return str(self.scan_opts['write_tar']) == 'True'
+        """ Returns boolean value if option 'tarball' is set. """
+        return str(self.scan_opts.get('tarball')) == 'True'
```
praiskup, line 47: nah write_tar

tkopecek:
LGTM

praiskup:
Again, merging early as a quick follow-up for #1234.

praiskup:
Thank you for checking!

PR #1332, opened 2024-02-16
Use dnf5 on Fedora 40+

hroncok:
Fixes https://bugzilla.redhat.com/2264535

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-branched.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-branched.tpl
@@ -8,7 +8,10 @@ config_opts['chroot_setup_cmd'] = 'install @{% if mirrored %}buildsys-{% endif %
 
 config_opts['dist'] = 'fc{{ releasever }}'  # only useful for --resultdir variable subst
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
-config_opts['package_manager'] = 'dnf'
+
+# https://fedoraproject.org/wiki/Changes/BuildWithDNF5 for Fedora 40+
+config_opts['package_manager'] = '{% if releasever|int >= 40 %}dnf5{% else %}dnf{% endif %}'
+
 config_opts['bootstrap_image'] = 'registry.fedoraproject.org/fedora:{{ releasever }}'
 
 config_opts['dnf.conf'] = """
--- a/releng/release-notes-next/f40-dnf5.config
+++ b/releng/release-notes-next/f40-dnf5.config
@@ -1,0 +1,5 @@
+Per the approved [Fedora 40 change](https://fedoraproject.org/wiki/Changes/BuildWithDNF5),
+[we switched][PR#1332] the default `package_manager` configuration
+for Fedora 40 or newer to `dnf5`.
+This was previously done when Fedora 40 was Rawhide,
+but it [regressed][rhbz#2264535] when Fedora 40 branched.


hroncok:
Alternatively, this can be hardcoded in all the fedora-40 configs.

Conan-Kudo (approved):

praiskup (approved):

hroncok:
working on changelog entry

praiskup:
Hm, actually ... have you tried it works?

mock-core-configs/etc/mock/templates/fedora-branched.tpl
```diff
@@ -8,7 +8,14 @@ config_opts['chroot_setup_cmd'] = 'install @{% if mirrored %}buildsys-{% endif %
 
 config_opts['dist'] = 'fc{{ releasever }}'  # only useful for --resultdir variable subst
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+
+{% if releasever | int >= 40 %}
+# https://fedoraproject.org/wiki/Changes/BuildWithDNF5
+config_opts['package_manager'] = 'dnf5'
```
praiskup, line 14: I don't think we can use Jinja like this :-/

There should be something like `config_opts["package_manager"] = "{% if relasever|int >= 40 %}dnf5{% else %}...`

hroncok: Why cannot we?

praiskup: The Jinja is being procesed inside `config_opts` fields. The whole file is not Jinja processed, it is a plain Python file.

hroncok:
ERROR: Config error: /etc/mock/fedora-39-x86_64.cfg: invalid syntax (<string>, line 15)

mock-core-configs/etc/mock/templates/fedora-branched.tpl
hroncok: I thought this entire file was a jinja template. But apparently, it is not. Sorry for the confusion.

hroncok:
Tested with F40 and F39 and it works as expected.

hroncok:
For the record, the first attempt created an invalid config and no CI job seems to have noticed that. Perhaps a test should be added that the configs are at least parse-able?

mock-core-configs/etc/mock/templates/fedora-branched.tpl
hroncok: For the record, the .tpl filename suffix made me assume the file is a jinja template.

praiskup:
@hroncok I can not find the old testing-farm run now :-/ but it already appeared failed/red actually.
In this case, the tests failed because we do most of the testing on Fedora 39 host against the Fedora 38 (branched) configs (current status quo).

But in general, we don't let the test for configs fail the CI.  Some of the configs are known to fail;  you have to take a look manually into the logs to see the failures.

praiskup:
Ad tests, take a look at the [failure here](https://github.com/rpm-software-management/mock/commit/2692b9cba4367d049113d998bd059e3842029ce6).

praiskup:
I can see `Fedora-39:x86_64:/testing-farm/plans/behave        | COMPLETE | OK      | PASSED ` now so the basic tests already passed.  Merging earlier and releasing so we have it in updates-testing ASAP.

hroncok:
This broke my own CI strcipt. We use `config_opts[f"{config_opts.package_manager}.conf"] += ...` which now fails with:

    ERROR: Config error: /tmp/fedora-39-x86_64-ci.cfg: '{% if releasever|int >= 40 %}dnf5{% else %}dnf{% endif %}.conf'

Is this something we should try to fix here, or was this never actually supported?

xsuchy:
This is evaluated by https://pypi.org/project/templated-dictionary/ so it should work.

Can you open a new issue and put there your whole config so we can reproduce it?

hroncok:
https://github.com/rpm-software-management/mock/issues/1357

praiskup:
> Is this something we should try to fix here, or was this never actually supported?

We e.g. did this in Copr: https://github.com/fedora-copr/copr/pull/3152

hroncok:
I did the same: https://src.fedoraproject.org/rpms/pyproject-rpm-macros/c/fa7c91c65dcee2b5cc21039533aafee4e52bde23

Issue #1271, opened 2023-12-13
RFE: Provide an explicit package_manager name for dnf 4


hroncok:
Currently, to use dnf 4, one would set:

```python
config_opts["package_manager"] = "dnf"
```

@befeleme said (in other words): It's not clear what this means, can we say "dnf 4" more explicitly?

I've checked the code in mock:

https://github.com/rpm-software-management/mock/blob/cec199ef78a089d0aeb76817840bc86120fcdf60/mock/py/mockbuild/package_manager.py#L28-L37

Apparently, we cannot.

Would you please add an alias (or similar) so we can do:

```python
config_opts["package_manager"] = "dnf4"
```

Or:

```python
config_opts["package_manager"] = "dnf-3"
```

Thanks.

Conan-Kudo:
We can add `dnf4`, that's what we did in KIWI for this a while back.

praiskup:
While working on pr #1335, I'm not confident this DNF => DNF4 rename in Mock is the natural thing to do.  What do others think?  CC @jan-kolarik from DNF.

DNF always was named just DNF before, only by "accident" the last version is 4.X (in 2018, there was [v3](https://github.com/rpm-software-management/dnf/commit/197901eeaa5619f00f2d11ef1c9e8d78dca98df0), too).

The DNF v5 is aliased as DNF5 (everyone uses this), but it has its own versioning scheme (now v5.2.X).  But nothing prevents the DNF5 upstream from raising the version to v6.X in the future.

Also, at some distant point in the future, we'd like to call DNF5 just DNF because nobody will care about DNF4.  Both here in mock, and also in general. :shrug:

hroncok:
In my head it's called dnf 4. Whether or not that is what upstream dnf wants I don't know. Perhaps it should be "legacy dnf" or something else entirely, but we would need to ask the dnf folks.

jan-kolarik:
Hi guys,

I haven't been around long enough to remember pre-dnf 4.x, but here are my thoughts.

I see **dnf** as simply a reference to the default package manager version (4 or 5) relevant to a system. With the packages now separated, referring to just **dnf** will cause confusion.

In this context, it's clearer to use the specific dnf version, as mock doesn't have a "dnf default".

Regarding versioning, as we agreed in the team, dnf 4.x will not have major version bumps anymore, and so will dnf5. That's why we moved to the 5.X.Y.Z versioning scheme.

PR #1335, opened 2024-02-16
Make 'package_manager = dnf4' equivalent to 'package_manager = dnf'

praiskup:
Fixes: https://github.com/rpm-software-management/mock/issues/1271

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -18,10 +18,11 @@ from .trace_decorator import traceLog, getLog
 from .mounts import BindMountPoint
 
 fallbacks = {
-    'dnf': ['dnf', 'dnf5', 'yum'],
-    'yum': ['yum', 'dnf', 'dnf5'],
-    'microdnf': ['microdnf', 'dnf','dnf5', 'yum'],
-    'dnf5': ['dnf5', 'dnf', 'yum'],
+    'dnf4': ['dnf4', 'dnf5', 'yum'],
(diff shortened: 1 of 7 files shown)


mock/py/mockbuild/package_manager.py
```diff
@@ -671,7 +674,7 @@ def _check_missing(output):
 
 
 class Dnf(_PackageManager):
-    name = 'dnf'
+    name = 'dnf4'
```
xsuchy, line 677: ???
```
 ls -l /usr/bin/dnf
lrwxrwxrwx. 1 root root 5  3. čec 02.00 /usr/bin/dnf -> dnf-3
```
On my F40.
BTW - what is DNF4? DNF3 is what we have now and DNF5 is the rewrite in C++. Right?

praiskup: I was told that `dnf-3` is a leftover from the Python 2 + 3 times; it was normal to add `-3` suffix for Python 3 executable (and it stayed as-is).  But it's a python, which means it is DNF4.

This `name = dnf4` thus IMO correct at this point.

mock/py/mockbuild/package_manager.py
```diff
@@ -30,7 +31,9 @@ def package_manager_from_string(name):
         return Dnf5
     if name == 'yum':
         return Yum
-    if name == 'dnf':
+    if name == 'dnf4':
+        return Dnf
+    if name == 'dnf': # backward compat
```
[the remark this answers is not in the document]
praiskup: Fixed.

xsuchy (approved):

PR #1337, opened 2024-02-22
Add a new option --scrub-all-chroots

### LINKED ISSUE #521: Add command to remove all existing chroots
I'm using mock for plenty of things, to test packages, to test builds, test my code behavior on a different system versions. Problem is that when you use mock heavily it's easy to get lost in which chroot still exists and your free space will leak this way.

I know you can look on /var/lib/mock and remove that one by one. However, this is not really inconvenient  when you have hundreds of chroots.

Nice solution have podman and buildah implemented already by ``--all`` option:

https://www.mankier.com/1/podman-rm#--all

### Output of `rpm -q mock`

mock-2.0-2.fc31.noarch

### Steps to reproduce issue

1. mock -r fedora-rawhide-x86_64 --init
2. <future> mock --remove-all-chroots

praiskup:
Fixes #521

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -61,23 +61,13 @@ import copy
 
 # pylint: disable=import-error
 from functools import partial
-from mockbuild import config
-from mockbuild import util
-from mockbuild.constants import MOCKCONFDIR, PYTHONDIR, VERSION
-from mockbuild.file_downloader import FileDownloader
-from mockbuild.mounts import BindMountPoint, FileSystemMountPoint
-
-# import all mockbuild.* modules after this.
-sys.path.insert(0, PYTHONDIR)
-
-# set up basic logging until config file can be read
-FORMAT = "%(levelname)s: %(message)s"
-logging.basicConfig(format=FORMAT, level=logging.WARNING)
-log = logging.getLogger()
 
 # our imports
-# pylint: disable=wrong-import-position
-
+from mockbuild import config
+from mockbuild import util
+from mockbuild.constants import MOCKCONFDIR, VERSION
+from mockbuild.file_downloader import FileDownloader
+from mockbuild.mounts import BindMountPoint, FileSystemMountPoint
 import mockbuild.backend
 from mockbuild.backend import Commands
 from mockbuild.buildroot import Buildroot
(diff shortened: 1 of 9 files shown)


xsuchy:
I would rather have it inside of mock.py and utilize the logic we have from `--list-chroots`.

praiskup:
I'm not against modifying `mock.py` per se, but `--list-chroots` doesn't know the previously used `--uniqueext` extensions.  This was rather a toy experiment for now.

praiskup:
@xsuchy does this sound better? :-)

mock/py/mockbuild/scrub_all.py
```diff
@@ -0,0 +1,110 @@
...
+from mockbuild.constants import MOCKCONFDIR
+from mockbuild.config import traverse_chroot_configs
+
+
+def _do_scrub(configs, weird, chroot, suffix=None):
+    if suffix and chroot + "-" + suffix in weird:
+        print(f"skipping weird scrub: {chroot} {suffix}")
+        return
+
+    base_cmd = ["mock", "--scrub=all", "-r"]
```
xsuchy, line 19: ??? Now you can call the function directly. Or is there an obstacle?

xsuchy:
Otherwise mucho bettero.

mock/py/mockbuild/scrub_all.py
praiskup: That would mean calling the following multiple times:

```python
commands = Commands(config_opts, uidManager, plugins, state, buildroot, bootstrap_buildroot)
commands.scrub(all) 
```
Which is _a lot_ of work (manpower, to code properly all the arguments)  :-/ my opinion is :shrug: keep it as is this is mostly cosmetics.  Admitedly that's not nice, but still better than nothing.

praiskup:
This deserves some documentation work, if acceptable at all.  @xsuchy lemme know what you think so I could finalize this PR.

praiskup:
Please take a look.

mock/py/mock.py
```diff
@@ -87,6 +87,7 @@
 from mockbuild.state import State
 from mockbuild.trace_decorator import traceLog
 import mockbuild.uid
+from mockbuild.scrub_all import scrub_all_chroots
```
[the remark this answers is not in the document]
praiskup: Ok, fixing this was easy, actually (separate commit).

mock/py/mockbuild/scrub_all.py
praiskup: Added the FIXME note at least.

xsuchy (approved):
+1

PR #1339, opened 2024-02-26
docs: add the case of user's default.cfg.

junaruga:
I am using Fedora. I was looking for a way of using an user's own `default.cfg` instead of the `/etc/mock/default.cfg`. Because I wanted to use the Fedora rawhide as default instead of the current default to the latest Fedora stable version 39. 

```
$ ls /etc/mock/default.cfg -l
lrwxrwxrwx. 1 root root 20 Feb 26 14:57 /etc/mock/default.cfg -> fedora-39-x86_64.cfg
```

I didn't want to modify the `/etc/mock/default.cfg`. Because it might prevent the future update of the file by upgrading the `mock-core-configs` RPM package.

I was not able to find the user's `default.cfg` in the current documents. Below is a part of explaining the `default.cfg`. But it doesn't explain about the user's `default.config` explicitly.

https://github.com/rpm-software-management/mock/blob/fc1f7d4d22777d17f41aa6482e627f1895258110/mock/docs/mock.1#L59-L69

So, I finally found a way to set up the user's `default.cfg`. That is one of either ways below.

```
$ ln -s /etc/mock/fedora-rawhide-x86_64.cfg ~/.config/mock/default.cfg
```

or 

```
$ cp /etc/mock/fedora-rawhide-x86_64.cfg ~/.config/mock/default.cfg
```

So, this PR is my proposal to update the current document to add this topic. You can see the modified `configuration.md` [here](https://github.com/junaruga/mock/blob/wip/doc-user-default-cfg/docs/configuration.md) on my forked repository.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,6 +78,13 @@ You may also copy and edit an existing configuration file into a new one:
 
     $ cp /etc/mock/fedora-rawhide-x86_64.cfg ~/.config/mock/foo.cfg
 
+The default chroot configuration file is `/etc/mock/default.cfg`, which is
+usually a symlink to one of the installed chroot configuration files.  You may
+create another symlink to an installed configuration file to change the default
+chroot config file:
+
+    $ ln -s /etc/mock/fedora-rawhide-x86_64.cfg ~/.config/mock/default.cfg
+
 If Koji is already using a config you need, then you can use the Koji client
 tool for generating the file:
 


docs/configuration.md
```diff
@@ -74,8 +74,10 @@ such file with an updated version of the package).  It is safer to install an
     # install make into the minimal build chroot
     config_opts['chroot_additional_packages'] = 'make'
 
-You may also copy and edit an existing configuration file into a new one:
+You may also copy and edit an existing configuration file including the
+`default.cfg` into a new one:
 
+    $ cp /etc/mock/fedora-rawhide-x86_64.cfg ~/.config/mock/default.cfg
```
praiskup, line 80: This seems a bit confusing when it is "out of context".  Can you please create a separate paragraph describing `default.cfg` semantics (similar to the manpage)? And perhaps document _there_ how to define a per-user default.cfg in $HOME?

Btw., note that we plan to drop the default.cfg symlink by default #563.

praiskup:
Thank you for the patch, @junaruga

docs/configuration.md
junaruga: OK. I rebased the PR as your suggestion.

junaruga:
I rebased this PR as your suggestion on the latest main branch. You can check the modified document [here](https://github.com/junaruga/mock/blob/wip/doc-user-default-cfg/docs/configuration.md). I referred to the mock's man page mock(1).

praiskup (approved):
LGTM, thank you!

Issue #1338, opened 2024-02-22
RFE: automatically import key for F(N+1) when building for rawhide


keszybz:
Every time we branch, there is a short period when local configuration thinks Rawhide==FN, but actually Rawhide==F(N+1), and gpg key verification fails. After branching, the packages in the F(N+1) compose are resigned using the F(N+1) key, so they do not pass verification with FN key.

It turns out that it's fairly easy to avoid this problem by importing the additional key.
I did such a change for mkosi and it solves the issue [1].

[1] https://github.com/systemd/mkosi/commit/f221562c945a48db9384f8521f67b9b02cd71ac1

<details>

```
Running transaction
Importing PGP key 0xA15B79CC:
 Userid     : "Fedora (40) <fedora-40-primary@fedoraproject.org>"
 Fingerprint: 115DF9AEF857853EE8445D0A0727707EA15B79CC
 From       : file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-40-primary
The key was successfully imported.
Importing PGP key 0xA15B79CC:
 Userid     : "Fedora (40) <fedora-40-primary@fedoraproject.org>"
 Fingerprint: 115DF9AEF857853EE8445D0A0727707EA15B79CC
 From       : file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-40-primary
The key was successfully imported.
Importing PGP key 0x18B8E74C:
 Userid     : "Fedora (39) <fedora-39-primary@fedoraproject.org>"
 Fingerprint: E8F23996F23218640CB44CBE75CF5AC418B8E74C
 From       : file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-39-primary
The key was successfully imported.

Transaction failed: Signature verification failed.
PGP check for package "curl-8.6.0-6.fc40.x86_64"
(/var/lib/mock/fedora-rawhide-x86_64/root/var/cache/dnf/fedora-2d95c80a1fa0a67d/packages/curl-8.6.0-6.fc40.x86_6
4.rpm) from repo "fedora" has failed: Import of the key didn't help, wrong key?
```
</details>

xsuchy:
We already do that. For long time. https://github.com/rpm-software-management/mock/blob/main/mock-core-configs/etc/mock/templates/fedora-rawhide.tpl#L45

keszybz:
IIUC, that is doing F(N-1), but we actually need F(N+1).
[...]

Issue #1328, opened 2024-02-15
Could mock-core-configs Fedora Rawhide use fedora-distro-aliases?


praiskup:
We always have to wrap new `mock-core-configs` with new branched Fedora around
the time of Fedora branching event (like we did yesterday).  I'm curious if we could
implement some interesting feature into `mock` or `mock-core-configs`, perhaps using
the [fedora-distro-aliases](https://github.com/rpm-software-management/fedora-distro-aliases) project, so that this is done automatically without a new
release?

praiskup:
see also https://pagure.io/fedora-infrastructure/issue/11809

praiskup:
see also https://github.com/rpm-software-management/fedora-distro-aliases/issues/9

praiskup:
see also https://github.com/rpm-software-management/fedora-distro-aliases/issues/11

praiskup:
I'm a bit skeptical with the `fedora-distro-aliases` right now, it adds a single
point of failure on external online system (Bodhi) for every Mock/Copr user.
Every build (quite a lot of API queries, actually).

praiskup:
Since everything seems to be interlinked now, note that this issue is a different RFE from #1338.
The #1338 is rather a Fedora vs. Mock's upstream communication/synchronization problem that
has rather an easy solution.

This issue is rather about solving the issue in a much more "generic" way.  The idea here would
be basically to drop all the `fedora-NN-*` configuration files, and rely on a reliable source giving
use the necessary constants.

praiskup:
Outage reported here: https://matrix.to/#/!rALCnapvdNTxUTuENl:matrix.org/$wlFOALMr7CPoZPpqg8DE2QCTRaPKWeOrOa6PZSVhSKQ?via=fedora.im&via=matrix.org&via=catgirl.cloud

PR #1342, opened 2024-03-02
configs: Rawhide to accept GPG key from future Fedora Rawhide+1

praiskup:
Fixes: #1338


mock-core-configs/etc/mock/templates/fedora-rawhide.tpl
```diff
@@ -42,7 +42,7 @@ user_agent={{ user_agent }}
 
 {%- macro rawhide_gpg_keys() -%}
 file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-$releasever-primary
-{%- for version in [releasever|int, releasever|int - 1]
+{%- for version in [releasever|int, releasever|int - 1, releasever|int + 1]
```
keszybz, line 45: Thanks! My only qualm is what happens if the key doesn't exist yet? We don't want any kind of warning or error in that case. Hmm, I think we get an error if they key doesn't exist:
```
cannot open file: (2) - No such file or directory [/usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-46-primary]
```

I think this should be wrapped in a check whether the key actually exists.

praiskup: Good point, I installed this fatal error into the branching script:

```
File /usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-43-primary not found!

The new Fedora Rawhide (42) requires a GPG key for Fedora 43,
which must be generated by Fedora Infrastructure team and shipped in the
distribution-gpg-keys.rpm package.  This either has not happened yet, the
package is outdated, or it is not even installed on this box.
```

praiskup:
This time, F42 key was generated [on 2024-02-12](https://src.fedoraproject.org/rpms/fedora-repos/c/d10a092adbcf0e0b2d7e4f05311255c805c878b5).
The plan was to have it about [~14 days before signing rawhide with f41 key](https://fedorapeople.org/groups/schedule/f-40/f-40-all-tasks.html), which IMO happened (the mass rebuild took a bit more time). Not sure.
Mock-core-configs with branched F40 [2024-02-14](https://github.com/rpm-software-management/mock/releases/tag/mock-core-configs-40.1-1). Which was admittedly too late, it would be nice to do that 14 days earlier next time.

xsuchy:
I have to say that I am not happy for this PR. IMO this will bring a new issue when we submit new config earlier than before branching (missing key). 
I would rather resolve it by dynamically using number for rawhide. Likely using python-fedora-distro-aliases https://github.com/rpm-software-management/fedora-distro-aliases  In such case we would not need to touch rawhide config after branching at all!

praiskup:
"fedora-repos update with the new keys is available in updates-testing" is though something different, I think that we should have a new entry in the F41 schedule like

    Generate Fedora N+2 key and ship in fedora-repos

Somewhere in the Fedora 41 schedule.  To have a guidance that assures everything works.

praiskup:
> I have to say that I am not happy for this PR. IMO this will bring a new issue
> when we submit new config earlier than before branching (missing key).

If we can get a new step into the Fedora release schedule, then this problem
would be non-existing.

> I would rather resolve it by dynamically using number for rawhide. Likely
> using python-fedora-distro-aliases
> https://github.com/rpm-software-management/fedora-distro-aliases In such case
> we would not need to touch rawhide config after branching at all!

This is being tracked here #1328 - though :-/ it's going to be a big change in
the way mock and core-configs behave.

xsuchy:
I tried this with DNF4 and DNF5 and it prints no error when the file does not exists and it is silently ignored. Not sure if this is generally good, but for our case it is good.
I changed my mind and I am +1

xsuchy:
I manually rebased this and merged as 66ca1e4b

PR #1346, opened 2024-03-09
Rewrites "original_name" method to check if "cls" has a "backmap"

nikitych:
In some cases, such as fail of "mock --chain --recurse some.src.rpm.tofail", FileDownloader "cls" may not have a "backmap" attribute. So we have to check if it exists before accessing it.

Fixes: #1345


praiskup:
Thank you!

PR #1348, opened 2024-03-20
Add initial c10s mock configs (Koji BUILDROOT only)

hroncok:


hroncok:
Koji also has i686 but there are no such c9s configs here, so I did not add them.

FrostyX:
I tried to build my package using these configs and it worked fine.

> name=CentOS Stream $releasever - Koji Local - BUILDROOT ONLY!

I still cannot see 10-stream here https://mirror.stream.centos.org/ so +1 for the PR as is.

praiskup:
Thank you!

PR #1353, opened 2024-03-31
bash-completion: add mock-parse-buildlog

pastalian:
I have a feeling that no one is using mock-parse-buildlog, but we are installing `%{_datadir}/bash-completion/completions/mock-parse-buildlog`, which does nothing right now.


praiskup (approved):
LGTM

praiskup:
thank you

PR #1355, opened 2024-04-04
scm: Use bytes for compatibility with subprocess stdout

dm0-:
I am getting the following type error when attempting to use SCM with Git timestamps on RHEL 9 and Fedora.  Can something like this be applied to fix it?

```
...
  File "/usr/lib/python3.9/site-packages/mockbuild/scm.py", line 118, in adjust_git_timestamps
    for f in proc.communicate()[0].split('\0')[:-1]:
TypeError: a bytes-like object is required, not 'str'
```

### DIFF
--- a/mock/py/mockbuild/scm.py
+++ b/mock/py/mockbuild/scm.py
@@ -114,14 +114,18 @@ class scmWorker(object):
         cwd_dir = util.pretty_getcwd()
         self.log.debug("Adjusting timestamps in %s", self.src_dir)
         os.chdir(self.src_dir)
-        proc = subprocess.Popen(['git', 'ls-files', '-z'], shell=False, stdout=subprocess.PIPE)
+        proc = subprocess.Popen(
+            ['git', 'ls-files', '-z'],
+            shell=False, stdout=subprocess.PIPE, universal_newlines=True,
+        )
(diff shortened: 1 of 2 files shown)


mock/py/mockbuild/scm.py
```diff
@@ -115,14 +115,14 @@ def adjust_git_timestamps(self):
         self.log.debug("Adjusting timestamps in %s", self.src_dir)
         os.chdir(self.src_dir)
         proc = subprocess.Popen(['git', 'ls-files', '-z'], shell=False, stdout=subprocess.PIPE)
-        for f in proc.communicate()[0].split('\0')[:-1]:
+        for f in proc.communicate()[0].split(b'\0')[:-1]:
```
praiskup, line 118: Would you mind using `Popen(..., universal_newlines=True)` instead?

praiskup: ... thank you for the patch!

dm0-: Okay, it's updated, thanks.

praiskup:
Nice update, thank you.  Can you please fill in the ChangeLog entry?  Otherwise LGTM

praiskup (approved):
LGTM

PR #1367, opened 2024-04-28
Enable RPM sysusers integration 

nikromen:
Provide users/groups via `/usr/lib/sysusers.d`. This is going
to be handled by RPM as described in
https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation
and https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/NEFOV236FJYS2RED2SEOV5YHDFLDX7DK/#KB6KS3U7RO4AYANGSUNVSG7UF5AW52Q2

Fixes https://github.com/fedora-copr/copr/issues/2789


xsuchy:
what about %files section?

mock-core-configs/mock-core-configs.spec
```diff
@@ -51,6 +51,9 @@ Rocky Linux and various other specific or combined chroots.
 
 %install
 mkdir -p %{buildroot}%{_sysusersdir}
+%if 0%{?fedora} > 39 || 0%{?rhel} > 8
+cp -a mock.conf %{buildroot}%{_sysusersdir}
+%endif
```
praiskup, line 56: What happens with the `%pre filesystem` section in `mock.spec` on F40 and EL9+?

xsuchy:
I think this still miss a few things: BuildRequires and %pre macro. See https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation

xsuchy:
Additionally, it would be nice if you amend https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation and for package that provides the user do s/mock/mock-core-configs/.

mock-core-configs/mock-core-configs.spec
praiskup: Hmmm, since which Fedora version is this feature supported?

praiskup:
> Additionally, it would be nice if you amend https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation and for package that provides the user do s/mock/mock-core-configs/.

.. or `mock-filesystem`.

mock-core-configs/mock-core-configs.spec
praiskup: Actually, in an UBI8 container:
```
# rpm --eval %_sysusersdir
/usr/lib/sysusers.d
```
Could the feature work even there?

xsuchy:
Why you removed that line from %files section?

nikromen:
The %sysusers_create_compat macro will generate something like this:

```sh
# generated from mock.conf.
getent group 'mock' >/dev/null || groupadd -r 'mock' || :
getent passwd 'mock' >/dev/null || \
    useradd -r -g 'mock' -d '/run/mock' -s '/sbin/nologin' -c 'Runs Uid '\''N'\'' Gid Emporium' 'mock' || :
```

so the line in files is no longer needed

xsuchy:
I meant:
```
%files
%{_sysusersdir}/mock.conf
```

mock/mock.spec
```diff
@@ -257,6 +259,7 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 %config(noreplace) %{_sysconfdir}/%{name}/*.ini
 %config(noreplace) %{_sysconfdir}/pam.d/%{name}
 %config(noreplace) %{_sysconfdir}/security/console.apps/%{name}
+%{_sysusersdir}/mock.conf
```
praiskup, line 262: Well, I know @xsuchy requested this, but ... do we need this?

mock/mock.spec
```diff
@@ -67,6 +66,7 @@ BuildRequires: python%{python3_pkgversion}-pylint
 %endif
 BuildRequires: python%{python3_pkgversion}-rpm
 BuildRequires: python%{python3_pkgversion}-rpmautospec-core
+BuildRequires:  systemd-rpm-macros
```
praiskup, line 69: I still don't understand why we are moving this logic from `mock-filesystem` to `mock-core-configs`, at least documentation is needed - but it seems wrong :shrug:

mock/mock.spec
praiskup: Ah, not needed by the scriptlet itself - but it is used by systemd (or the deamon) which may take care of useradd/groupadd.

praiskup:
I tried to install on UBI8:
```
  Installing       : mock-5.6.post1-1.git.3754.50829bf.el8.noarch                                                                                                                                                                     52/53 
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root

  Installing       : mock-core-configs-40.6-1.git.3754.50829bf.el8.noarch                                                                                                                                                             53/53 
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
warning: group mock does not exist - using root
...
```

mock/mock.spec
xsuchy: +1

nikromen: thanks for noticing, I did this firstly in mock-core-config but then moved it to mock-filesystem. This is just a remnant I forgot to remove

mock/mock.spec
```diff
@@ -213,13 +216,15 @@ install -d %{buildroot}/var/cache/mock
...
+install -p -D -m 0644 mock.conf %{buildroot}%{_sysusersdir}
+
 sed -i 's/^_MOCK_NVR = None$/_MOCK_NVR = "%name-%version-%release"/' \
     %{buildroot}%{_libexecdir}/mock/mock
 
 %pre filesystem
-# check for existence of mock group, create it if not found
-getent group mock > /dev/null || groupadd -f -g %mockgid -r mock
-exit 0
+echo "je tu nejaky skript nebo ne"
```
praiskup, line 226: Very likely there is:
```
rpm --eval "%sysusers_create_compat ./mock-core-configs/mock.conf"

# generated from mock.conf
getent group 'mock' >/dev/null || groupadd -f -g '135' -r 'mock' || : 
```

xsuchy: :] I hope this do not get in final version.

mock/mock.spec
```diff
@@ -17,7 +16,8 @@ License: GPL-2.0-or-later
 # cd mock
 # git reset --hard %%{name}-%%{version}
 # tito build --tgz
-Source: %{name}-%{version}.tar.gz
+Source0: %{name}-%{version}.tar.gz
+Source1: https://raw.githubusercontent.com/rpm-software-management/mock/main/mock-core-configs/mock.conf
```
xsuchy, line 20: ???

praiskup: This is the way to address the file in %pre when macros are evaluated (before the tarball is extrected).

xsuchy: I think you are confusing `%pre` and `%prep`.

praiskup: No no ... the %sysusers_create_compat macro is evaluated at the time when the tarball isn't extracted, yet.  See how it is defined, it contains the shell-script macro `%()`.

xsuchy: So this is bad?
https://src.fedoraproject.org/rpms/bodhi-server/blob/rawhide/f/bodhi-server.spec#_125

praiskup: According to my experiments, ... yes?  But maybe I'm doing something wrong.

praiskup: Yeah, I tried to install bodhi-server in fedora:rawhide container, and there's no `bodhi` user/group:
```
[root@db5467446600 /]# grep bodhi /etc/group /etc/passwd
[root@db5467446600 /]# 
```
The reason nobody noticed is that they (probably) install it on servers with systemd.

nikromen: they also try to work around this by defining the file directly in spec
https://src.fedoraproject.org/rpms/bodhi-server/blob/rawhide/f/bodhi-server.spec#_84

it seems like hard task to push the .sysusers file from upstream to `%pre`...

praiskup: That is wrong approach, the -f file needs to contain *shell script*, not the sysusersdir content.

mock/mock.spec
```diff
@@ -169,6 +169,8 @@ done
 
 ./precompile-bash-completion "mock.complete"
 
+%sysusers_create_compat mock.conf >> sysusers_script
```
praiskup, line 172: You need to call the shell script baked in `%sysusers_create_compat`, not the macro itself.

releng/release-notes-next/sysusers-handled-by-rpm.feature
```diff
@@ -0,0 +1,6 @@
+This provides users/groups via /usr/lib/sysusers.d. This is going
```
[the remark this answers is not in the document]
Conan-Kudo: This is not for homed.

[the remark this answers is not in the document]
praiskup: https://www.freedesktop.org/software/systemd/man/latest/systemd-sysusers.html

mock/mock.spec
```diff
@@ -169,6 +169,9 @@ done
 
 ./precompile-bash-completion "mock.complete"
 
+# this is what %sysusers_create_compat will expand to
+%{_rpmconfigdir}/sysusers.generate-pre.sh mock.conf >> sysusers_script
```
praiskup, line 173: nit: I'd prefer `s/>>/>/`

praiskup:
The release notes snippet deserves update, otherwise LGTM

mock/mock.spec
```diff
@@ -169,6 +169,9 @@ done
 
 ./precompile-bash-completion "mock.complete"
 
+# this is what %sysusers_create_compat will expand to
```
praiskup, line 172: Macro in comment

praiskup:
Hmpf, now I recall I've seen [this](https://rpm-software-management.github.io/rpm/manual/users_and_groups.html) feature before, and guess what :)
```
[root@a7cd3c91a144 /]# rpm -q --provides mock-5.6.post1-1.git.3755.7d754b2.fc41.noarch | grep group
group(mock) = ZyBtb2NrIDEzNSAt
```

mock/mock.spec
```diff
@@ -257,6 +262,7 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 %config(noreplace) %{_sysconfdir}/%{name}/*.ini
 %config(noreplace) %{_sysconfdir}/pam.d/%{name}
 %config(noreplace) %{_sysconfdir}/security/console.apps/%{name}
+%{_sysusersdir}/mock.conf
```
praiskup, line 265: Hmm, shouldn't we keep this in mock-filesystem?

praiskup: Shall this be `%config`?

mock/mock.spec
```diff
@@ -169,6 +169,9 @@ done
 
 ./precompile-bash-completion "mock.complete"
 
+# this is what %%sysusers_create_compat will expand to
+%{_rpmconfigdir}/sysusers.generate-pre.sh mock.conf > sysusers_script
```
praiskup, line 173: I eventually think that we do not need this `%pre` dance at all for systems with RPM >= 4.19. I'd use something like:
```
%if 0%{?fedora} && ( 0%{?rhel} <= 9 || ... others older distros may come ... )
%{_rpmconfigdir}/sysusers.generate-pre.sh mock.conf > sysusers_script
```

praiskup:
Interesting reading:
- https://github.com/rpm-software-management/rpm/blob/7ae180573869908dceab88942b4b18424c2144d6/lib/tagexts.c#L989
- https://github.com/rpm-software-management/rpm/blob/7ae180573869908dceab88942b4b18424c2144d6/lib/psm.c#L290

mock/mock.spec
praiskup: Actually, as I realized in the other thread, the feature is now RPM built-in. But it still doesn't work for the bodhi-server :-) confusing as hell.  I think the problem actually is that the dropin has `.sysusers` suffix instead of `.conf`.  I'm trying to propose [a fix here](https://src.fedoraproject.org/rpms/bodhi-server/pull-request/3).

mock/mock.spec
praiskup: Without this EL10+ condition, we risk a conflict of the "built-in" mechanism and the %pre code.

releng/release-notes-next/sysusers-handled-by-rpm.feature
```diff
@@ -0,0 +1,4 @@
+New Mock RPM package provides the systemd-sysusers drop-in configuration files automatic
+`mock 135` group ID allocation.
+https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation
+
```
praiskup, line 4: nits: Trailing newline, "s/automatic/for automatic/", "s/files/file/"

We are also using this: https://rpm-software-management.github.io/rpm/manual/users_and_groups.html

mock/mock.spec
xsuchy: Still valid.

xsuchy (approved):
+1

Issue #1372, opened 2024-05-15
Add c10s compose to the centos-stream-10 config


hroncok:
https://composes.stream.centos.org/stream-10/development/latest-CentOS-Stream/compose/ exists now

Let's update https://github.com/rpm-software-management/mock/blob/main/mock-core-configs/etc/mock/templates/centos-stream-10.tpl

(I can do this, but not right away, so opening this issue instead.)

PR #1383, opened 2024-06-11
Add composed c10s repos

hroncok:
Fixes https://github.com/rpm-software-management/mock/issues/1372

Depends on: https://github.com/xsuchy/distribution-gpg-keys/pull/119

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
@@ -1,4 +1,4 @@
-config_opts['chroot_setup_cmd'] = 'install @build'
+config_opts['chroot_setup_cmd'] = 'install bash bzip2 centos-stream-release coreutils cpio diffutils findutils gawk glibc-minimal-langpack grep gzip info patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
 config_opts['dist'] = 'el10'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '10'
 config_opts['package_manager'] = 'dnf'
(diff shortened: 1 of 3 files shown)


hroncok:
I have tested this together with https://github.com/xsuchy/distribution-gpg-keys/pull/119 and it works.

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
```diff
@@ -34,7 +34,99 @@ user_agent={{ user_agent }}
...
+[crb]
+name=CentOS Stream $releasever - CRB
+baseurl=https://composes.stream.centos.org/stream-10/production/latest-CentOS-Stream/compose/CRB/$basearch/os
+#metalink=https://mirrors.centos.org/metalink?repo=centos-crb-$stream&arch=$basearch&protocol=https,http
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official-SHA256
+gpgcheck=1
+repo_gpgcheck=0
+metadata_expire=6h
+countme=1
+enabled=0
```
Conan-Kudo, line 111: CRB needs to be enabled.

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
hroncok: Thanks

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
```diff
@@ -34,7 +34,99 @@ user_agent={{ user_agent }}
...
 cost=2000
-enabled=1
+enabled=0
 skip_if_unavailable=False
 
+[baseos]
+name=CentOS Stream $releasever - BaseOS
+baseurl=https://composes.stream.centos.org/stream-10/production/latest-CentOS-Stream/compose/BaseOS/$basearch/os
+#metalink=https://mirrors.centos.org/metalink?repo=centos-baseos-$stream&arch=$basearch&protocol=https,http
+gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official-SHA256
```
Conan-Kudo, line 44: This should use the versioned key path...
```suggestion
gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-$releasever
```

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
hroncok: c9s config doesn't

hroncok: But happy to do that if this is what the mock maintainers want.

praiskup: While it would be cleaner, I don't think it is completely necessary. :shrug:

praiskup:
Can you please bump the distribution-gpg-keys requirement to 1.104 and provide the towncrier snippet?

praiskup:
Per this CI log:
```
Please provide release notes snippet, per the instructions
https://rpm-software-management.github.io/mock/Release-Notes-New-Entry

**Comment by hroncok** [comment]
I can do that later. Or you can amend my PR (I'm OK with that), if you want to merge sooner.

**Comment by davide125** [comment]
This needs to be reworked to point to the mirrors instead of composes now that we have c10s published on the mirrors. It also needs an additional stanza for Extras (which provides SIG release packages), similar to the one we have in the c9s template.

**Comment by hroncok** [comment]
> This needs to be reworked to point to the mirrors instead of composes now that we have c10s published on the mirrors.

Not even the c10s centos-stream-release package has that yet.

https://gitlab.com/redhat/centos-stream/rpms/centos-release/-/blob/c10s/centos.repo?ref_type=heads

> It also needs an additional stanza for Extras (which provides SIG release packages), similar to the one we have in the c9s template.

Where do I find the repo URLs?

---

I disagree this **needs** to be done. We can have gradual improvements. This has been tested and works.

Once the centos-stream-release package is updated with mirrored repos, we can do the same. 

---

@praiskup changelog added, gpg-keys requirement bumped

**Comment by davide125** [comment]
> > It also needs an additional stanza for Extras (which provides SIG release packages), similar to the one we have in the c9s template.
> 
> Where do I find the repo URLs?

https://mirror.stream.centos.org/SIGs/10-stream/extras/x86_64/extras-common/

> I disagree this **needs** to be done. We can have gradual improvements. This has been tested and works.
> 
> Once the centos-stream-release package is updated with mirrored repos, we can do the same.

Totally agree, incremental improvements are the way to go. Re-reading my comment it came out stronger than I meant, sorry about that.

File: mock-core-configs/etc/mock/templates/centos-stream-10.tpl [hunk 3]
```diff
@@ -34,7 +35,89 @@ user_agent={{ user_agent }}
 name=CentOS Stream $releasever - Koji Local - BUILDROOT ONLY!
 baseurl=https://kojihub.stream.centos.org/kojifiles/repos/c{{ releasever }}s-build/latest/$basearch/
 cost=2000
+enabled=0
+
+[baseos]
+name=CentOS Stream $releasever - BaseOS
+#baseurl=http://mirror.stream.centos.org/$releasever-stream/BaseOS/$basearch/os/
+metalink=https://mirrors.centos.org/metalink?repo=centos-baseos-$releasever-stream&arch=$basearch&protocol=https,http
```
praiskup, line 43: There's an experimental commit now using this Metalink URL, which is IMO supposed to work (eventually) but it isn't (yet).  At the same time, the modified baseurl above works -> so I assume there's some temporary mirror synchronization issue?

praiskup:
Sorry for the changelog-change CI failure;  the file clearly _is_ there, but the [changed-files addon](https://github.com/hroncok/mock-1/blob/2ab18702e764169c81be5c7b345c74b456f374a4/.github/workflows/towncrier.yml#L18-L22) doesn't detect it.

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
hroncok: That's why I suggested gradual improvements.

praiskup: But the `baseurl` I provided here works, so I assume it is a matter of hours till the mirrors get synced?

hroncok: I don't know. Maybe, maybe not.

praiskup: I think we can wait till tomorrow or something, or report a mirroring issue.

Is there a reason to use the compose URL instead of the mirror baseurl?

hroncok: That's what the c10s package is using.

davide125: Looks like the metalinks are still being set up: https://gitlab.com/redhat/centos-stream/rpms/centos-release/-/merge_requests/54

praiskup: Seems to work now :partying_face:

praiskup:
I'm for merging the current state.  @hroncok @Conan-Kudo @davide125 wdyt?

Conan-Kudo (approved):

praiskup:
```
PASSED: centos-stream-10-x86_64.cfg (tmpfs)
PASSED: centos-stream-10-x86_64.cfg
PASSED: centos-stream-8-x86_64.cfg (tmpfs)
PASSED: centos-stream-8-x86_64.cfg
PASSED: centos-stream-9-x86_64.cfg (tmpfs)
PASSED: centos-stream-9-x86_64.cfg

hroncok:
Not all repos were converted to metalinks. See https://github.com/rpm-software-management/mock/pull/1390

PR #1385, opened 2024-06-12
Enable CentOS 10, RHEL 10, AlmaLinux 10, Oracle Linux 10, and Circle Linux builds

nkadel:
CentOS 10 has been published in the upstream repos. I enabled each RHEL variant's release of version 10 with a different pull request, so they can be cherry picked as needed.

### DIFF
--- a/mock-core-configs/etc/mock/alma+epel-10-aarch64.cfg
+++ b/mock-core-configs/etc/mock/alma+epel-10-aarch64.cfg
@@ -1,0 +1,5 @@
+include('almalinux-10-aarch64.cfg')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = "alma+epel-10-{{ target_arch }}"
+config_opts['description'] = 'AlmaLinux 10 + EPEL'
--- a/mock-core-configs/etc/mock/alma+epel-10-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/alma+epel-10-ppc64le.cfg
@@ -1,0 +1,5 @@
+include('almalinux-10-ppc64le.cfg')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = "alma+epel-10-{{ target_arch }}"
+config_opts['description'] = 'AlmaLinux 10 + EPEL'
--- a/mock-core-configs/etc/mock/alma+epel-10-s390x.cfg
+++ b/mock-core-configs/etc/mock/alma+epel-10-s390x.cfg
@@ -1,0 +1,5 @@
+include('almalinux-10-s3100x.cfg')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = "alma+epel-10-{{ target_arch }}"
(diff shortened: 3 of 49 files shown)


Conan-Kudo:
None of these exist yet. We should not merge this.

praiskup:
Agreed, we should merge once these are available - the preference is to not confuse users with non-working configs.

nkadel:
The EPEL enabled CFGs do not work yet, though the centos-stream-10 configs do.

praiskup:
See also the older #1383

carlwgeorge:
It doesn't make any sense to add configs for distros that don't exist yet, and it doesn't make sense for adding the configs to all be in one pull request because they will be released at different times.  Furthermore, as written these configs reference templates that don't exist and aren't part of the pull request, like centos-stream-10.tpl and epel-10.tpl.  Finally, EPEL 10 is going to have minor versions, and needs some planning put into how we approach this in the mock configs.  In my opinion, this PR should just be closed.

nkadel:
"Each will be released differently" is why I submitted each file as a distinct commit, the merge can be cherry-picked as each release occurs.

EPEL is a separately complex issue, being put through a kaleidoscope of release practices by RHEL's attempts to discard point releases with the often unwelcome "stream" model, and EPEL's attempt to separately revert to a release based model. Staring at this, I'd suggest "don't mix the streams!!!", it's likely to generate complex dependency interactions, especially with tools like python modules.

I didn't like the switch to "streaming" releases, I had some colorful metaphors for just what "stream" we were being asked to drink from. But i don't see splitting the RHEL streams to provide numbered EPEL release management in mock as serving anyone. I've previously published notes on how to do so, mainly by working from the distribution media. I don't see how mock and EPEL can really wedge into doing that unless EPEL maintains its own internal point-release based repositories for RHEL, which flies in the face of Red Hat's published policies.

If *that* is needed...... I have old tools available for just such purposes at https://github.com/nkadel/nkadel-rsync-scripts .

Conan-Kudo:
I don't think this is actionable right now, so I'm closing this and we'll incrementally handle this as Enterprise Linux 10 distributions are made available and compatible with EPEL 10.

carlwgeorge:
> EPEL is a separately complex issue, being put through a kaleidoscope of release practices by RHEL's attempts to discard point releases with the often unwelcome "stream" model, and EPEL's attempt to separately revert to a release based model.

This is not the place to re-litigate or complain about previous decisions by RHEL, the CentOS Board, or the EPEL Steering Committee.

> But i don't see splitting the RHEL streams to provide numbered EPEL release management in mock as serving anyone.

RHEL has minor versions, CentOS Stream reflects the upcoming minor version, and EPEL 10 is going to integrate with these minor versions.  mock-core-configs will reflect these facts.

nkadel:
Conan, I think you're right that this is not actionable right now. I definitely wanted to get the pull request published and reviewable for examples that can be cherry-picked as the individual operating systems are enabled

PR #1388, opened 2024-06-15
configs: allow c10s extras to be non-existing in s390x

praiskup:


Conan-Kudo:
I'm surprised the repository doesn't exist on s390x...

praiskup:
I reported: https://issues.redhat.com/browse/RHEL-41263

praiskup:
@asamalik can you please take a quick look?

praiskup:
No response :shrug: is anyone against merging this?

Conan-Kudo:
I would prefer we did not.

asamalik:
I moved it to the CS Jira project, that's where it would get noticed!

Although you've done the logical thing with selecting CentOS Stream 10 and "releng" as a component. I have no idea why it exists, but I'll add it to our CS team inbox so we see those next time.

I'll ask Fabian to have a look. Looks like just an empty repo needs to be created on the mirror (as there's no s390x in CBS) so it's not breaking stuff by not existing.

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
```diff
@@ -72,7 +72,9 @@ metalink=https://mirrors.centos.org/metalink?repo=centos-extras-sig-extras-commo
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official-SHA256
 gpgcheck=1
 enabled=1
-skip_if_unavailable=False
+{% if repo_arch == 's390x' %}
+skip_if_unavailable=True
+{% endif %}
```
xsuchy, line 77: Should there be else-branch with original line?

praiskup:
@asamalik ping

praiskup:
I eventually submited Fedora Infra issue https://pagure.io/fedora-infrastructure/issue/12169 according to Fabian's wish in Jira.

praiskup:
This PR is not needed now; the mirror manager link has been fixed (see the fedora infra ticket).

PR #1393, opened 2024-06-26
Support for isolated builds

praiskup:
- [x] Don't misuse the package_state plugin for this purpose
- [x] ~~Standardize the output JSON format~~ format fixed for now, this will take time
- [x] add a feature documentation page
- [x] manual pages
- [x] add unit-tests && behave test
- [x] assert the bootstrap image sha1


praiskup:
One of the experiments might look like:

```
1. prepare `f41-bootstrap` image that has dnf5-plugins available
2. download RPMs (cache): mock --calculate-build-dependencies -r fedora-rawhide-x86_64 /tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm
3. copy cached RPMs aside: cp -r /var/cache/mock/fedora-rawhide-x86_64/dnf_cache/fedora-2d95c80a1fa0a67d/packages/ /tmp/rawhide
4. create repo: createrepo_c /tmp/rawhide
5. isolated build: mock -r fedora-isolated /tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm --config-opts local_directory=/tmp/rawhide --config-opts=bootstrap_image=localhost/f41-bootstrap:latest --config-opts=bootstrap_image_skip_pull=True
```

Alternative way to 5 might look like:

```
5. install build deps: mock -r fedora-isolated --installdeps /var/lib/mock/fedora-rawhide-x86_64/result/python-copr-1.132-1.git.0.787a685.fc41.buildreqs.nosrc.rpm  --config-opts local_directory=/tmp/rawhide --config-opts=bootstrap_image=localhost/f41-bootstrap:latest --config-opts=bootstrap_image_skip_pull=True
6. run the final build (no clean): mock -r fedora-isolated --no-clean /tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm --config-opts local_directory=/tmp/rawhide --config-opts=bootstrap_image=localhost/f41-bootstrap:latest --config-opts=bootstrap_image_skip_pull=True

File: mock/py/mockbuild/backend.py [hunk 1]
```diff
@@ -720,6 +721,7 @@ def rebuild_package(self, spec_path, timeout, check, dynamic_buildrequires):
         # --nodeps because rpm in the root may not be able to read rpmdb
         # created by rpm that created it (outside of chroot)
         check_opt = []
+        depsonly = self.config["calculatedeps"]
```
xsuchy, line 724: depsonly vs. calculatedeps confuses me when reviewing this PR. I would use the same name. I.e. s/depsonly/calculatedeps/.

xsuchy (commented):
This "skip pull" commit is weird. I do not understand it.

praiskup:
> This "skip pull" commit is weird. I do not understand it.

The purpose is to not "podman pull" the image (and fail) if the image is not pushed into any registry.  I'm going to document it (or drop the commit), sorry for the confusion.

praiskup:
Updated with a bit more convenient use-case, see below.

First we calculate the build-deps:

    $ mock -r fedora-rawhide-x86_64 --calculate-build-dependencies /tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm

Then, using the JSON file in result directory, we download the RPMs and prepare
the local repository:

    $ repo-from-json --json /var/lib/mock/fedora-rawhide-x86_64/result/mock-build-environment.json --output-repo /tmp/repo

Then, with the local-repo prepared, we can run the isolated build:

    $ mock -r fedora-isolated --isolated-build-config /tmp/prepared/mock-build-environment.json /tmp/repo/ /tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm

praiskup:
The last part is now simplified, one just needs to

    $ mock --isolated-build-config /tmp/prepared/mock-build-environment.json /tmp/repo/ /tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm

Warning: this whole PoC only works with bootstrap_image_ready = True.

praiskup:
The experimental use-case looks like this now (after the recent update):

    # we use the srpm twice
    srpm=/tmp/tito/python-copr-1.132-1.git.0.787a685.fc40.src.rpm

    # record list of RPMs and bootstrap image we need to download
    mock --calculate-build-dependencies -r fedora-rawhide-x86_64 "$srpm"

    # backup the json file
    cp -r /var/lib/mock/fedora-rawhide-x86_64/result/mock-build-environment.json /tmp/

    # pre-fetch the files
    mock-isolated-repo --json /tmp/mock-build-environment.json --output-repo /tmp/repo

    # run the isolated build
    mock --isolated-build-config /tmp/mock-build-environment.json /tmp/repo/ "$srpm"

mock/py/mock-isolated-repo.py
```diff
@@ -0,0 +1,82 @@
...
+
+import requests
+
+
+def download_file(url, outputdir):
+    """
+    Download a single file (pool worker)
+    """
+    file_name = os.path.join(outputdir, os.path.basename(url))
+    print(f"Downloading {url}")
```
xsuchy, line 24: Should this be `log.info()`?

xsuchy: End everywhere else in this file?

mock/py/mockbuild/plugins/package_state.py
```diff
@@ -62,6 +67,55 @@ def _availablePostYumHook(self):
...
+    def _installedPreBuildHook2(self):
+        filename = "mock-build-environment.json"
+        statename = "Outputting the file " + filename
+        try:
+            with self.buildroot.uid_manager:
+                self.state.start(statename)
+                out_file = os.path.join(self.buildroot.resultdir, filename)
+                chrootpath = self.buildroot.make_chroot_path()
+
+                package_list_cmd  = "rpm --root {0} -qa".format(shlex.quote(chrootpath))
```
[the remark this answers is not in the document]
tkopecek: Adding licenses - they're needed for SBOM generation.

mock/py/mock.py
```diff
@@ -391,6 +394,8 @@ def command_parse():
                       help=("Additional package to install into the buildroot before "
                             "the build is done.  Can be specified multiple times."))
 
+    parser.add_option("--isolated-build-config", nargs=2, metavar=("json", "localrepo"))
```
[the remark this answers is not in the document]
praiskup: done

mock/py/mockbuild/plugins/package_state.py
```diff
@@ -62,6 +67,55 @@ def _availablePostYumHook(self):
...
+                out, _ = self.buildroot.doOutChroot(package_list_cmd, returnOutput=True, shell=True)
+                packages = out.splitlines()
+
+                cmd = "dnf -q --installroot={0} repoquery --location".format(shlex.quote(chrootpath))
+                cmd += " " + " ".join(shlex.quote(p) for p in packages)
+                out, _ = self.buildroot.doOutChroot(cmd, returnOutput=True, shell=True)
```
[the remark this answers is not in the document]
praiskup: done

mock/py/mock-isolated-repo.py
```diff
@@ -0,0 +1,82 @@
...
+    except:
+        print(f"traceback for {url}")
+        raise
+
+
+def _argparser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--json", required=True)
+    parser.add_argument("--output-repo", required=True)
+    return parser
```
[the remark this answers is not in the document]
praiskup: done

mock/py/mockbuild/plugins/package_state.py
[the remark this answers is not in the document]
praiskup: It's (a bit) tricky because we need to map the output of `rpm --qf` with the output from `repoquery` (which reorders the stuff according to its own rules, hopefully it is just sorted using NVRA).

The `repoquery --location` can't be combined with `--qf` :-) and there's no `--qf %location` in DNF4's repoquery.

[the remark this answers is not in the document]
praiskup: Ok, done, but the `signature` field seems undocumented in the Koji Content Generator, and I'm not sure how to get it (therefore there's a hack parsing the %{sigpgp:pgpsig} output for now).  Anyway, we don't use signed packages in Koji/Brew, or?  The current package metadata looks like:
```
            {
                "license": "LicenseRef-Fedora-Public-Domain",
                "name": "filesystem",
                "version": "3.18",
                "release": "23.fc41",
                "arch": "x86_64",
                "epoch": null,
                "nvra": "filesystem-3.18-23.fc41.x86_64",
                "sigmd5": "dc6edb2b7e390e5f0994267d22b9dc1a",
                "signature": null,
                "url": "http://ftp.sh.cvut.cz/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/f/filesystem-3.18-23.fc41.x86_64.rpm"
            },

```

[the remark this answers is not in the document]
tkopecek: We don't use them for building - we use stripped versions of them. But rpms can be imported and/or signed and this info is stored in koji. Also CG-imported content could have been created from such. In case we wouldn't have checksums of the signed content we wouldn't be able to match them for 100%.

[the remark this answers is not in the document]
praiskup: OK, got an offline confirmation that just the last 8 characters from the `%{sigpgp:pgpsig}` are needed (available/provided e.g. for `rhel-9-x86_64` RPMs from RH's CDN).  PR is updated (thread marked as resolved).

This `rpm -q --qf` + `dnf repoquery` dance is pretty clumsy - @tkopecek proposed to have a dedicated script working with the python3-rpm library - which sounds reasonable, but not trivial in Mock's abstraction layers.  Moving the idea to #1433.

docs/Plugin-BuildrootLock.md
```diff
@@ -0,0 +1,60 @@
...
+
+This plugin will not work correctly with `--offline` mode.
+
+## Format of the lockfile
+
+We currently do not provide a compatibility promise.  Only the very same version
+of Mock that produced the file is guaranteed to read and process it later (see
+the feature page, TODO).
+
+Example of such lockfile:
```
xsuchy, line 23: It would be nice to provide JSON schema too. See my similar work in https://gitlab.com/fedora/legal/fedora-license-data/-/blob/main/tools/fedora-license-schema.json?ref_type=heads and validation https://gitlab.com/fedora/legal/fedora-license-data/-/blob/main/Makefile?ref_type=heads#L64

docs/Plugin-BuildrootLock.md
```diff
@@ -0,0 +1,60 @@
...
+                    "epoch": null,
+                    "nvra": "fedora-repos-rawhide-42-0.1.noarch",
+                    "sigmd5": "4378929dac9e51ed8470de5173ae664e",
+                    "signature": null,
+                    "url": "http://ftp.sh.cvut.cz/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/f/fedora-repos-rawhide-42-0.1.noarch.rpm"
+                },
+                ...
+            ]
+        },
+        "config": {
```
xsuchy, line 44: Do we really want this as array? I would probably put it here as multiline string.

mock/docs/mock.1
```diff
@@ -137,6 +149,13 @@ Do a yum install PACKAGE inside the chroot. No 'clean' is performed.
 \fB\-\-installdeps\fP
 Find out deps for SRPM or RPM, and do a yum install to put them in the chroot. No 'clean' is performed
 .TP
+\fB\-\-isolated\-build\-config \fImock\-build\-environment.json\fR \fIREPO\fR \fISRPM\fR
+Perform an isolated RPM build from SRPM (= an offline one, without the need to
+access Internet).  After running the \fB\-\-calculate\-build\-dependencies\fR
+first to get the \fImock\-build\-environment.json\fR file, and then running the
+\fImock\-isolated\-repo(1)\fR helper for generating \fIREPO\fR, Mock has all the
+necessary info for building RPMs from the \fISRPM\fR fully offline.
+.TP
```
xsuchy, line 158: I would explicitely state that the file is filename relative to PWD. To not repeat history with -r option.

mock/mock.spec
```diff
@@ -188,6 +194,7 @@ cp -a etc/pam/* %{buildroot}%{_sysconfdir}/pam.d/
 
 install -d %{buildroot}%{_sysconfdir}/mock
 cp -a etc/mock/* %{buildroot}%{_sysconfdir}/mock/
+cp -a etc/isolated-build.cfg %{buildroot}%{_sysconfdir}/mock/
```
xsuchy, line 197: Can you put `isolated-build.cfg` to 'etc/mock' ? It will save several special handling like this one.

mock/py/mock-isolated-repo.py
```diff
@@ -0,0 +1,105 @@
...
+    Download a single file (pool worker)
+    """
+    file_name = os.path.join(outputdir, os.path.basename(url))
+    log.info("Downloading %s", url)
+    try:
+        with requests.get(url, stream=True, timeout=30) as response:
+            if response.status_code != 200:
+                return False
+            with open(file_name, "wb") as fd:
+                shutil.copyfileobj(response.raw, fd)
```
xsuchy, line 36: Reading https://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content I wonder: should you use iter_content?

mock/py/mockbuild/config.py
```diff
@@ -398,6 +404,53 @@ def multiply_platform_multiplier(config_opts):
...
+        data = json.load(fd)
+
+    if not data["config"].get("bootstrap_image_ready"):
+        raise exception.BadCmdline(
+            f"The file {json_conf} did not record the bootstrap_image_ready=True "
+            "config which means we are not able to prepare the bootstrap chroot "
+            "in an isolated mode.")
+
+    for opt, val in data["config"].items():
+        config_opts[opt] = val
```
xsuchy, line 430: I would rather utilize https://github.com/rpm-software-management/mock/blob/main/mock/py/mockbuild/config.py#L679 It will need some alternation, but you get all these includes for free.

mock/docs/mock.1
praiskup: It may be any file, not necessary relative to CWD

docs/Plugin-BuildrootLock.md
praiskup: s/array/dict/?  Do you suggest having "dict" encoded as string inside the upper level dict?

xsuchy: Ignore this. I just learn that multiline values are not supported in JSON.

docs/Plugin-BuildrootLock.md
praiskup: Going to provide the schema-as-doc, and add a unit-test "testing" the schema.

mock/py/mock-isolated-repo.py
praiskup: It is possible to [work with `response.raw`](https://requests.readthedocs.io/en/latest/api/#requests.Response.raw), and then you can read the file using chunks with [copyfileobj]( https://docs.python.org/3/library/shutil.html#shutil.copyfileobj).

mock/py/mockbuild/config.py
praiskup: The update_config_from_file() works with filenames though, and does a lot of magic with includes... while here we don't want to interpret includes, and we don't even have a "Python file / Mock config file" in hand; we have just a poor dictionary object in hand that we want to merge with config_opts...   I created a new utility method `update_config_from_dict` which logically complements `update_config_from_file` so this fact is clear.... though.... I may be missing something.

The rest of this method should otherwise stay as-is IMO, this seems to be the correct place to set "isolated build" defaults (e.g. set the `bootstrap_image_skip_pull = True`).

mock/docs/mock.1
praiskup: I added notes like `Evaluate  all  the SRPM (= file name, path on your system)`.

docs/Plugin-BuildrootLock.md
praiskup: Still TBD

mock/py/mockbuild/plugins/buildroot_lock.py
```diff
@@ -0,0 +1,92 @@
...
+                    # work with a single-config-for-all-arches now.
+                    "target_arch",
+                    "legal_host_arches",
+                    "dist",
+                    "package_manager",
+                    # At this point, we only support isolated builds iff
+                    # bootstrap_image_ready=True, so these two options are
+                    # useful for implementing "assertion" in the
+                    # process_isolated_build_config() method.
+                    "bootstrap_image",
```
tkopecek, line 83: Bootstrap image would need also a sha256. Now it refers e.g. to "registry.fedoraproject.org/fedora:rawhide" which is always changing pointer.

praiskup: new field bootstrap.image_digest added

praiskup: and with the updated code, mock also asserts that the current digest equals the previous one

mock/docs/buildroot-lock-schema-1.0.0.json
```diff
@@ -0,0 +1,107 @@
...
+                "type": "string"
+              },
+              "name": {
+                "description": "Name of the package",
+                "type": "string"
+              },
+              "nvra": {
+                "description": "Name-Version-Release-Arch 4-ary, dash-separated.",
+                "type": "string"
+              },
```
praiskup, line 44: Going to drop `nvra` field (redundant info)

mock/docs/buildroot-lock-schema-1.0.0.json
```diff
@@ -0,0 +1,107 @@
...
+    "version": {
+      "description": "Version of the https://raw.githubusercontent.com/rpm-software-management/mock/main/mock/docs/buildroot-lock-schema.json schema the document conforms to. Semantic versioned. Mock that implements v2.Y.Z versions no longer reads v1.Y.Z.",
+      "const": "1.0.0"
+    },
+    "buildroot": {
+      "description": "The object that describes the Mock buildroot",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "packages": {
```
praiskup, line 18: Going to rename "packages" into "rpms" to make it prepared for the future.  Even though we don't plan to support them with lockfile right now, mock is able to install other packages too (pypi, gems, ..).

praiskup:
We can not do isolated builds with centos-stream for now, reported: https://issues.redhat.com/browse/CS-2506

mock/py/mockbuild/plugins/buildroot_lock.py
```diff
@@ -0,0 +1,111 @@
+"""
+Produce a lockfile for the prepared buildroot by Mock.  Once available, we
+should use the DNF built-in command from DNF5:
+https://github.com/rpm-software-management/dnf5/issues/833
+"""
```
praiskup, line 5: I'm explicitly linking this PR to https://github.com/rpm-software-management/dnf5/issues/833

tkopecek:
Saving whole (parsed) mock config could be valuable? E.g. what if I run the ``--isolated-build`` phase on different machine with different ``site-defaults.cfg``? Maybe

praiskup:
> About json versioning: Should we also store in the config part which mock version created it? I'm more inclined to "no" as mock probably doesn't have enough information about itself (was it e.g. modified in the distribution?) and leave it on other tooling. 

I agree here.  When it comes to full reproducibility, it's by far not just about Mock version.  We probably want to reproduce on the same kernel, and run Mock-in-container, from the very same container image, with the very same version of container tooling, etc.  The "feature-isolated-build" feature page discusses this a bit (the buildroot environment maintained by mock != the environment we run mock in).

> OTOH, it can be easilly lost which mock should be used for the rebuild

The NVR of Mock in use is dumped into logs, and if the situation comes - we should be able to find the right version by git-bisect.  Though, ... if you prefer to dump the mock version there, I will not object too much!  And I really expect a heavy development here, so we may add it later - this PR is just the starting point.

> Saving whole (parsed) mock config could be valuable? E.g. what if I run the --isolated-build phase on different machine with different site-defaults.cfg? Maybe it is also up to "process above" to store/provide correct mock configs?

Yes, my vote goes for "process above".  If we have a mock container image ID/sha256 [like this one](https://github.com/praiskup/mock-container-image), we know both the Mock version and the set of configuration files.

If all of this implicates we should document this more, please say what and where! :+1:

tkopecek:
> If all of this implicates we should document this more, please say what and where! 👍

I think we should just stress it a bit more in "Limitations" section. For some it could look like it is providing reproducible builds without taking in account state of host machine software/hardware stack. Just a one sentence that there is still a big influence of external factors.

praiskup:
I borrowed your wording and updated the Limitation section.  Thank you!

xsuchy (approved):

PR #1397, opened 2024-07-02
configs: CentOS 7 is EOL

praiskup:


praiskup:
I'd experiment with https://vault.centos.org/7.9.2009/ but it seems this is x86_64-only, am I missing something?

xsuchy (approved):
+1

Conan-Kudo (changes requested):
EPEL 7 is EOL too, so that needs to be handled.

Conan-Kudo:
It would make sense to move all EL7 stuff to EOL now.

nvkhoi112358:
I think Oracle Linux 7 (+ its EPEL) and RHEL 7 should be handled, too.

praiskup:
Do we need to make it in one step?

praiskup:
https://endoflife.date/oracle-linux seems to have a few more months 

I EOLed everything which is related to EPEL 7, though.  PTAL

Conan-Kudo (approved):

PR #1410, opened 2024-07-18
Mock/buildroot: respect nspawn_args when bootstrapping chroot

thozza:
The 'nspawn_args' configuration value was not used when executing the systemd-nspawn command during the chroot bootstrapping. Make sure that the value is used during the chroot bootstrapping in case systemd-nspawn isolation is used.

This issue was found when trying to workaround an issue [1] when SELinux policy forbid systemd-machine to create a varlink socket and thus start. This resulted in systemd-nspawn not being able to register a machine. To workaround this, I added the following snippet to the configuration:

```
config_opts['nspawn_args'] = ['--register=no']
```

so that systemd-nspawn does not try to register the machine with systemd-machine. However, this had no effect. I could see that the argument was not used when executing systemd-nspawn and it still failed.

[1] https://issues.redhat.com/browse/RHEL-49567

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -137,6 +137,11 @@ class Buildroot(object):
         self._setup_nspawn_fuse_device()
         self._setup_nspawn_loop_devices()
 
+    def _get_nspawn_args(self):
+        nspawn_args = []
+        if util.USE_NSPAWN:
+            nspawn_args.extend(self.config['nspawn_args'])
+        return nspawn_args
 
     def set_package_manager(self, fallback=None):
         """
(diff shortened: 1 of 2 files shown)


praiskup:
Notes for myself:
- There's yet another instance of the very same problem in the "bootstrap_image" and "bootstrap_image_ready = False" case
- The release notes comment needs fix, this is actually fixing the `rpm -qa` (chroot package listing)
- there's likely yet another similar problem in Mock that actually [failed here](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild/-/jobs/7374793681#L2068)

praiskup:
I've tried to fix this in #1456, as it is a bit more complicated, but I kept you as co-author.

Issue #1421, opened 2024-08-12
EPEL 10 chroots


hroncok:
Please add CentOS Stream 10 + EPEL 10 chroot configs. There is no compose yet, just the Koji buildroot.

Thanks.

cc @carlwgeorge

hroncok:
https://kojipkgs.fedoraproject.org/repos/epel10.0-build/latest/

carlwgeorge:
I was planning on doing this after we had it published and mirrored, to avoid hitting the koji tag repo to hard.  If that's not a concern I can send a PR with the configs, and then another later to switch to the mirror metalinks.

hroncok:
I already hit the Koji repo manually, making it available in the official mock-core-configs might expose a bit more, but I don't think it will destroy Koji. There's `--enable-repo=local` and there's Fedora i686 config using Koji only already.

carlwgeorge:
I'll be happy to review #1424.

For epel-release we're going to use the `$releasever_major${releasever_minor:+.$releasever_minor}` pattern that was already suggested in that PR.  I think the best approach will be to have two epel10 templates, one without the minor version and one with it, sort of like the separate rawhide/branched templates for Fedora.  CentOS Stream would use epel-10.tpl without the minor version and wouldn't need to be aware of the minor version at all, since the published repo will be `pub/epel/10`.  RHEL would use epel-10-branched.tpl (I'm open to better file names) with the minor version.  This would work best with the RHEL template also integrating the minor versions.  We could use `config_opts['releasever_major'] = '10'` in rhel-10.tpl, and then have files like rhel+epel-10.0-x86_64 that set `config_opts['releasever_minor'] = '0'`.

As an aside, this [visual aid from the EPEL 10 hackfest at Flock](https://carlwgeorge.fedorapeople.org/presentations/epel-10-hackfest-2024-flock.pdf) may be useful for you in understanding which parts of the pipeline will have the minor version and which ones won't.

praiskup:
The `centos-stream+epel-10` chroots for now use `10.0` version, per 3802ca3 (therefore this one was closed).  But there's #1427 where we may use what you suggest.  I'm going to review PRs if you already know which way you prefer.

> then have files like rhel+epel-10.0-x86_64

[...]

Issue #1295, opened 2024-01-23
RFE: Add a shortcut to use a Koji side tag


hroncok:
Mock already has an option to enable the Koji repository with `--enablerepo=local`.
It would be really convenient to have a similar option to enable a Koji side-tag repo.

Currently, this is possible with something like:

```
mock -r centos-stream-9-x86_64 --addrepo='https://kojihub.stream.centos.org/kojifiles/repos/c9s-build-side-2685-stack-gate/latest/x86_64/'
mock -r fedora-rawhide-x86_64 --addrepo='https://kojipkgs.fedoraproject.org/repos/f40-build-side-81735/latest/x86_64/'
```

However, that's rather long and hard to remember.

I would appreciate it if I could do something like `--side-tag=f40-build-side-81735`.

I see several options to achieve this.

### Add side_tag variable to the local repo baseurl

E.g. replace this:

```
[local]
name=local
baseurl=https://kojipkgs.fedoraproject.org/repos/rawhide/latest/$basearch/
...
enabled={{ not mirrored }}
...
```

With:

```
[local]
name=local
baseurl=https://kojipkgs.fedoraproject.org/repos/{{ side_tag or "rawhide" }}/latest/$basearch/
...
enabled={{ side_tag != None or not mirrored }}
...
```

The `side_tag` would obviously be `None` by default and as an user I could override it by `--side-tag=...`

Note: This could be challenging with the combined centos-stream+epel repos, where there are 2 local repositories, one for each Koji, but I guess not unsolvable.

### Add common --addrepo aliases

This option is more general but a tad more flexible.

In short, assume I could do: `--addrepo=koji://f40-build-side-81735` or `--addrepo=copr://@python/python3.13`.

Such aliases would need to be well-defined and doing it this way requires more thought. Copr itself already has the concept for `copr://` when adding external repositories, so this could be reused in Copr and more users could benefit from this.

praiskup:
Wouldn't a simple script `koji-side-tag-mock-config` generate the mock-config (with include the default) help here?

praiskup:
Otherwise, good idea with the `--addrepo`.  If anyone wanted to submit patch, here goes the code that needs to be updated:

https://github.com/rpm-software-management/mock/blob/4b3fb22e2f74a17bf7879395b6f6418192430163/mock/py/mock.py#L773-L775

The `copr://` syntax is already implemented in copr-frontend:
[...]

PR #1424, opened 2024-08-14
Add CentOS Stream 10 + EPEL 10 configs

praiskup:
These work with Koji EPEL buildroot only for now since there are no composes, yet.

Fixes: #1421

### DIFF
--- a/mock-core-configs/etc/mock/centos-stream+epel-10-s390x.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-10-s390x.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel"
+include('templates/centos-stream-10.tpl')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = 'centos-stream+epel-10-s3100x'
+config_opts['description'] = 'CentOS Stream 10 + EPEL'
+config_opts['target_arch'] = 's3100x'
+config_opts['legal_host_arches'] = ('s3100x',)
--- a/mock-core-configs/etc/mock/templates/epel-10.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-10.tpl
@@ -1,0 +1,11 @@
+config_opts['chroot_setup_cmd'] += " epel-rpm-macros"
+
+config_opts['dnf.conf'] += """
+
+[local]
+name=Extra Packages for Enterprise Linux $releasever - Koji Local - BUILDROOT ONLY!
+baseurl=https://kojipkgs.fedoraproject.org/repos/epel10.0-build/latest/$basearch/
+cost=2000
+enabled=1
+skip_if_unavailable=False
+"""
--- a/mock-core-configs/etc/mock/centos-stream+epel-10-aarch64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-10-aarch64.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel"
+include('templates/centos-stream-10.tpl')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = 'centos-stream+epel-10-aarch64'
+config_opts['description'] = 'CentOS Stream 10 + EPEL'
+config_opts['target_arch'] = 'aarch64'
+config_opts['legal_host_arches'] = ('aarch64',)
--- a/mock-core-configs/etc/mock/centos-stream+epel-10-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-10-ppc64le.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel"
+include('templates/centos-stream-10.tpl')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = 'centos-stream+epel-10-ppc64le'
+config_opts['description'] = 'CentOS Stream 10 + EPEL'
+config_opts['target_arch'] = 'ppc64le'
+config_opts['legal_host_arches'] = ('ppc64le',)
--- a/mock-core-configs/etc/mock/centos-stream+epel-10-x86_64.cfg
+++ b/mock-core-configs/etc/mock/centos-stream+epel-10-x86_64.cfg
@@ -1,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel"
+include('templates/centos-stream-10.tpl')
+include('templates/epel-10.tpl')
+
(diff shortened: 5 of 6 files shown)


mock-core-configs/etc/mock/templates/epel-10.tpl
```diff
@@ -0,0 +1,11 @@
+config_opts['chroot_setup_cmd'] += " epel-rpm-macros"
+
+config_opts['dnf.conf'] += """
+
+[local]
+name=Extra Packages for Enterprise Linux $releasever - Koji Local - BUILDROOT ONLY!
+baseurl=https://kojipkgs.fedoraproject.org/repos/epel10.0-build/latest/$basearch/
```
praiskup, line 7: Note the `10.0` part of the link, while baseurl=10 :-/

hroncok: Can we use `$releasever_major${releasever_minor:+.$releasever_minor}`?

Conan-Kudo: Is there a way we could retrieve the latest koji epel 10 buildroot to populate that value? Because otherwise we'll have to update `10.y` into new numbers every few months.

praiskup: I'm trying to tl;dr the original proposal https://discussion.fedoraproject.org/t/epel-10-proposal/44304/25 and it seems that at least the mirror/metalinks would (not ATM, since there are no composes yet) have a "floating" target eventually (the default).

Would we consider having something like `epel10-latest -> epel10.0` symlink in Koji?

Conan-Kudo: Probably worth asking @nirik about it.

carlwgeorge: We need to do something like this to get `fedpkg verrel` working.  My theory at this point is to ask koji for the epel10 target, which points you to the epel10.0-build tag (and will move forward through time), and then get the distcore macro from that.  We could set an additional macro with the version in a different format to support mock and whatever else needs it.

praiskup: I don't think we want to ask `fedpkg verrel` for every single `mock` run (query API).  It seems even easier/wiser/cost-effective to html-parse the koji repository URL (if dirlisting is available) or check for HTTP 200 vs 404. But still quite an inconvenient change in EL10 honestly.  The koji-side symlink would be easier.

praiskup:
Actually, is anyone against merging this as-is?  It's not an optimal long-term solution, but it should work well for now, even in Fedora Copr.

mock-core-configs/etc/mock/templates/epel-10.tpl
carlwgeorge: Sorry I wasn't clear, I wasn't suggesting running `fedpkg verrel` from mock, just that for that command we'll need to solve a similar problem and the source of truth will be the koji epel10 build target.  I'm not quite sure what you mean by a koji symlink.

praiskup: Both  `verrel` and Mock will need to get the info from Koji API. This is an equivalent problem to [fedora-distro-aliases](https://github.com/rpm-software-management/fedora-distro-aliases), and the #1328, or maybe #1295.

>  I'm not quite sure what you mean by a koji symlink

It would be nice if those were symlinks server-side:

```
https://kojipkgs.fedoraproject.org/repos/epel10.0-build/
https://kojipkgs.fedoraproject.org/repos/epel10-latest-build/
```

... and if the `epel10-latest-build` symlink was updated anytime we start building for a new minor.

praiskup:
Since nobody votes against merging this, and neither TF complaints:
```
PASSED: centos-stream+epel-10-x86_64.cfg (tmpfs)
PASSED: centos-stream+epel-10-x86_64.cfg
```

... I will merge it so we have it in the [next release](https://github.com/rpm-software-management/mock/pull/1425).

mock-core-configs/etc/mock/centos-stream+epel-10-s390x.cfg
```diff
@@ -0,0 +1,8 @@
+config_opts["koji_primary_repo"] = "epel"
+include('templates/centos-stream-10.tpl')
+include('templates/epel-10.tpl')
+
+config_opts['root'] = 'centos-stream+epel-10-s3100x'
+config_opts['description'] = 'CentOS Stream 10 + EPEL'
+config_opts['target_arch'] = 's3100x'
+config_opts['legal_host_arches'] = ('s3100x',)
```
hroncok, line 8: s/s3100x/s390x/

mock-core-configs/etc/mock/templates/epel-10.tpl
hroncok: This is how https://kojipkgs.fedoraproject.org/repos/rawhide/latest/ works. The relengs update it from f41-build to 42-build on branching. See e.g. https://pagure.io/releng/issue/12255

So indeed, we would probably want epel10-c10s and epel10-rhel10 links to latest and second to latest eventually

nirik:
So, we could do links, although I would prefer to avoid it as much as possible as it means more 'manual' steps.

Also, just to note, as soon as we have real epel10 repos we should default to pointing to those, not the buildroot. IMHO, the buildroot should be like the 'local' repo for fedora, disabled by default, but possible for developers who need it for some specific region. I always worry about more and more people hitting kojipkgs for these things instead of using the mirror network. ;)

mock-core-configs/etc/mock/centos-stream+epel-10-s390x.cfg
praiskup: Fixed in 911ea87e250d4db37bd3ac009db0f1e043eece21

mock-core-configs/etc/mock/templates/epel-10.tpl
praiskup: second to latest?

praiskup:
> So, we could do links, although I would prefer to avoid it as much as possible as it means more 'manual' steps.

That's the truth and the unfortunate cons of the change, but I'm afraid that if we don't do this manual step once in the source, every consumer (like mock-core-configs) needs to do the manual steps instead. Which seems even worse.

> IMHO, the buildroot should be like the 'local' repo for fedora, disabled by default

Absolutely!  As I understood it, there will be such a floating target in composes, too (no need to update mock-core-configs for minor RHEL bumps every 6 months).

Issue #1412, opened 2024-07-22
New Mock & configs release


praiskup:

xsuchy:
Related https://github.com/rpm-software-management/mock/issues/1418 
If we do that, I propose to bump the major version.

praiskup:
#1417 is related too.

PR #1425, opened 2024-08-14
Release mock-core-configs v41.1

praiskup:
/usr/bin/tito tag --use-version=41.1
./releng/generate-release-notes --config-only --use-version 41.1


praiskup:
Relates to #1412

PR #1428, opened 2024-08-15
Don’t ship rpmautospec plugin with main package

nphilipp:


mock/mock.spec
```diff
@@ -256,6 +256,8 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 %exclude %{python_sitelib}/mockbuild/__pycache__/scm.*
 %exclude %{python_sitelib}/mockbuild/plugins/lvm_root.*
 %exclude %{python_sitelib}/mockbuild/plugins/__pycache__/lvm_root.*
+%exclude %{python_sitelib}/mockbuild/plugins/rpmautospec.*
+%exclude %{python3_sitelib}/mockbuild/plugins/__pycache__/rpmautospec.*.py*
```
praiskup, line 260: Are you intentionally using python3_sitelib in the second line?

praiskup: nevermind, I see it used like this on multiple places, so probably innocent c&p

Issue #1429, opened 2024-08-21
package_state output doesn't count with %generate_buildrequires


praiskup:
The package state output, `installed_pkgs.log`, is supposed to provide a list of installed needed packages for the package to build.  But we provide a list of packages after installing static build requires, before calculating the dynamic dependencies (so we do not include them).  CC @tkopecek

PR #1431, opened 2024-08-22
The installed_pkgs.log file now covers dynamic builddeps

praiskup:
This change required us to create a new "postdeps" hook with post-%generate_buildrequires timing.

Fixes: #1429
Relates: #1393


mock/py/mockbuild/plugins/package_state.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, plugins, conf, buildroot):
         self.inst_done = False
         self.online = self.buildroot.config['online']
         plugins.add_hook("postyum", self._availablePostYumHook)
-        plugins.add_hook("prebuild", self._installedPreBuildHook)
```
xsuchy, line 46: Although this hook is not documented, it is used by several plugins. I believe it should not be removed.

praiskup: The hooks in this "prebuild" namespace are still called, this particular hook is just moved to a later timing...

xsuchy: Right. I was confused and thought you are replacing the hook itself. Sorry for the noise.

xsuchy:
+1

PR #1435, opened 2024-08-29
towncrier: update to v45 && use defaults

praiskup:
The since_last_remote_commit doesn't seem to work correctly;  for some reason, it sometimes gives us a limited set of changed files (appears like it was generated since the last commit in the very same PR, not from the target branch), example:
https://github.com/rpm-software-management/mock/actions/runs/10616932275/job/29431772686?pr=1393


xsuchy (approved):
+1

PR #1444, opened 2024-09-04
CI: lint the whole project

praiskup:
This requires a top-level pylintrc file.  The previous per-subdir
solution was broken (each subdir would need to provide a specfile, or
the .vcs-diff-lint.yml file to work correctly).


praiskup:
Releng/CI only commit, merging early to not block other PRs.

PR #1446, opened 2024-09-04
ci: drop our custom Towncrier task

praiskup:
The Chronographer does the same thing more generic way.  The only missing part was to configure the hint that should point our users to the Release-Notes-Next documentation page, and it is done now.


praiskup:
These config changes always take effect after the PR is merged.

PR #1447, opened 2024-09-04
ci: allow us to skip Packit builds with "no-ci-builds" label

praiskup:


praiskup:
This is to reduce the cost of 'push --force'.  CI/releng only, merging early.

Issue #1442, opened 2024-09-03
Make the CentOS Stream bootstrap image "ready"


praiskup:
We need to make python3-dnf-plugins-core part of the CS images:
https://issues.redhat.com/browse/CS-2506

Then we need to set boostrap_image_ready=True in the configuration.

The behave test-case from #1393 should start using native images.

praiskup:
#1451 resolves C10S, #1458 resolves C9S, and C8S is EOL.

PR #1451, opened 2024-09-14
Enable bootstrap container for CentOS Stream 10

carlwgeorge:
Since CentOS Stream 10 hasn't been released yet, the container image exists but is using the tag stream10-development instead of stream10. At some point (probably the release announcement) it will start using the stream10 tag.  This image is "ready" (i.e. `dnf builddep` is pre-installed).

cc: @tdawson

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
@@ -5,8 +5,8 @@ config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['description'] = 'CentOS Stream 10'
 
-config_opts['use_bootstrap_image'] = False
-config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream10'
+config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream10-development'
+config_opts['bootstrap_image_ready'] = True
 
 config_opts['dnf.conf'] = """
 [main]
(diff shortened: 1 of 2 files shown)


praiskup:
Nice! Thank you. Can you mark #1442 Fixed in the commit message?

carlwgeorge:
Of course, amended the commit.

praiskup:
Two more things preferably in the same commit...  would you please remove the temporary hack from behave config:
```
diff --git a/behave/features/isolated-build.feature b/behave/features/isolated-build.feature
index ffd623f5..ae22eb9a 100644
--- a/behave/features/isolated-build.feature
+++ b/behave/features/isolated-build.feature
@@ -12,9 +12,6 @@ Feature: Mock 5.7+ supports isolated builds
     @isolated_build
     Scenario: Isolated build against a DNF4 distribution
         Given an unique mock namespace
-        # Temporary image, until we resolve https://issues.redhat.com/browse/CS-2506
-        And next mock call uses --config-opts=bootstrap_image=quay.io/mock/behave-testing-c9s-bootstrap option
-        And next mock call uses --config-opts=bootstrap_image_ready=True option
         When deps for mock-test-bump-version-1-0.src.rpm are calculated against centos-stream+epel-9-x86_64
         And a local repository is created from lockfile
         And an isolated build is retriggered with the lockfile and repository
```
And provide a release note snippet? (towncrier create c10s-bootstrap-ready.config)

praiskup:
These are small things, let me know if you prefer me to do that in a separate commit...

praiskup:
Oh, taking back ... that is not a c10s test-case!   And actually, the issue #1442 is related to CentOS Stream 8+, not just 10 - we we are not fixing the issue right now.

Can you please keep the issue open, i.e., s/Fixes/Relates/, keep the test-case as-is, but provide the release notes snippet?

carlwgeorge:
No problem at all, if it makes sense in the same commit then that's how we should do it.  I've amended again with those adjustments, let me know if there are any other tweaks needed.

praiskup:
```
PASSED: centos-stream+epel-10-x86_64.cfg (tmpfs)
PASSED: centos-stream+epel-10-x86_64.cfg
```

praiskup (approved):
LGTM

PR #1456, opened 2024-09-19
Respect nspawn_args whenever doChroot is called

praiskup:
Which also includes all doOutChroot(), because that method calls doChroot() internally.

This issue was found when trying to workaround an issue [1] when SELinux policy forbid systemd-machine to create a varlink socket and thus start. This resulted in systemd-nspawn not being able to register a machine. To workaround this, Tomáš added the following snippet to the configuration:

    config_opts['nspawn_args'] = ['--register=no']

So that systemd-nspawn does not try to register the machine with systemd-machine.  However, this had no effect (and the argument was not visible on command-line and still failed).

[1] https://issues.redhat.com/browse/RHEL-49567

Co-authored-with: Tomáš Hozza <thozza@redhat.com>

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -65,12 +65,6 @@ class Commands(object):
         self.private_network = not config['rpmbuild_networking']
         self.rpmbuild_noclean_option = None
 
-    def _get_nspawn_args(self):
-        nspawn_args = []
-        if util.USE_NSPAWN:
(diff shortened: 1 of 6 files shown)


praiskup:
Please ignore the similarity warning, it _is_ an issue, but it is not a new thing (specific format that breaks csdiff comparison I believe).

xsuchy (approved):
+1

PR #1461, opened 2024-09-24
packit: specify mock-deps as additional_repo

praiskup:


praiskup:
ci-only, merging early

PR #1464, opened 2024-09-25
move opensuse-leap-15.4 to eol directory

xsuchy:
Leap 15.4 EOLed at 07 Dec 2023 according to https://endoflife.date/opensuse


xsuchy:
Seems that regexp for towncrier snippets does not like dots in filename. Should I fix the name or can we ignore it?

praiskup:
> Should I fix the name or can we ignore it?

I think yes.  Reported here: https://github.com/sanitizers/chronographer-github-app/issues/44

praiskup (approved):

webknjaz:
@xsuchy I haven't checked your towncrier use yet, but it's not just the bot, the underlying tool itself parses the fragments similarly since period-delimited components have defined semantics, unless the fragment is orphaned. My immediate instinct would be fixing the file name.

xsuchy:
@webknjaz I will fix the filename. That is not a problem. But I normally executed `towncrier create leap-15.4.config` because that seemed natural in this context and towncrier allowed it while the bot later complained.

xsuchy:
File renamed.

webknjaz:
The first part `leap-15` is an issue number in Towncrier. You just happen not to render it. Then, there's `4` which is used for sorting if there's several fragments with that issue number that you don't want ordered randomly. And only after that, there's the category.

Have you tried `towncrier check` on that fragment? I think it might be a bug on their side with insufficient validation.

The bot fails because it expects the identifier to be a number. The traditional use is putting an issue or a PR ID there, which is the common pattern the bot facilitates originally. I'll need to think if that could be improved.

webknjaz:
Oh, wait, the number comes after the category. So towncrier will thing that `4` is the category. Try running `towncrier build --draft` with this fragment. I'm pretty sure it'll either not show up at all or traceback somewhere..

praiskup:
We are using a custom template: https://github.com/rpm-software-management/mock/blob/acf15e81b0c858c3dc0341ec0afbe61cfdd2cc7d/pyproject.toml#L6

And the overall use of towncrier is non-standard (the tool doesn't seem to ready to handle monorepos, and we should somehow standardize our release policies for release notes).

The commit looks good, merging!  Thank you.

webknjaz:
@praiskup yeah, I've found it already. For monorepos, you could have multiple configs and run towncrier once per each subproject, making multiple change log files that you'd then merge in post-processing, I suppose. You can also have multiple configs for different audiences in the same project, [a-la scriv](https://scriv.rtfd.io), by calling towncrier differently. If you can outline the needs, I can look into it more in-depth.

Issue #1467, opened 2024-09-26
Mock resultdir created by chroot_scan plugin has weird permissions


praiskup:
Two subsequent calls of mockchain fail with:

```
$ mockchain /tmp/tito/python-copr-1.132-1.git.75.658230d.fc41.src.rpm /tmp/tito/copr-cli-1.112-1.git.78.658230d.fc41.src.rpm

## mockchain is deprecated, use mock --chain instead ##

INFO: mock.py version 5.6.post1 starting (python version = 3.13.0, NVR = mock-5.6.post1-1.git.3802.3dd0024.fc41), args: /usr/libexec/mock/mock --chain /tmp/tito/python-copr-1.132-1.git.75.658230d.fc41.src.rpm /tmp/tito/copr-cli-1.112-1.git.78.658230d.fc41.src.rpm
Start(bootstrap): init plugins
INFO: tmpfs initialized
INFO: selinux enabled
INFO: chroot_scan: initialized
Finish(bootstrap): init plugins
Start: init plugins
INFO: tmpfs initialized
INFO: selinux enabled
INFO: chroot_scan: initialized
Finish: init plugins
INFO: Signal handler active
Start: run
INFO: chroot_scan: 1 files copied to /var/lib/mock/fedora-41-x86_64/result/chroot_scan
INFO: /var/lib/mock/fedora-41-x86_64/root/var/log/dnf5.log
Traceback (most recent call last):
  File "/usr/libexec/mock/mock", line 1135, in <module>
    exitStatus = main()
  File "/usr/lib/python3.13/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 887, in main
    result = run_command(options, args, config_opts, commands, buildroot)
  File "/usr/lib/python3.13/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 930, in run_command
    commands.init(do_log=True)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.13/site-packages/mockbuild/backend.py", line 156, in init
    self.buildroot.resetLogging()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
[...]

PR #1468, opened 2024-09-26
Fix the permissions for (re)created resultdir

praiskup:
Fixes: #1467

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -477,7 +477,6 @@ class Commands(object):
                     resultdir = os.path.join(self.config['local_repo_dir'], pdn)
                     self.buildroot.resultdir = resultdir
                     self.buildroot.resetLogging(force=True)
-                    file_util.mkdirIfAbsent(resultdir)
                     success_file = os.path.join(resultdir, 'success')
                     build_ret_code = 0
                     try:
--- a/mock/py/mockbuild/plugins/chroot_scan.py
+++ b/mock/py/mockbuild/plugins/chroot_scan.py
@@ -13,7 +13,7 @@ import subprocess
 
 # our imports
 from mockbuild.trace_decorator import getLog, traceLog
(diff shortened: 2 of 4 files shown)


praiskup:
I plan to merge this early and wrap the release (once the CI gets green).  Please comment ASAP.

mock/py/mockbuild/backend.py
```diff
@@ -477,7 +477,6 @@ def chain(self, args, options, buildroot):
                     resultdir = os.path.join(self.config['local_repo_dir'], pdn)
                     self.buildroot.resultdir = resultdir
                     self.buildroot.resetLogging(force=True)
-                    file_util.mkdirIfAbsent(resultdir)
```
xsuchy, line 480: Uff. I'm not sure about this. But yeah, in theory it should not be needed. Let be brave and do it.

xsuchy (approved):
+1

mock/py/mockbuild/backend.py
praiskup: https://github.com/rpm-software-management/mock/blob/8122bda5d159f75f068199b87ae6172325cd436e/mock/py/mockbuild/buildroot.py#L627-L632

This is the force=True case, I think we are safe.

mock/py/mockbuild/plugins/chroot_scan.py
```diff
@@ -56,7 +56,7 @@ def __scanChroot(self):
         regexstr = "|".join(self.scan_opts['regexes'])
         regex = re.compile(regexstr)
         chroot = self.buildroot.make_chroot_path()
-        file_util.mkdirIfAbsent(self.resultdir)
+        self.buildroot.create_resultdir()
```
praiskup, line 59: Argh! While it _was_ causing issues before, now it is broken as well -> because `self.resultdir` is not `serlf.buildroot.resultdir`, nah.  chroot_scan is broken in 5.7

PR #1473, opened 2024-09-27
configs: update ELN repos

yselkowitz:
ELN now has Pungi composes instead of OSBS, and are downloadable and mirrorable like other Fedora-hosted projects.  Also, with multilibs no longer being built, the mirrored conditional is no longer necessary.

### DIFF
--- a/mock-core-configs/etc/mock/templates/fedora-eln.tpl
+++ b/mock-core-configs/etc/mock/templates/fedora-eln.tpl
@@ -3,11 +3,6 @@ config_opts['eln_rawhide_releasever'] = '42'
 
 config_opts['root'] = 'fedora-eln-{{ target_arch }}'
 
-# Fedora ELN i386 doesn't get composes (isn't mirrored on
-# odcs.fedoraproject.org), we need to build using the Koji buildroot.
-# Note that similar idiom used in fedora-branched.tpl and fedora-rawhide.tpl.
-config_opts['mirrored'] = config_opts['target_arch'] != 'i686'
-
 config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils fedora-release-eln findutils gawk glibc-minimal-langpack grep gzip info patch redhat-rpm-config rpm-build sed tar unzip util-linux which xz'
 
 config_opts['dist'] = 'eln'  # only useful for --resultdir variable subst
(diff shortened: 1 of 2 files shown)


sgallagher (approved):
LGTM

I will note that `config_opts['bootstrap_image'] = 'quay.io/fedoraci/fedora:eln'` will also need to change in the near future, but it's not quite ready yet. The existing bootstrap container is still present, but hasn't been updated for a while. It will be changing in the near future to `quay.io/fedora/eln:latest`, but we haven't had an upload to that location yet.

praiskup:
Can you please add a changelog entry?
hhttps://rpm-software-management.github.io/mock/Release-Notes-New-Entry

mock-core-configs/etc/mock/templates/fedora-eln.tpl
```diff
@@ -53,11 +48,10 @@ file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primar
 {%- endfor %}
 {%- endmacro %}
 
-{% if mirrored %}
 [eln-baseos]
 name=Fedora - ELN BaseOS - Developmental packages for the next Enterprise Linux release
-baseurl=https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose/BaseOS/$basearch/os/
-#metalink=https://mirrors.fedoraproject.org/metalink?repo=eln&arch=$basearch
+#baseurl=https://dl.fedoraproject.org/pub/eln/1/BaseOS/$basearch/os/
+metalink=https://mirrors.fedoraproject.org/metalink?repo=eln-baseos-1&arch=$basearch
```
praiskup, line 54: Why the `-1` suffix?  Are we going to bump in the future?  How often?  Do we want to have a variable for this?

sgallagher: That's a limitation of mirror manager. It has no actual meaning and should never change.

Mirror manager just can't cope without a "version" (and adapting every place that it special-cases "rawhide" was prohibitively difficult). So we just decided it would be version "1" forever.

praiskup (approved):

yselkowitz:
> I will note that `config_opts['bootstrap_image'] = 'quay.io/fedoraci/fedora:eln'` will also need to change in the near future, but it's not quite ready yet. The existing bootstrap container is still present, but hasn't been updated for a while. It will be changing in the near future to `quay.io/fedora/eln:latest`, but we haven't had an upload to that location yet.

Uploads have begun, so updated this accordingly.

Issue #1427, opened 2024-08-15
mock-core-configs: maintenance of EPEL 10 minor versions


praiskup:
The problem was discussed in #1424, the `[local]` Koji repository contains `10.0` string, and we'll need to move it to `10.1` at some point (maintenance hell).  We should avoid this inconvenience.

Important reading first:
- https://discussion.fedoraproject.org/t/epel-10-proposal/44304

carlwgeorge:
I think we have two possible options we can take for EPEL 10 mock configs.  Which option we choose affects how the koji local baseurl problem would be handled.

### option 1

We fully embrace minor versions for EPEL and RHEL in mock-core-configs.  This could look something like this:

```
templates/centos-stream-10.tpl
centos-stream-10-aarch64.cfg
centos-stream-10-ppc64le.cfg
centos-stream-10-s390x.cfg
centos-stream-10-x86_64.cfg

templates/rhel-10.0.tpl
rhel-10.0-aarch64.cfg
rhel-10.0-ppc64le.cfg
rhel-10.0-s390x.cfg
rhel-10.0-x86_64.cfg

templates/epel-10.tpl
centos-stream+epel-10-aarch64.cfg
centos-stream+epel-10-ppc64le.cfg
centos-stream+epel-10-s390x.cfg
centos-stream+epel-10-x86_64.cfg
epel-10-aarch64.cfg -> centos-stream+epel-10-aarch64.cfg  (symlink)
epel-10-ppc64le.cfg -> centos-stream+epel-10-ppc64le.cfg  (symlink)
epel-10-s390x.cfg   -> centos-stream+epel-10-s390x.cfg    (symlink)
epel-10-x86_64.cfg  -> centos-stream+epel-10-x86_64.cfg   (symlink)

templates/epel-10-branched.tpl
rhel+epel-10.0-aarch64.cfg
rhel+epel-10.0-ppc64le.cfg
rhel+epel-10.0-s390x.cfg
rhel+epel-10.0-x86_64.cfg
```

[...]

PR #1484, opened 2024-10-09
Add epel and epel-testing repos to the EPEL 10 config

carlwgeorge:
Previously the EPEL 10 config only included the koji local repo, because that was the only thing available at the time.  We now have working MirrorManager metalinks that we can use for the proper repos.  Now that we have those repos, we can set the koji local repo to disabled by default, matching earlier EPEL configs.

We also now have a koji symlink that will be updated over time to point to the latest minor version tag buildroot repo, which can be used in the koji local baseurl.  This avoids the inconvenience of updating that baseurl manually every six months.

Fixes: #1427

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
@@ -1,6 +1,7 @@
 config_opts['chroot_setup_cmd'] = 'install bash bzip2 centos-stream-release coreutils cpio diffutils findutils gawk glibc-minimal-langpack grep gzip info patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
 config_opts['dist'] = 'el10'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '10'
+config_opts['releasever_major'] = '10'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['description'] = 'CentOS Stream 10'
--- a/mock-core-configs/etc/mock/templates/epel-10.tpl
+++ b/mock-core-configs/etc/mock/templates/epel-10.tpl
@@ -2,10 +2,39 @@ config_opts['chroot_setup_cmd'] += " epel-rpm-macros"
 
 config_opts['dnf.conf'] += """
 
+# The metalinks below differ from the ones used in epel-release.  The
+# epel-release metalinks rely on a dnf variable expansion feature that is only
+# available since dnf 4.18.0.  In some configurations mock will be using the
+# host dnf to bootstrap the chroot, which could be older and lack those
+# features, leading to problems.  The alternative is to use this simplified
+# metalink here in this template for CentOS Stream, and a different metalink in
+# another template for RHEL.
+
+[epel]
+name=Extra Packages for Enterprise Linux $releasever - $basearch
+metalink=https://mirrors.fedoraproject.org/metalink?repo=epel-{{ releasever_major }}&arch=$basearch
+gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-{{ releasever_major }}
+gpgcheck=1
+countme=1
+
+[epel-testing]
+name=Extra Packages for Enterprise Linux $releasever - Testing - $basearch
+metalink=https://mirrors.fedoraproject.org/metalink?repo=testing-epel{{ releasever_major }}&arch=$basearch
+gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-{{ releasever_major }}
+gpgcheck=1
+countme=1
+enabled=0
+
+{% if koji_primary_repo == "epel" %}
+
+# The baseurl below is a symlink in the koji infrastructure that moves forward
+# over time to the buildroot repo of the latest minor verison tag.
+
 [local]
 name=Extra Packages for Enterprise Linux $releasever - Koji Local - BUILDROOT ONLY!
-baseurl=https://kojipkgs.fedoraproject.org/repos/epel10.0-build/latest/$basearch/
+baseurl=https://kojipkgs.fedoraproject.org/repos/epel{{ releasever_major }}/centos/$basearch/
 cost=2000
-enabled=1
+enabled=0
 skip_if_unavailable=False
+{% endif %}
 """
--- a/mock-core-configs/etc/mock/epel-10-aarch64.cfg
+++ b/mock-core-configs/etc/mock/epel-10-aarch64.cfg
@@ -1,0 +1,1 @@
+centos-stream+epel-10-aarch64.cfg
--- a/mock-core-configs/etc/mock/epel-10-ppc64le.cfg
+++ b/mock-core-configs/etc/mock/epel-10-ppc64le.cfg
@@ -1,0 +1,1 @@
+centos-stream+epel-10-ppc64le.cfg
--- a/mock-core-configs/etc/mock/epel-10-s390x.cfg
+++ b/mock-core-configs/etc/mock/epel-10-s390x.cfg
@@ -1,0 +1,1 @@
+centos-stream+epel-10-s390x.cfg
--- a/mock-core-configs/etc/mock/epel-10-x86_64.cfg
+++ b/mock-core-configs/etc/mock/epel-10-x86_64.cfg
@@ -1,0 +1,1 @@
+centos-stream+epel-10-x86_64.cfg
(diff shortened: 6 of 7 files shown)


mock-core-configs/etc/mock/templates/epel-10.tpl
```diff
@@ -2,10 +2,35 @@ config_opts['chroot_setup_cmd'] += " epel-rpm-macros"
...
+# available since dnf 4.18.0.  In some configurations mock will be using the
+# host dnf to bootstrap the chroot, which could be older and lack those
+# features, leading to problems.  The alternative is to use this simplified
+# metalink here in this template for CentOS Stream, and a different metalink in
+# another template for RHEL.
+
+[epel]
+name=Extra Packages for Enterprise Linux $releasever - $basearch
+metalink=https://mirrors.fedoraproject.org/metalink?repo=epel-$releasever_major&arch=$basearch
+gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-$releasever_major
```
Conan-Kudo, line 16: Was `$releasever_major` backported to RHEL 8 DNF? Otherwise, this might only work on RHEL 9+ unless you use a bootstrap image.

carlwgeorge: This was discussed at length in #1427, and is why this PR also defines `releasever_major` in centos-stream-10.tpl.

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
```diff
@@ -1,6 +1,7 @@
 config_opts['chroot_setup_cmd'] = 'install bash bzip2 centos-stream-release coreutils cpio diffutils findutils gawk glibc-minimal-langpack grep gzip info patch redhat-rpm-config rpm-build sed shadow-utils tar unzip util-linux which xz'
 config_opts['dist'] = 'el10'  # only useful for --resultdir variable subst
 config_opts['releasever'] = '10'
+config_opts['releasever_major'] = '10'
```
Conan-Kudo, line 4: This doesn't do anything. There's nothing wired to this option in Mock.

Conan-Kudo (changes requested):

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
carlwgeorge: I could adjust the config to only use `releasever`, I was just trying to be explicit about the places that are definitely always just the major version, even if someone tries to use the template on a distro that sets `releasever` to a full version such as 10.0.  What do you think @praiskup?  If it's preferable to stick with `releasever_major` I can send a separate PR to "wire it up" as Neal described (but I could use a pointer to where in the code this happens).

Conan-Kudo: `I would certainly prefer it to be wired up (alongside `releasever_minor`)  once that becomes useful.

Conan-Kudo: I would certainly prefer it to be wired up (alongside `releasever_minor`)  once that becomes useful.

mock-core-configs/etc/mock/templates/epel-10.tpl
praiskup: The problem Neal describes (at least IIUC) is that Mock doesn't map its configuration options to DNF variables. Would you mind doing `repo=epel-{{ releasever_major }}` instead?

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
praiskup: By defining `releasever_major = 10` in this template you wire it up; then just use the `{{ jinja }}` statement to read that value.

praiskup:
Thank you for the follow-up PR.

Just not sure, s/Fixes: #1427/Relates: #1427/ ? This doesn't seem to deal with the minor versions.

carlwgeorge:
Since this replaced the koji local baseurl with one that didn't need the minor version (the symlink you asked for), I felt it resolved that issue under the previous title of "The EPEL10 local repository now bakes-in RHEL minor version".  The conversation in that issue certainly grew to be wider in scope, so I can understand the title change, and agree that it would change fixes to relates in the context of this PR.

mock-core-configs/etc/mock/templates/centos-stream-10.tpl
carlwgeorge: So if I'm understanding you right, `config_opts['releasever']` doesn't actually correspond to `$releasever` in templates?  I can see how https://github.com/rpm-software-management/mock/blob/main/mock-core-configs/etc/mock/templates/centos-stream-9.tpl uses `{{ releasever }}`, so I can do the same style.

carlwgeorge: Would wiring it up properly just involve adding `releasever_major` to https://github.com/rpm-software-management/mock/blob/main/mock/py/mockbuild/package_manager.py#L455-L458 ?

praiskup: Good catch there; that mechanism though probably only affects bootstrap at this point in time (and some users still --no-bootstrap-chroot is still supported).  I am not saying that we shouldn't change mock so this substitution works everywhere, but the change would mean new release of mock (not just configs), and probably a dedicated separate pull-request.

carlwgeorge: Understood, I'll start with the jinja style for now, and look into the other way in the future.  Thanks for the clarification.

mock-core-configs/etc/mock/templates/epel-10.tpl
```diff
@@ -2,10 +2,35 @@ config_opts['chroot_setup_cmd'] += " epel-rpm-macros"
...
+gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-{{ releasever_major }}
+gpgcheck=1
+countme=1
+enabled=0
+
+{% if koji_primary_repo == "epel" %}
 [local]
 name=Extra Packages for Enterprise Linux $releasever - Koji Local - BUILDROOT ONLY!
-baseurl=https://kojipkgs.fedoraproject.org/repos/epel10.0-build/latest/$basearch/
+baseurl=https://kojipkgs.fedoraproject.org/repos/epel{{ releasever_major }}/centos/$basearch/
```
praiskup, line 31: Is this a floating symlink to some other repository eventually?  Can you perhaps document the trick here, or link a documentation that documents it?

praiskup:
Thank you, except that I'd like to understand the `epel10/centos` trick first, LGTM

Conan-Kudo (approved):

mock-core-configs/etc/mock/templates/epel-10.tpl
carlwgeorge: It's a symlink in the `/mnt/fedora_koji` network share in the Fedora infrastructure, similar to `rawhide/latest`.

```
[carlwgeorge@compose-rawhide01 repos][PROD-IAD2]$ pwd
/mnt/fedora_koji/koji/repos
[carlwgeorge@compose-rawhide01 repos][PROD-IAD2]$ ll rawhide/latest
lrwxrwxrwx. 1 root root 19 Aug 13 23:52 rawhide/latest -> ../f42-build/latest
[carlwgeorge@compose-rawhide01 repos][PROD-IAD2]$ ll epel10/centos
lrwxrwxrwx. 1 root root 24 Oct  9 01:54 epel10/centos -> ../epel10.0-build/latest
```

My thinking on the symlink naming is that it would be best to use the names centos and rhel.  In the other issues we've thrown around name pairs like development+stable and latest+non-latest, but none of those sit quite right with me.  It's the latest EPEL for CentOS and the latest EPEL for RHEL, so I think the symlinks should just match that.  In the future, we'll have `epel10/centos -> ../epel10.1-build/latest` and `epel10/rhel -> ../epel10.0-build/latest`, and they will move forward over time.

I don't have this written down in formal documentation yet, but my team (CPE EPEL) is tracking our EPEL 10 work in some hackmd docs, which will serve as the basis for a future SOP in the releng and/or EPEL docs.  I've got steps in there for adjusting these symlinks over time.

- [EPEL 10.0 work tracker](https://hackmd.io/@carlwgeorge/ry_xGk35C)
- [EPEL 10.1 work tracker](https://hackmd.io/@carlwgeorge/Sk3j-12qA)

I can add a comment above this line with a basic summary that the link is a symlink that will move forward over time.

praiskup: Thank you for the update! Makes sense.

praiskup (approved):

Issue #289, opened 2019-07-11
replace groupadd (in mock-core-configs) with sysusers.d


xsuchy:
See https://manpages.debian.org/stretch-backports/systemd/sysusers.d.5.en.html

PR #1498, opened 2024-11-01
mock: avoid the %pre scriptlet if possible

praiskup:
The script for generating the %pre scriptlet is non-existing on new distributions (like Mageia now, but Fedora will drop that sooner or later, too).  The build of Mock on Mageia previously relied on sysusers.generate-pre.sh, and failed.

The sysusers feature is an RPM built-in anyway (on modern systems), so there's no reason to duplicate the logic in %pre.  Generate the %pre scriptlet only for older distributions.

Complements: #1367
Relates: #289
Relates: https://github.com/fedora-copr/copr/issues/2789

### DIFF
--- a/mock/mock.spec
+++ b/mock/mock.spec
@@ -1,6 +1,18 @@
 %bcond_with lint
 %bcond_without tests
 
+# Modern distributions (using RPM v4.19+; for example, Fedora 39+) do not
+# require the %%pre scriptlet for creating users/groups because the sysusers
+# feature is now built directly into RPM.  Simply including the sysusers
+# `mock.conf` file in a package payload is sufficient to leverage this feature.
+# However, for older distributions that lack this capability, we still define
+# the %%pre scriptlet.
+%if (0%{?rhel} && 0%{?rhel} < 10) || (0%{?mageia} && 0%{?mageia} < 10) || (0%{?suse_version} && 0%{?suse_version} < 1660)
+%bcond_without sysusers_compat
+%else
+%bcond_with sysusers_compat
+%endif
+
 %global __python %{__python3}
 %global python_sitelib %{python3_sitelib}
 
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Wouldn't it be simpler to just vendor the script in the packaging so that it works everywhere?

praiskup:
@Conan-Kudo PTAL

praiskup:
@Conan-Kudo does it make sense now?

praiskup:
I think I did what you mentioned (not sure). This is tricky though, it would be nice to see a "reference" implementation.

Conan-Kudo (approved):

mock/mock.spec
```diff
@@ -1,6 +1,19 @@
...
+# sysusers feature.  To enable it, we simply include the `mock.conf` file in the
+# package payload so that RPM knows what actions to take.  For older
+# distributions, however, %%pre is needed.  On some of those older distros the
+# %%sysusers_*_compat macros could help (but not everywhere), so turns out it is
+# just way easier and to expand those macros and hardcode them in this specfile.
+%if 0%{?rhel} && 0%{?rhel} <= 9
+%bcond_without sysusers_compat
+%else
+%bcond_with sysusers_compat
+%endif
```
Conan-Kudo, line 15: This needs to be flipped. We only know for _sure_ that newer Fedora and RHEL support it, and everything else we don't know.

Conan-Kudo (changes requested):

mock/mock.spec
Conan-Kudo:
```suggestion
%if 0%{?fedora} || 0%{?rhel} >= 10
%bcond_with sysusers_compat
%else
%bcond_without sysusers_compat
%endif
```

Conan-Kudo: Or alternatively if we assume everyone is going to get it, then we need something like this:

```suggestion
%if (0%{?rhel} && 0%{?rhel} < 10) || (0%{?mageia} && 0%{?mageia} < 10) || (0%{?suse_version} && 0%{?suse_version} < 1660)
%bcond_without sysusers_compat
%else
%bcond_with sysusers_compat
```

praiskup: the first suggestion accepted, thank you

praiskup:
ptal

praiskup:
Updated, @Conan-Kudo & @nikromen could you please take a look?

PR #1505, opened 2024-11-15
mock: move backoff python requirement

c0m1c5an5:
Move python backoff requirement from pytest requirements to the mock requirements.txt as it is not a test dependency.
This change does not require release notes, as it does not affect the RPM build process.

### DIFF
--- a/mock/requirements.txt
+++ b/mock/requirements.txt
@@ -1,7 +1,7 @@
+backoff
 distro
 jinja2
 pyroute2
 requests
 rpmautospec-core
 templated-dictionary
-backoff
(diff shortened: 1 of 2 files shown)


mock/requirements.txt
```diff
@@ -4,3 +4,4 @@ pyroute2
 requests
 rpmautospec-core
 templated-dictionary
+backoff
```
praiskup, line 7: Thank you, would you mind moving it up to keep the alphabetical order?

c0m1c5an5: Sure.

praiskup: Thank you!

PR #1508, opened 2024-11-20
Use non development bootstrap image for CentOS Stream 10

Romain-Geissler-1A:
cc @carlwgeorge Who did the last commit using the development image for CentOS 10 Streams.

### DIFF
--- a/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
+++ b/mock-core-configs/etc/mock/templates/centos-stream-10.tpl
@@ -6,7 +6,7 @@ config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['description'] = 'CentOS Stream 10'
 
-config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream10-development'
+config_opts['bootstrap_image'] = 'quay.io/centos/centos:stream10'
 config_opts['bootstrap_image_ready'] = True
 
 config_opts['dnf.conf'] = """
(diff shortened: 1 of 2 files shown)


carlwgeorge:
Yup, I can confirm the regular `stream10` tag was added, see [this issue](https://issues.redhat.com/browse/CS-2618).

carlwgeorge (approved):

praiskup (approved):
Awesome, thank you!

praiskup:
PASSED: centos-stream-10-x86_64.cfg (tmpfs)
PASSED: centos-stream-10-x86_64.cfg

martinpitt:
@praiskup @Romain-Geissler-1A Please please revert this. The production compose lags behind weeks and months, see https://issues.redhat.com/browse/CS-2309 . I just tried to move our CentOS 10 image from devel to prod compose, and it's at least a 5 week step backwards. This will make packages build against outdated libraries/toolchains.

The CentOS 10 composes have been a mess since their inception, and this needs sorting out, see also the last comments on https://issues.redhat.com/browse/CS-2642 . But as of today this switch is still premature.

carlwgeorge:
This should not be reverted.  Neither container image is pointing to the compose repos.  They have identical repo files which point to the mirror network, so they see the same packages for the purposes of mock.  Reverting this PR won't change what packages are visible.  Furthermore, for the examples given in the linked issues, the packages baked into the image are all either the same or newer in the `centos:stream10` image, so it's objectively better to keep using this image.

This is the correct tag to use going forward, as per https://issues.redhat.com/browse/CS-2618.  Any issues with the composes is orthogonal to this mock change.

martinpitt:
> Neither container image is pointing to the compose repos. They

Assuming s/compose/devel/, ok -- so in effect this has always been outdated? Thanks.

carlwgeorge:
Well, there are two composes, the development compose and the production compose.

- https://composes.stream.centos.org/stream-10/development/latest-CentOS-Stream/compose/
- https://composes.stream.centos.org/stream-10/production/latest-CentOS-Stream/compose/

The production composes are what gets pushed out to the main mirror and duplicated through the mirror network.

- https://mirror.stream.centos.org/10-stream/

Both the `centos:stream10-development` and `centos:stream10` images are both configured to use the mirror network and see the same packages.

PR #1514, opened 2024-12-02
Use layers digests for comparing podman images

tkopecek:
Previous method checked default digest provided by podman. This digest is "local" and changed every time image is saved/load or at any other point manifest is modified. This doesn't mean that it is a different image.

Viable way for our purposes is to compare that all layers are identical and in the same order. Simple way to distill this into one value is to concatenate individual layers' digests in order of appearance in RootFS.

# Add a reference to related issue - preferably in the git commit message
Fixes: #NUMBER

# Add release note. See https://rpm-software-management.github.io/mock/Release-Notes-New-Entry
# or let us know to add `label no-release-notes` if you think the change does
# not deserve a release note.


tkopecek:
@praiskup Do you think this is the viable approach for comparing images?

tkopecek:
Proposal from slack is to also add checking of Config section. If we decide for this I can add e.g. checksum of sorted dict of Config.

mock/docs/buildroot-lock-schema-1.0.0.json
```diff
@@ -82,8 +82,8 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "image_digest": {
-          "description": "Digest got by the 'podman image inspect --format {{ .Digest }}' command, sha256 string",
+        "image_layers_digest": {
+          "description": "SHA256 digest concatenated layer digests from 'podman image inspect --format {{ .RootFS }}' command, sha256 string",
```
praiskup, line 86: Do we want to keep both?  I'm not a jsonschema guru, does this imply a schema version change @xsuchy ?

mock/py/mockbuild/buildroot.py
```diff
@@ -283,7 +283,7 @@ def _fallback(message):
                 if digest_expected:
                     getLog().info("Checking image digest: %s",
                                   digest_expected)
-                    digest = podman.get_image_digest()
+                    digest = podman.get_layers_digest()
                     if digest != digest_expected:
                         getLog().warning(
```
praiskup, line 288: Do we want to make this failing again?

praiskup:
> @praiskup Do you think this is the viable approach for comparing images?

Probably? I'm not the best person to judge :-) have you tried to contact someone from the Podman team?

mock/py/mockbuild/buildroot.py
tkopecek: Probably needs a bit of testing in the wild. I would turn it on by default e.g. one or two version later.

mock/docs/buildroot-lock-schema-1.0.0.json
praiskup: I'm not a JSON person :-) this was my first scheme I prepared... 

But I think I'd prefer to add version 2.0.0 here, or 1.1 (if we keep the old `image_digest` field).  Or rather radical change - drop the versioning entirely (because it seems just wrong to change the schema within the same version this way).

mock/py/mockbuild/podman.py
```diff
@@ -114,21 +116,33 @@ def mounted_image(self):
...
+        except json.JSONDecodeError as exc:
+            raise BootstrapError(f"The RootFS layers of {self.image} "
+                                  "are not json-formatted.") from exc
+        except (KeyError, TypeError) as exc:
+            raise BootstrapError(f"The Layers of {self.image} "
+                                  "are not in expected format.") from exc
+        sha = hashlib.sha256()
+        for layer in layers:
+            sha.update(layer.encode()) # as it was decoded from json, it is unicode again
+        return sha.hexdigest()
```
praiskup, line 145: This doesn't give us a JSON formatted output:
```
$ podman image inspect -f=json --format '{{ .RootFS }}' registry.access.redhat.com/ubi8/ubi 
{layers [sha256:3200b171908c5b74d4e1781a79a0fc729b13c29fc3ce70687384c8cc349f509d]}
```

The `podman image inspect` output seems to be JSON formatted by default, unless you provide the option `--format`.

praiskup:
```
diff --git a/mock/py/mockbuild/podman.py b/mock/py/mockbuild/podman.py
index 1a1063cf..6287013e 100644
--- a/mock/py/mockbuild/podman.py
+++ b/mock/py/mockbuild/podman.py
@@ -122,20 +122,19 @@ class Podman:
         all images containing same order of layers, thus it can be used
         as the check that we've loaded same image.
         """
-        check = [self.podman_binary, "image", "inspect", self.image,
-                 "--format", "{{ .RootFS }}"]
+        check = [self.podman_binary, "image", "inspect", self.image]
         result = subprocess.run(check, stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE, check=False,
                                 encoding="utf8")
         if result.returncode:
-            raise BootstrapError(f"Can't get {self.image} podman image digest: {result.stderr}")
+            raise BootstrapError(f"Can't inspect {self.image} podman image: {result.stderr}")
         result = result.stdout.strip()
         try:
-            rootfs = json.loads(result)
-            layers = rootfs['layers']
+            inspect_data = json.loads(result)
+            layers = inspect_data[0]['RootFS']['Layers']
         except json.JSONDecodeError as exc:
-            raise BootstrapError(f"The RootFS layers of {self.image} "
-                                  "are not json-formatted.") from exc
+            raise BootstrapError(f"The inspect output of {self.image} "
+                                  "is not a json-formatted string.") from exc
         except (KeyError, TypeError) as exc:
             raise BootstrapError(f"The Layers of {self.image} "
                                   "are not in expected format.") from exc

**Comment by praiskup** [comment]
Tested with `fedora-rawhide-x86_64` and
```
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -285,7 +285,7 @@ class Buildroot(object):
                                   digest_expected)
                     digest = podman.get_layers_digest()
                     if digest != digest_expected:
-                        getLog().warning(
+                        raise BootstrapError(
                             f"Expected digest for image {podman.image} is"
                             f"{digest_expected}, but {digest} found.")
```
And it seems to work.

**Comment by praiskup** [comment]
Tested with `centos-stream+epel-{9,10}-x86_64`  and it works as well.

**Comment by praiskup** [comment]
Can you please rebase on top of the current `main`?   That should make CI green.

Also note that there's https://github.com/rpm-software-management/mock/blob/main/behave/features/hermetic-build.feature, so the `testing-farm:fedora-40-x86_64:mock` should be green first before we merge this.

It would be nice to mention the [podman discussion I started](https://github.com/containers/podman/discussions/24818) somewhere in the code (or commit message).

**Comment by praiskup** [comment]
CI fail: ` ERROR: Config section of registry.fedoraproject.org/fedora:rawhide is missing.`

**Comment by praiskup** [comment]
` ERROR: RootFS section of quay.io/centos/centos:stream9 is missing.`

File: mock/py/mockbuild/podman.py [hunk 4]
```diff
@@ -145,7 +143,7 @@ def get_layers_digest(self):
         # data which should be sufficient to confirm the image
         data = {
             'RootFS': data['RootFS'],
-            'Config': data['Config'],
+            'Config': data.get('Config'),
```
praiskup, line 146: Hmm, `podman image inspect fedora:41 --format "{{ .RootFS.Layers }} {{ .Config }}"` works.  Same for `quay.io/centos/centos:stream9`.

I think we need to access the first item in the upper-level array, like:
`podman image inspect quay.io/centos/centos:stream9 | jq '.[0].RootFS'`

tkopecek: updated (and tested with fedora rawhide)

praiskup: works: https://artifacts.dev.testing-farm.io/2ce4be47-b383-4139-a622-29dbe950709f/

mock/py/mockbuild/podman.py
```diff
@@ -114,21 +116,38 @@ def mounted_image(self):
...
+                                  "are not json-formatted.") from exc
+        if 'RootFS' not in data:
+            raise BootstrapError(f"RootFS section of {self.image} is missing.")
+        if data['RootFS']['Type'] != 'layers':
+            raise BootstrapError(f"Unexpected format for RootFS in {self.image}.")
+
+        # data which should be sufficient to confirm the image
+        data = {
+            'RootFS': data['RootFS'],
+            'Config': data.get('Config'),
```
praiskup, line 146: Please revert back to `data["Config"]`

mock/docs/buildroot-lock-schema-1.0.0.json
praiskup: Still relevant;  + we should document that we are using the `.Config` value, too

praiskup (changes requested):
- we need to resolve the json versioning issue
- new release notes entry
- I would make it failing on checksum mismatch again, as this method is reliable
- please squash, and refer the commit to https://github.com/containers/podman/discussions/24818

Otherwise thank you, LGTM, nice one :)

mock/py/mockbuild/buildroot.py
```diff
@@ -279,11 +279,11 @@ def _fallback(message):
                                            "bootstrap.tar")
                     podman.import_tarball(tarball)
 
-                digest_expected = self.config.get("image_assert_digest", None)
+                digest_expected = self.config.get("image_assert_digest", True)
                 if digest_expected:
                     getLog().info("Checking image digest: %s",
```
praiskup, line 284: To implement the assert, this needs to be switched to exception raising

mock/py/mockbuild/config.py
```diff
@@ -102,7 +102,7 @@ def setup_default_config_opts():
     config_opts['bootstrap_image_ready'] = False
     config_opts['bootstrap_image_fallback'] = True
     config_opts['bootstrap_image_keep_getting'] = 120
-    config_opts['bootstrap_image_assert_digest'] = None
+    config_opts['bootstrap_image_assert_digest'] = True
```
praiskup, line 105: This is supposed to be the string digest.

praiskup:
Seems to work, the CI failure was probably some cache problem / race condition (after re-run, it is green now) .  What remains is to handle the JSON scheme reasonably.  I think we could keep the old scheme as-is, and not rename the field.  Just change the documentation for the field that it is not the built-in digest.

praiskup:
Nit: See the comment 2, there's one pep8speaks issue.

tkopecek:
> Seems to work, the CI failure was probably some cache problem / race condition (after re-run, it is green now) . What remains is to handle the JSON scheme reasonably. I think we could keep the old scheme as-is, and not rename the field. Just change the documentation for the field that it is not the built-in digest.

So, do you think that also the method should stage ``get_image_digest`` or is ``get_layers_digest`` better?

praiskup:
There's this error though:

> ERROR: Expected digest for image registry.fedoraproject.org/fedora:rawhide isTrue, but 953d47de95893c3502da491bb1100e401ddf23b321dec4f3d7c54e6e095b3252 found.

tkopecek:
> There's this error though:
> 
> > ERROR: Expected digest for image registry.fedoraproject.org/fedora:rawhide isTrue, but 953d47de95893c3502da491bb1100e401ddf23b321dec4f3d7c54e6e095b3252 found.

Fixed, anyway - how you were able to trigger it? Older lockfile?

tkopecek:
> > So, do you think that also the method should stage get_image_digest or is get_layers_digest better?
> 
> No strong opinion, but it's not "just layers", we include ".Config" too.

Nor it is an image digest (as understood in container world).

praiskup:
> Fixed, anyway - how you were able to trigger it? Older lockfile?

That was triggered by the testing-farm CI, the latest one is being run right now by Packit:
https://github.com/rpm-software-management/mock/pull/1514/checks?check_run_id=34539968098

praiskup:
> Nor it is an image digest (as understood in container world).

:man_shrugging: yeah I agree, we can call it `mock_oci_digest` ? :) or pick whatever name you want ... I'd still prefer to expose this digest calculation as a separate `from mockbuild.podman podman_mock_image_digest` method (or alike). Just to have the method published "somewhere", so other users may start doing the thing we do here.. (and e.g., check that mock's digest is valid).

mock/py/mockbuild/podman.py
```diff
@@ -132,26 +134,38 @@ def mounted_image(self):
...
+        as the check that we've loaded same image.
         """
-        the_image = self.image
-        if the_image.startswith("oci-archive:"):
-            # We can't query digest from tarball directly, but note
-            # the image needs to be tagged first!
-            the_image = self._tagged_id
-        check = [self.podman_binary, "image", "inspect", the_image,
-                 "--format", "{{ .Digest }}"]
+        check = [self.podman_binary, "image", "inspect", self.image]
```
praiskup, line 143: This still needs to work with `self._tagged_id` if we work with `oci-archive:` images.

praiskup (approved):
LGTM, thank you.

PR #1517, opened 2024-12-05
tox: add Python 3.13, drop 3.7

praiskup:
The version 3.7 is no longer available:
py37: skipped because could not find python interpreter with spec(s): py37


praiskup:
Merging early to not block CI

PR #1520, opened 2024-12-13
releng: mock-from-git to have a higher NVR than released versions

praiskup:
Historically, Tito only modified the Release tag for `--test` builds. Now, we also bump the Version field to ensure that Mock built from Git revisions always has a higher NVR than released packages.

https://github.com/rpm-software-management/tito/releases/tag/tito-0.6.27-1


xsuchy (approved):
+1

Issue #1522, opened 2024-12-14
hermetic: turn off the "normal bootstrap install"


praiskup:
We have faced a weird cache/race issue in #1514

```
     WARNING: Can't initialize from bootstrap image, falling back to bootstrap installation: Expected digest for image registry.fedoraproject.org/fedora:rawhide isb63185f5458c16da45cf43dd0f0f87a32bb0825e03a28569fc625b165641310c, but 16a00d9d0910fd092ae97d3ad32ab712e6af39cd2be0b325763e1e3775170336 found.
      INFO: Package manager dnf5 detected and used (fallback)
```

It doesn't make sense to try to install bootstrap via DNF in hermetic mode.

https://artifacts.dev.testing-farm.io/3a167e23-6785-4ba0-a4db-bc256bf24219/

PR #1524, opened 2024-12-18
hermetic: do not install buildroot via DNF

praiskup:
Fixes: #1522

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -810,6 +810,10 @@ def process_hermetic_build_config(cmdline_opts, config_opts):
     # with the same image.
     config_opts["bootstrap_image_assert_digest"] = data["bootstrap"]["image_digest"]
 
+    # It doesn't make sense to fallback to `dnf install dnf --installroot ...`,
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Why can't we have a fallback?

praiskup:
> Why can't we have a fallback?

C&P form the towncrier snippet: it doesn't make sense because we don't have the bootstrap packages pre-downloaded in the local "offline" repository.

The repository only contains the packages that are needed to re-construct the buildroot repository (no DNF is there, typically).

Conan-Kudo:
So then... how does a build with dynamic buildrequires work? We need DNF there to be able to install more packages...

praiskup:
The hook that generates the lock file is "postdeps"; IOW we have **all** the dynamic build-deps covered in lockfile, and pre-downloaded in the offline repository (but not DNF, as that is typically not in BuildRequires).

praiskup:
> We need DNF there to be able to install more packages...

Yeah and, ... DNF is provided by `bootstrap_image_ready=True`.

praiskup:
self-merging, so I can wrap the release, this seems rather trivial anyway

PR #1534, opened 2025-01-31
doc: `--config-opts` single-item array wants second to be blank

evan-goode:
When specifying an array with a single element using `--config-opts`, the user should specify the second item as empty, not the first:

Wrong:

```
$ mock --config-opts=foo= --config-opts=foo=bar --debug-config | grep foo
config_opts['foo'] = ['', 'bar']
```

Correct:

```
$ mock --config-opts=foo=bar --config-opts=foo= --debug-config | grep foo
config_opts['foo'] = ['bar']
```

The release notes where the feature was announced, Release-Notes-1.4.8.md, were correct but the man page was not.

I think this is probably `label no-release-notes`.


xsuchy (approved):
+1

praiskup (approved):

PR #1533, opened 2025-01-31
Add AlmaLinux Kitten 10 configs

Conan-Kudo:
This pull request adds the AlmaLinux Kitten 10 configs for Mock.

This requires https://github.com/rpm-software-management/distribution-gpg-keys/pull/128.

PR #1535, opened 2025-02-05
ci: do not swallow log output from test-suite

praiskup:
Relates: #1533


Conan-Kudo (approved):

Conan-Kudo:
Do we want this to not have a history fragment?

praiskup:
> Do we want this to not have a history fragment?

I don't think it is useful info for end-users.  I added the needed tag to silence CI.

The Fedora 42 build failure is expected, and can be ignored.

PR #1548, opened 2025-02-15
More strict hermetic image download

praiskup:
hermetic: more image-pedantic work with bootstrap
    
    When producing the buildroot lockfile, include additional bootstrap
    image metadata that can be later used for a precise image pull.
    
    The mock-hermetic-repo script is also modified here to respect the
    additional metadata.  This allows us to, e.g., download bootstrap image
    of a different (cross) architecture then the platform/host architecture
    is.  In turn, the script is now fully arch-agnostic (any host arch may
    be used for downloading files from any arch specific lockfile).
    
    Closes: #1548

### DIFF
--- a/behave/steps/other.py
+++ b/behave/steps/other.py
@@ -245,7 +245,7 @@ def step_impl(context):
 
     schemafile = os.path.join(os.path.dirname(__file__), '..', '..',
                               "mock", "docs",
-                              "buildroot-lock-schema-1.0.0.json")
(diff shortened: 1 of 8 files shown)


praiskup:
@tkopecek can you please take a look?  This should allow us to have a separate Tekton Task for fetching rpms/boostrap image (one for all architectures, and can be run pod-natively).

xsuchy (approved):
LGTM

Issue #1487, opened 2024-10-16
Mock in Docker container fails: "Insufficient rights." (with --privileged/--cap-add=SYS_ADMIN permissions)


dreibh:
### Short description of the problem
Running "mock -h" inside a "fedora:40" or "fedora:latest" Docker container (with --privileged/--cap-add=SYS_ADMIN) since a few days just prints "Insufficient rights.", without any further useful information. Mock worked before, also in GitHub Actions. There is probably a recent change, or a recent configuration update of Fedora, breaking Mock in containers.

### Output of `rpm -q mock`

mock-5.9-1.fc40.noarch

### Steps to reproduce issue

1. docker run --cap-add=SYS_ADMIN --privileged --rm -it fedora:latest bash
2. dnf install -y mock
3. mock -h
The output is always: "Insufficient rights."

### Any additional notes

<details>
  <summary>
    <p>Output of `mock --debug-config`</p>
  </summary>

Insufficient rights.

</details>

dreibh:
Using the "fedora:39" container works fine, i.e. the issue is related to Fedora 40. The version of Mock in Fedora 39 is the same:
```
[root@ecb8a5f27055 /]# cat /etc/fedora-release 
Fedora release 39 (Thirty Nine)
[root@ecb8a5f27055 /]# rpm -q mock
mock-5.9-1.fc39.noarch
[root@ecb8a5f27055 /]# mock -h
usage: 
       mock [options] {--init|--clean|--scrub=[all,chroot,cache,root-cache,c-cache,yum-cache,dnf-cache,lvm,overlayfs]}

```

dreibh:
This is the end of the strace output of the Mock run on Fedora 40. It may help identifying the problem:
`newfstatat(AT_FDCWD, "/etc/login.defs", {st_mode=S_IFREG|0644, st_size=8888, ...}, AT_SYMLINK_NOFOLLOW) = 0
openat(AT_FDCWD, "/etc/login.defs", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=8888, ...}) = 0
read(3, "#\n# Please note that the paramet"..., 4096) = 4096
read(3, "ID_MIN                  1000\nUID"..., 4096) = 4096
brk(0x597831d81000)                     = 0x597831d81000
read(3, " line length in the\n# group file"..., 4096) = 696
read(3, "", 4096)                       = 0
close(3)                                = 0
[...]

PR #1555, opened 2025-02-20
CI: add docker build

junghans:
Test for #1487

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,0 +1,16 @@
+FROM registry.fedoraproject.org/fedora:rawhide
+
+RUN dnf -y install fedpkg tito argparse-manpage bash-completion python3-devel python3-jsonschema python3-pytest
+RUN useradd -m -G mock -u 1001 user
+WORKDIR /home/user
+
+COPY mock/ /home/user/mock
+RUN chown -R user:user /home/user/mock
+USER user
+WORKDIR /home/user/mock/mock-core-configs
+RUN tito build --rpm
+WORKDIR /home/user/mock/mock
+RUN tito build --test --rpm
+USER root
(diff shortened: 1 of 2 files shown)


junghans:
Surprisingly on the recent `main` the issue seems to be fixed.

praiskup:
Thank you for working on this, @junghans :rocket:

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Build new image with mock in it and Load it into Docker for testing
+        uses: docker/build-push-action@v6
+        with:
+          tags: mockenv:latest
+          file: mock/.github/workflows/Dockerfile
+          context: .
+          pull: true
+          push: false
+          load: true
+      - run: docker run --privileged --cap-add=SYS_ADMIN -t --rm mockenv:latest sh -c "mock -h"
```
praiskup, line 26: Is it intentional to just run `-h`?  I'd expect something like `--shell echo hello world`, which would imply that buildroot and bootstrap chroots are properly initialized and used.

junghans: Good point, added!

.github/workflows/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN useradd -m -G mock -u 1001 user
+WORKDIR /home/user
+
+COPY mock/ /home/user/mock
+RUN chown -R user:user /home/user/mock
+USER user
+WORKDIR /home/user/mock/mock-core-configs
+RUN tito build --rpm
+WORKDIR /home/user/mock/mock
+RUN tito build --rpm
```
praiskup, line 13: Please use `tito build --test --rpm` here, otherwise you build an old (the latest "released") variant of the package.

praiskup:
Otherwise LGTM

praiskup:
Thank you!

PR #1556, opened 2025-02-27
Explain why Mock exists

### LINKED ISSUE #1481: Document the reasons for using Mock / key features
We frequently explain why plain "rpmbuild call in podman/buildah" isn't a replacement for Mock.  It would be nice to have documentation documenting the key benefits of Mock from this perspective (with some technical details).

/note for myself: ADR

FrostyX:
Fix #1481

### DIFF
--- a/docs/Why-Mock.md
+++ b/docs/Why-Mock.md
@@ -1,0 +1,138 @@
+# Why Mock?
+
+This document tries to answer two questions:
+
+1. Why you should use Mock instead of lower-level tooling like `rpmbuild`
+2. Why you should not try to reimplement Mock and rather join the project
+
+
+## Mock builds packages in a dedicated environment
+
+Builds are done in a dedicated buildroot that contains only a minimal set of
+packages. As a consequence, build failures are deterministic and reproducible
+because they cannot be affected by something that user installed/configured on
+their host system.
+
+Additionally, building RPM packages is inherently dangerous because specfiles
+are written in a Turing-complete language. Building an untrusted package can
+easily lead to a compromised system. This is also true for self-written
+packages. One small bug like `rpm -rf %{?nonexisting}/` can wipe your entire
+system if you are unwise enough to run `rpmbuild` as `root`. Or at least your
+user home directory.
+
+Mock does many parts of the build process inside a `systemd-nspawn`
+container (or at least [chroot][chroot]) to avoid such disasters as much as
+possible.
+
+For fully offline builds, Mock suports [hermetic builds][hermetic-builds].
+
+## Mock can build packages for different distributions
+
+Mock comes with the `mock-core-configs` package which provides a list
+[configuration files for many RPM-based distributions][mock-core-configs].
+Mock allows building a package for any of them, regardless of what distribution
+is installed on your host system.
+
+Users can easily create and distribute
+[their own configuration files][mock-configuration].
+
+Let's say you use Fedora 41 as your operating system and want to build a package
+for RHEL 8 (or any other distribution). Mock then reads the configuration file
+to know what repositories should use, what macros to define, what packages
+to install into the minimal buildroot, etc. It does all that, and then builds
+the package inside this chroot.
+
+
+## Mock installs build-time dependencies
+
+Unlike `rpmbuild`, Mock automatically installs all `BuildRequires` dependencies
+needed by the package. It also supports [dynamic dependencies][DynamicBuildRequires]
+through `%generate_buildrequires`, which is the standard way of dealing with
+dependencies for Python, Go, Rust, and other languages.
+
+Additionally, as an opt-in, Mock can install
+[dependencies from non-RPM sources][external-dependencies] like PyPI, RubyGems,
+etc.
+
+
+## Mock is not constrained by the host system
+
+You can easily prepare minimal buildroot with something like this:
+
+```
+dnf --installroot /tmp/buildroot --use-host-config install @buildsys-build
+```
+
+But what if, instead of using your system configuration, you want to use
+repositories for a different distribution. In some cases, it works just fine.
+
+But if the distribution is older (e.g. RHEL) than your host system, it may use a
+different package manager (e.g. Yum) which can install a different set of
+packages, the metadata can use different hashing algorithms, etc.
+
+Conversely, if the distribution is newer (e.g. Fedora Rawhide) it may require
+some RPM features or macros, that are not yet available on the host system.
+
+Mock solves this by using a [bootstrap chroot][bootstrap-chroot].
+
+This feature unlocked a lot of innovation in RPM because the development is not
+limited by the OS used in the builder infrastructure anymore.
+
+
+## Mock prepares a directory structure for rpmbuild
+
+Before `rpmbuild` can be used, it requires a specific directory structure to be
+created. Users typically do it by running `rpmdev-setuptree` on their systems,
+but inside of a chroot, the structure needs to be created manually with correct
+location and permissions.
+
+
+## Mock elevates permissions responsibly
+
+Mock does a lot of things to prepare a buildroot. For some of them
+(e.g. installing build dependencies) Mock needs to elevate permissions to
+root. But it does so only for a limited time-frame, and drops back to normal
+user as soon as possible.
+
+
+## Mock sets build-specific macros
+
+Some distributions or architectures require specific macros to be set, so that
+RPM packages are built correctly. For example, some chroots set `%_host_cpu` to
+`ppc64le`, some chroots adjust the `%dist` macro, etc. The `mock-core-configs`
+package provides reasonable defaults, but on top of that, users define or change
+any macro they want.
+
+The same goes for DNF/Yum placeholders like `$releasever`, `$arch`, etc.
+
+
+## Mock prepares special devices
+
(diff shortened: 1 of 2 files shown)


FrostyX:
I think this is a decent draft, but I am not knowledgeable to explain some of the sections properly. If you think something is not quite true, or you know some good examples to illustrate the points, please let me know :-)

Also, did I miss some key reasons?

praiskup:
@Conan-Kudo any remarks?

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+needed by the package. It also supports [dynamic dependencies][DynamicBuildRequires]
+through `%generate_buildrequires`, which is the standard way of dealing with
+dependencies for Python, Go, Rust, and other languages.
+
+Additionally, not enabled by default, Mock can install
+[dependencies from non-RPM sources][external-dependencies] like PyPI, RubyGems,
+etc.
+
+
+## Mock is not bounded by the host system
```
Conan-Kudo, line 43: Constrained is probably a better word

Conan-Kudo (changes requested):

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
+# Why Mock?
+
+This document tries to answer two questions:
+
+1. Why you should use Mock instead of lower-level tooling like `rpmbuild`
+2. Why you should not try to reimplement Mock and rather join the project
+
+
+## Mock builds packages inside an isolated environment
```
praiskup, line 9: I'd prefer calling it "separated" rather than "isolated" because Mock isn't in charge of isolation.

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+Mock builds packages inside of `systemd-nspawn` containers to prevent such
+disasters from happening.
+
+
+## Mock can build packages for different distributions
+
+Mock comes with the `mock-core-configs` package which provides a list
+[configuration files for many RPM-based distributions][mock-core-configs].
+Mock allows building a package for any of them, regardless of what distribution
+is installed on your host system.
```
praiskup, line 26: Can you emphasize how mock is doing that, e.g. that on Fedora 42 machine we install Mageia build chroot, and we build "inside" that chroot?  This is probably the #1 reason.

This is probably the #1 reason to use Mock.

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+
+Users can easily create and distribute their own configuration files.
```
praiskup, line 28: Please link with mock-configs docs.

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## Mock installs build-time dependencies
+
+Unlike `rpmbuild`, Mock automatically installs all `BuildRequires` dependencies
+needed by the package. It also supports [dynamic dependencies][DynamicBuildRequires]
+through `%generate_buildrequires`, which is the standard way of dealing with
+dependencies for Python, Go, Rust, and other languages.
+
+Additionally, not enabled by default, Mock can install
```
praiskup, line 38: s/not enabled by default/as an opt-in/

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+But what if, instead of using your system configuration, you want to use
+repositories for a different distribution. In some cases, it works just fine.
+
+But if the distribution is older (e.g. RHEL) than your host system, it may use a
+different package manager (e.g. Yum) which can install a different set of
+packages, the metadata can use different hashing algorithms, etc.
+
+Conversely, if the distribution is newer (e.g. Fedora rawhide) it may require
```
praiskup, line 58: s/rawhide/Rawhide/

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+This feature unlocked a lot of innovation in RPM because the development is not
+limited by the OS used in the builder infrastructure anymore.
+
+
+## Mock prepares a directory structure for rpmbuild
+
+Before `rpmbuild` can be used, it requires a specific directory structure to be
+created. Users typically do it by running `rpmdev-setuptree` on their systems,
+but inside of a chroot, the structure needs to be created manually with correct
+location and permissions.
```
praiskup, line 72: Additional info (related to this one, maybe deserves its own paragraph) is that Mock automatically elevates permissions to root, but only for a limitted time-frame (when some package (e.g. BuidlRequires) need to be installed, or if some special directories or even users/groups need to be pre-created).

This way, building Mock from sources (specfile or source RPM) is a matter of running a single command (build deps are installed and rpmbuild executed fully automatically).

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## Mock sets build-specific macros
+
+Some distributions or architectures require specific macros to be set, so that
+RPM packages are built correctly. For example, some chroots set `%_host_cpu` to
+`ppc64le`, some chroots adjust the `%dist` macro, etc. The `mock-core-configs`
+package provides reasonable defaults, but on top of that, users define or change
+any macro they want.
```
praiskup, line 81: Similar with DNF/Yum placeholders like `$releasever` or `$arch`.

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+Some packages require special devices like `/dev/fuse` `/dev/loop-control`,
+`/dev/hwrng`, `/dev/prandom`, etc to build successfully. Mock makes them
+available inside of the buildroot.
+
+
+## Mock caches downloaded RPM packages
+
+Not all dependencies have to be downloaded every time. This brings significant
+performance improvements especially when consecutively rebuilding the same RPM
+package or a package from the same ecosystem (e.g. multiple Python packages).
```
praiskup, line 95: Mock also caches DNF metadata (often tens of MBs).

praiskup:
Awesome start, @FrostyX, thank you for moving this forward.  My notes here are mostly "wishes".  We can fix them now or anytime later.

docs/Why-Mock.md
Conan-Kudo: It's not separate either, because we don't use VMs to create separation (we have had plenty of bugs related to leakage from the host). So it's probably better to just say "dedicated environment".

FrostyX:
Thank you for the review @praiskup and @Conan-Kudo, please don't merge yet, I will address all the points soon.

docs/Why-Mock.md
```diff
@@ -0,0 +1,107 @@
...
+## Mock builds packages inside an isolated environment
+
+Building RPM packages is inherently dangerous because specfiles are written in a
+Turing-complete language. Building an untrusted package can easily lead to a
+compromised system. This is also true for self-written packages. One small bug
+like `rpm -rf %{?nonexisting}/` can wipe your entire system if you are unwise
+enough to run `rpmbuild` as `root`. Or at least your user home directory.
+
+Mock builds packages inside of `systemd-nspawn` containers to prevent such
+disasters from happening.
```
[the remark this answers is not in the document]
praiskup: I'd be still a bit more picky about the words here; first, systemd-nspawn is not always used (even if --isolation=nspawn, e.g. when installing packages into bootstrap).  Then, I'm not 100% sure that things like `Version: 10%(rm -rf /*)` are safe with Mock (I think that Mock still parses spec files on host in some cases).

I'd slightly modify this statement to:
```
Mock e.g. does many parts of the build process inside `systemd-nspawn` containers (or at least chroot(2)) to avoid such disasters as much as possible.

File: docs/Why-Mock.md [hunk 11]
```diff
@@ -0,0 +1,138 @@
...
+easily lead to a compromised system. This is also true for self-written
+packages. One small bug like `rpm -rf %{?nonexisting}/` can wipe your entire
+system if you are unwise enough to run `rpmbuild` as `root`. Or at least your
+user home directory.
+
+Mock builds packages inside of `systemd-nspawn` containers to prevent such
+disasters from happening.
+
+For truely isolated, offline builds, Mock also supports
+[hermetic builds][hermetic-builds].
```
praiskup, line 27: While hermetic allows 100% offline builds, it doesn't isolate from the host.  The hermetic mode page documents this well, but maybe we can elaborate a bit even here...

Basically, anyone who wants to be 100% safe and isolated should go with `podman run --rm -ti --privileged` wrapper.

Conan-Kudo: We should not call it hermetic. It's just a reproducible build mode, not a hermetic build mode.

praiskup:
Thank you very much for the update!  I still think the part describing "isolation" perspective deserves some care.  But if you can't spend more time on this, we can merge and fix in subsequent PR.

docs/Why-Mock.md
praiskup: Two most difficult problems in SW industry: Naming & Caching. The chosen "feature name" has been "hermetic" -- mostly to align with the SLSA definition https://slsa.dev/spec/v0.1/requirements#hermetic ... I am not against changing the feature name long term, but that would be a different PR than this one.

FrostyX:
Updated and squashed

praiskup (approved):

PR #1558, opened 2025-03-05
podman: bail clearly when image pull fails

AdamWill:
We retry the image pull with a `@backoff.on_predicate` wrapper, but it can still fail entirely. When that happens, `image_id` remains as `None` and this causes us to hit a subprocess exception in `tag_image`. Because the buildroot `_fallback` decorator only handles `PodmanError` exceptions, we blow up on that exception instead of falling back smoothly to chroot.

This solves that by giving the backoff wrapper an `on_giveup` handler that raises `PodmanError`, so the fallback mechanism should kick in.

### DIFF
--- a/mock/py/mockbuild/podman.py
+++ b/mock/py/mockbuild/podman.py
@@ -91,6 +91,13 @@ def podman_check_native_image_architecture(image, logger=None, podman_binary=Non
     return True
 
 
+def pull_fail_handler(details):
+    """
+    Raise an error when image pull fails, because lambdas can't raise.
+    """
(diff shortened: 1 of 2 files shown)


AdamWill:
This was causing failures in openQA, e.g. https://openqa.fedoraproject.org/tests/3272055# . I don't have the text log because of how openQA works and because the test isn't set up to upload mock's logs in this case, but from the traceback visible in the screenshot and examining the code it seems pretty clear what happened.

praiskup (approved):
Thank you!

Issue #1583, opened 2025-05-24
mockbuild.util shouldn't use pretty_getcwd() to set ORIGINAL_CWD on import


ferdnyc:
The `mockbuild.util` module contains code that sets the global variable `ORIGINAL_CWD` as an import side-effect:

https://github.com/rpm-software-management/mock/blob/4258bcc902be6aaac5f71cb3d715a59d443d62f7/mock/py/mockbuild/util.py#L898-L899

The problem is, `pretty_getcwd` is a `@traceLog()`-decorated function, which can cause issues in unexpected contexts, and really shouldn't be executed without intention (such as when importing the module).

One manifestation of this can be a nasty traceback when `pydoc3` scans `FedoraReview` (which imports `mockbuild.config` in its own `FedoraReview.mock` module). On my Fedora 42 system, using the system python / site-packages:

```console
$ pydoc3 -k Review

pitivi.mediafilespreviewer 
pitivi.timeline.previewers - Previewers for the timeline.
pyexiv2.preview - Provide the Preview class.
pdfviewer using PyMuPDF (GPL)
/usr/lib64/python3.13/site-packages/wx/lib/pubsub/__init__.py:23: wxPyDeprecationWarning: wx.lib.pubsub has been deprecated, please migrate your code to use pypubsub, available on PyPI.
  warnings.warn('wx.lib.pubsub has been deprecated, please migrate your '
FedoraReview - Tools for helping Fedora package reviewers
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib64/python3.13/logging/__init__.py", line 1150, in emit
    msg = self.format(record)
  File "/usr/lib64/python3.13/logging/__init__.py", line 998, in format
    return fmt.format(record)
           ~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/logging/__init__.py", line 711, in format
    record.message = record.getMessage()
                     ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.13/logging/__init__.py", line 400, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
TypeError: not enough arguments for format string
Call stack:
  File "/home/ferd/.local/bin/pydoc3", line 5, in <module>
    pydoc.cli()
  File "/usr/lib64/python3.13/pydoc.py", line 2778, in cli
    apropos(val)
  File "/usr/lib64/python3.13/pydoc.py", line 2288, in apropos
[...]

PR #1584, opened 2025-05-24
Remove traceLog decorator from pretty_getcwd

ferdnyc:
`pretty_getcwd()` is called during module import (to set `ORIGINAL_CWD`), and using a traced function creates unwanted side effects.

Fixes: #1583

I think this can be `label no-release-notes`, let me know if you disagree.


xsuchy:
I think removing the tracelog decorator will be simpler fix. We can live without it.

praiskup (approved):

xsuchy:
the TF failure is not relevant to this PR. So merging.

Issue #1594, opened 2025-06-16
Mock on Rawhide fails to start under non-root user


praiskup:
```
$ mock --shell

ERROR: [Errno 1] Operation not permitted

ERROR: The most common cause for this error is trying to run /usr/libexec/mock/mock as an unprivileged user.
ERROR: You should not run /usr/libexec/mock/mock directly.
```

https://bugzilla.redhat.com/show_bug.cgi?id=2372865

praiskup:
Python ` python3-3.14.0~b2-3.fc43.x86_64` doesn't seem to allow us to drop root permissions here:
```
180             with ProcessPoolExecutor(max_workers=1) as executor:
181  ->             future = executor.submit(self.drop_privs_forever_and_execute,
182                                          method, *args, **kwargs)
```

PR #1595, opened 2025-06-16
ProcessPoolExecutor changed behaviour with Python 3.14

xsuchy:
Fixes: #1594
RHBZ#2372865

Previously, the default method of starting a process was fork. With Python3.14 it is no longer true (I think that default is spawner). But the new default does not work with Mock code and causes traceback:

```
  File "/usr/libexec/mock/mock", line 1140, in <module>
    main()
    ~~~~^^
  File "/usr/lib/python3.14/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/libexec/mock/mock", line 718, in main
    config_opts = uidManager.run_in_subprocess_without_privileges(
            config.load_config, config_path, options.chroot)
  File "/usr/lib/python3.14/site-packages/mockbuild/uid.py", line 183, in run_in_subprocess_without_privileges
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.14/concurrent/futures/_base.py", line 450, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.14/concurrent/futures/_base.py", line 395, in __get_result
    raise self._exception
PermissionError: [Errno 1] Operation not permitted
```

This change explicitely set the start method to "fork".

### DIFF
--- a/mock/py/mockbuild/uid.py
+++ b/mock/py/mockbuild/uid.py
@@ -8,8 +8,10 @@ import atexit
 import ctypes
 import errno
 import grp
+import multiprocessing
 import os
 import pwd
+import sys
 from concurrent.futures import ProcessPoolExecutor
 from contextlib import contextmanager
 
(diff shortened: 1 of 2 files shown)


praiskup:
Thank you!

PR #1599, opened 2025-06-17
Re-introduce live logs in TF

praiskup:


praiskup:
Seems fine, logs appear to be live again, so `pty: true` probably worked.  It still feels I'm facing some issues with caches (i.e. lines are not appearing as fast as I'd wish) but it generally, the thing looks good now.  Thanks @thrix.

Issue #1588, opened 2025-05-30
The group installation doesn't work?


praiskup:
I just took a fresh vm (1minutetip fedora41, internal tool..), did `dnf install -y mock`, and seen this:

```
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                                                                                                                                                                                                                                                                            
```

The ownership of the mock package is wrong:

```
# rpm -V mock
......G..    /usr/share/bash-completion/completions/mock
......G..    /usr/share/bash-completion/completions/mock-parse-buildlog
......G..    /usr/share/doc/mock
......G..  d /usr/share/doc/mock/buildroot-lock-schema-1.0.0.json
......G..  d /usr/share/doc/mock/buildroot-lock-schema-1.1.0.json
......G..  d /usr/share/doc/mock/site-defaults.cfg
......G..    /var/cache/mock
......G..    /var/lib/mock
```

Probably some ordering problem?  `dnf reinstall mock` helps.

PR #1604, opened 2025-06-18
configs: drop `mock` ownership from config files

praiskup:
The `mock` group ownership seems to be an old typo in the commit 588ef71feeb2393ac64b506b2555574c2c6a4d30 that was just trying to fix the file mods (per %changelog).  We don't need %defattr these days either.

Fixes: #1588


mock/mock.spec
```diff
@@ -329,6 +322,12 @@ pylint-3 py/mockbuild/ py/*.py py/mockbuild/plugins/* || :
 %dir  %{_datadir}/cheat
 %config(noreplace) %{_sysusersdir}/mock.conf
 
+# cache & build dirs, writeable by mock group
+%defattr(0775, root, mock, 0775)
+%dir %{_localstatedir}/cache/mock
+%dir %{_localstatedir}/lib/mock
```
praiskup, line 328: Nah?  RPM is still not able to create the group before installing those files:
```
[11/85] Installing mock-filesystem-0:6.2.test-1.20250618070049420054.pr1604.17.gee0a0816.fc41.noarch                                                                                                100% |   1.2 MiB/s |  19.1 KiB |  00m00s
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                            
>>> [RPM] group mock does not exist - using root                                                                                                                                                                                            
[12/85] Installing librdmacm-0:51.0-5.fc41.x86_64     
...
>>> Finished trigger-install scriptlet: systemd-0:256.15-1.fc41.x86_64                                                                                                                                                                      
>>> Scriptlet output:                                                                                                                                                                                                                       
>>> Creating group 'mock' with GID 135.    
```

praiskup: this only works on F42+ :-(

FrostyX (approved):

PR #1606, opened 2025-06-20
Fix unclosed file warning

pastalian:
No functional changes - just silencing a ResourceWarning by ensuring the file handle is explicitly closed.
ResourceWarning is typically ignored by default but some distributions enable it.

```
$ python -m pytest -Wdefault tests
...
mock/tests/test_config_loader.py::TestConfigLoader::test_config_paths
  /home/me/mock/mock/py/mockbuild/config.py:722: ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/me/mock/mock/tests/data/config-001/site-defaults.cfg' mode='r' encoding='UTF-8'>
    content = open(config_file).read()
```

> not deserve a release note.

I don't think this fix needs a release note.

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -719,7 +719,8 @@ def include(config_file, search_path, paths):
             return ""
 
         paths.add(config_file)
-        content = open(config_file).read()
+        with open(config_file) as f:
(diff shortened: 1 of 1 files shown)


praiskup:
thank you

Issue #1612, opened 2025-07-16
Mock should use systemd-nspawn --suppress-sync=true instead of LD_PRELOADing the nosync library when possible.


Spudd86:
Since systemd 250 systemd-nspawn has supported the --suppress-sync option. 

> `--suppress-sync=`
> Expects a boolean argument. If true, turns off any form of on-disk file system synchronization for the container payload. This means all system calls such as [sync(2)](https://man7.org/linux/man-pages/man2/sync.2.html), fsync(), syncfs(), … will execute no operation, and the O_SYNC/O_DSYNC flags to [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) and related calls will be made unavailable. This is potentially dangerous, as assumed data integrity guarantees to the container payload are not actually enforced (i.e. data assumed to have been written to disk might be lost if the system is shut down abnormally). However, this can dramatically improve container runtime performance – as long as these guarantees are not required or desirable, for example because any data written by the container is of temporary, redundant nature, or just an intermediary artifact that will be further processed and finalized by a later step in a pipeline. Defaults to false.

This would entirely sidestep the few issues the current approach of using the nosync library has, particularly the fact that it fails to load correctly ( #846 ), and has issues if the host doesn't have multilib.

xsuchy:
Interesting.
Just did quick check:
* rhel8 has systemd 239
* rhel9 has 252

PR #1613, opened 2025-07-17
use --suppress-sync=yes in systemd-nspawn when available

xsuchy:
Fixes: #1612

# Add release note. See https://rpm-software-management.github.io/mock/Release-Notes-New-Entry
# or let us know to add `label no-release-notes` if you think the change does
# not deserve a release note.


xsuchy:
Hmm there is some issue with networking in Copr:
```
Error: Invalid value for '[PATH_OR_URL]': Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v --tags -- https://github.com/rpm-software-management/mock /tmp/tmph7_7am9w
  stderr: 'Cloning into '/tmp/tmph7_7am9w'...
fatal: unable to access 'https://github.com/rpm-software-management/mock/': Failed to connect to github.com port 443 after 136237 ms: Could not connect to server
```

mock/py/mockbuild/util.py
```diff
@@ -775,6 +783,9 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None,
     if _check_nspawn_resolv_conf():
         nspawn_argv.append("--resolv-conf=off")
 
+    if check_nspawn_has_suppress_sync_option():
+        nspawn_argv += ['--suppress-sync=yes']
```
praiskup: Is nosync something to be set by default? Even Fedora Koji is using nspawn nowadays, so the change affects mostly every build out there (not konflux, no nspawn there).

releng/release-notes-next/use-suppress-sync.feature
```diff
@@ -0,0 +1,3 @@
+Mock now pass --suppress-sync=yes to every systemd-nspawn call (when available
+i.e. on RHEL9+). This turns off any form of on-disk file system synchronization
+for the container payload. This dramatically improve container runtime performance.
```
praiskup: improves

mock/py/mockbuild/util.py
xsuchy: I would say yes. It should have no negative effects (like ld preload).

praiskup: I just needed to say this aloud, I'd prefer (but world is not ideal) having an option (something like `config_opts["no_fsync"] = True`), clearly documenting we do our best to turn the disk syncs off.

There are these things that make me nervous a bit: 
- either nosync was never "on by default", just opt-in, some people might not want to do the risk (and some people do weird things in mock, not just RPM building)
- nosync was +- under our control (maintained by our neighborhood folks), now we rely on a feature that we don't know how works or how widely is used, or maybe we do?  Does it rely on on-host hacks, or in-chroot hacks?  If in-chroot, is it working for all chroots (works both with rawhide & epel8 glibc)?
- Copr e.g. uses tmpfs, so the benefit of the change will be minimal (if any)
- we change all builds in Koji at once (ping @nirik fyi)
- Konflux builds are not affected, because they run Mock in container (= --isolation=simple) :-(

releng/release-notes-next/use-suppress-sync.feature
xsuchy: fixed.

nirik:
Thanks for the heads up. I expect this should be fine (and make things faster), but will keep an eye out for problems.

It might be nice to have a way to change it back without downgrading mock, but that should at least work if needed.

releng/release-notes-next/use-suppress-sync.feature
```diff
@@ -0,0 +1,16 @@
...
+    config_opts['nspawn_args'] = ['--capability=cap_ipc_lock', '--suppress-sync=yes']
+
+This dramatically improves container runtime performance - up to 16 % for big packages.
+It does not have an impact on building small packages.
+
+The only drawback is that in case of a hard shutdown (power outage) during the build, some changes in the buildroot can be lost.
+As buildroot is supposed to be ephemeral and reconstructed for every build, we enabled this by default.
+If you want to disable this feature, you can put in your config:
+
+    config_opts['nspawn_args'] = ['--capability=cap_ipc_lock']
```
praiskup, line 16: Cut-and-paste error here ^

xsuchy: Discussed on mtg. No it is not an error. This line is the previous default.

praiskup:
Note #1644 - the default is being reverted.

PR #1621, opened 2025-08-27
add config for gemini reviews

xsuchy:
I want to try the Gemini review in this repo. I've got good feedback on it from the Packit team.
If no one objects, I will enable 
  https://github.com/marketplace/gemini-code-assist

I treat it as an experiment. If that makes more noise than help, we can disable it later.


xsuchy:
Where do you see that it has write access. AFAIK it should only put comments in PR.

PR #1623, opened 2025-08-28
Remove myself from the list of Oracle Linux contacts

Djelibeybi:
I am no longer an Oracle Linux product manager, therefore I am removing myself from the list of contacts for that distribution.

This should probably be tagged `label no-release-notes` IMO


Conan-Kudo (approved):

xsuchy (approved):
+1

Issue #1562, opened 2025-03-17
Detect unused BuildRequires


xsuchy:
Just a heads up that there is a new project
  https://github.com/mkoncek/unbreq
that detects unused buildrequires. So far it is functional prototype.

Once it is fully functional and mature we may link to that from our documentation. Or maybe even merge to this repo.

mkoncek:
The project just reached another milestone (replacing subprocess calls with `librpm` / `libdnf` native calls).
I added a Copr repo with the latest build: https://copr.fedorainfracloud.org/coprs/mkoncek/unbreq/.
There are still some unsolved issues:
* chroot / nspawn handling should be done using mock's internal util helpers, i tried to use them but it didn't work for me
* Proper isolation - this should be fixable when the plugin is packaged and available to install as a regular package during package builds:
  *  `fanotify` binary should be installed into the buildroot and ran via chroot / nspawn so that it's filesystem visibility is limited to only buildroot
  * `resolve` binary should be installed into the bootstrap root and ran with chroot / nspawn with bind-mounted buildroot (used as `dnf --installroot` / `rpm --root`)

PR #1630, opened 2025-09-15
Add plugin for detection of unused `BuildRequires`

mkoncek:
# Add a reference to related issue - preferably in the git commit message
Fixes: #1562

## Issues
* [x] `--isolation=simple` does not work yet though there is code for it
* [ ] Missing tests
* [x] `get_buildrequires` uses `import rpm` from the host, it should use one from the bootstrap chroot
* [ ] When the chroot is mounted on a filesystem with `relatime` option, modifies the atime / mtime of the chroot files. Inspect how this affects reproducible builds
* [ ] Add some built-in filters mostly for dynamic languages which access module/import files regardless whether they are used

### DIFF
--- a/docs/Plugin-Unbreq.md
+++ b/docs/Plugin-Unbreq.md
@@ -1,0 +1,74 @@
+---
+layout: default
+title: Plugin Unbreq
+---
+
+Detector of unused `BuildRequires` of RPM builds.
+
+# Usage
+
+> [!Note]
+> The current implementation requires that the mock chroot is **not** on a filesystem mounted with the `noatime` option.
+> You may need to remount the relevant directory with options `strictatime,lazytime`.
+
+Enable it upon `mock` execution via a flag:
+```
+mock --enable-plugin=unbreq ...
+```
+
+In logs you should see messages like:
+```
+INFO: enabled unbreq plugin(...)
+```
+
+If Unbreq detects an unneeded `BuildRequire` it prints a message like:
+```
+WARNING: unbreq plugin: the following BuildRequires were not used:
+...
+```
+
+## Configuration
+The configuration options for this plugin are listed below.
+
+Configuration via the `.cfg` file is done this way:
+```
+config_opts['plugin_conf']['unbreq_opts']['${OPTION}'] = ${VALUE}
+```
+
+Configuration from the command line is done this way:
+```
+--plugin-option='unbreq:${OPTION}=${VALUE}'
+```
+
+### Configuration options
+* **`exclude_accessed_files`** - *`List[str]`*
+  * A list of regular expressions that are used to ignore file accesses of certain files.
+
+    *Example*: `xmvn` always reads all files inside `/usr/share/maven-metadata/`.
+    The exclusion filter `^/usr/share/maven-metadata/` excludes these files from the listing.
+
+## How it works
+
+### Before the build
+The tool reads the `BuildRequires` of the SRPM files (usually only one) in their standard location.
+This step is executed after resolving dynamic `BuildRequires` and this tool is able to recognize them.
+
+It also marks a timestamp before executing the build for use as reference.
+
+Depending of the detected filesystem mount options, if `relatime` options is detected, then all the files owned by all the installed dependencies have their access times and modify times set to `0`, to make it possible to detect further file accesses during the build.
+
+### After the build
+After executing the build, the relevant files have their access time compared to the saved timestamp.
+
+Each `BuildRequire` field may cause the installation of multiple packages.
+These are obtained by iteratively executing `dnf --assumeno remove ${BuildRequires}`.
+
+If any of the files owned by any of the package in this set was accessed, then this `BuildRequires` field can not be removed.
+
+Each iteration of the `remove` query will also include all the `BuildRequires` fields for which were shown to be unnecessary.
+This is because the removal of multiple packages at once can produce a different result than removing them separately.
+
+### Output
+The output of the tool is logged in the build log at the end of the build.
+
+Available since version 6.4.
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -376,6 +376,15 @@
 #     'cmd_base': ['/usr/bin/rpmautospec', 'process-distgit'],
 # }
 
+### unbreq plugin
+# This plugin is able to detect unused BuildRequires from an RPM build based
+# on file accesses during the build. It is disabled by default.
+#config_opts['plugin_conf']['unbreq_enable'] = False
+#
+# You can add custom regular expression filters which make the tool ignore
+# certain file paths being accessed.
+#config_opts['plugin_conf']['unbreq_opts']['exclude_accessed_files'] = []
+
 #############################################################################
 #
 # environment for chroot
--- a/mock/py/mockbuild/plugins/unbreq.py
+++ b/mock/py/mockbuild/plugins/unbreq.py
@@ -1,0 +1,353 @@
+#!/usr/bin/python3
+
+"""
+A plugin which detects unused BuildRequires based on file accesses during the
+RPM build.
+
+Author: Marián Konček <mkoncek@redhat.com>
+"""
+
+# python library imports
+import subprocess
+import os
+import re
+from typing import Generator
+from contextlib import contextmanager
+
+# our imports
+from mockbuild.trace_decorator import getLog, traceLog
+import mockbuild.util
+from mockbuild.util import USE_NSPAWN
+import mockbuild.mounts
+import mockbuild.file_util
+
+requires_api_version = "1.1"
+
+class AtimeDict(dict):
+    """
+    A caching lazy dictionary mapping file paths to their access time.
+    """
+
+    def __missing__(self, key: str) -> float:
+        result = os.stat(key).st_atime
+        self[key] = result
+        return result
+
+@traceLog()
+def init(plugins, conf, buildroot) -> None:
+    """
+    Plugin entry point.
+    """
+    Unbreq(plugins, conf, buildroot)
+
+class Unbreq:
+    """
+    Mock plugin that detects unused BuildRequires in RPM builds.
+
+    Works by tracking file access times during the build process to determine
+    which BuildRequire packages had their files accessed. Reports any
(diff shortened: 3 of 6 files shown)


docs/Plugin-Unbreq.md
```diff
@@ -0,0 +1,60 @@
...
+    
+    *Example*: `xmvn` always reads all files inside `/usr/share/maven-metadata/`.
+    The exclusion filter `^/usr/share/maven-metadata/` excludes these files from the listing.
+
+## Issues
+* Currently does not work with `--isolation=simple`
+
+## How it works
+The tool marks a timestamp before executing the build.
+After executing the build, relevant files have their access time compared to the saved timestamp.
```
xsuchy: the relevant

docs/Plugin-Unbreq.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+Configuration from the command line is done this way:
+```
+--plugin-option='unbreq:${OPTION}=${VALUE}'
+```
+
+### Configuration options
+* **`exclude_accessed_files`** - *`List[str]`*
+  * A list of regular expressions which are used to ignore file accesses of certain files.
```
xsuchy: expressions that are used

docs/Plugin-Unbreq.md
```diff
@@ -0,0 +1,60 @@
...
+* Currently does not work with `--isolation=simple`
+
+## How it works
+The tool marks a timestamp before executing the build.
+After executing the build, relevant files have their access time compared to the saved timestamp.
+
+1. The tool runs `dnf --assumeno remove ${BuildRequire}` for each field in the SRPM file to get the list of RPMs that would be removed along with the `BuildRequire`.
+2. Then the tool checks the files owned by all the RPMs, if they were accessed during the build.
+3. If no, then the `BuildRequire` is added to the list of removable fields.
+   In the next iteration the `dnf` query is executed with the next `BuildRequire` field together with all the fields which can be removed.
```
xsuchy: fields that can be removed

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/python3
+# python library imports
+import subprocess
+import os
+import re
+from typing import *
```
xsuchy: Personal opinion here: I dislike these asterisk imports. Can you enumerate what you use?

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -0,0 +1,226 @@
...
+                process.args, process.returncode, process.stderr.rstrip()
+            ))
+        result = []
+        for line in process.stdout.splitlines():
+            if not line.startswith(" "):
+                continue
+            nvr = line.split()
+            if len(nvr) != 6:
+                continue
+            result.append(nvr[0] + "-" + nvr[2] + "." + nvr[1])
```
xsuchy: Can you join this rather using format() or f-string?

docs/Plugin-Unbreq.md
```diff
@@ -0,0 +1,60 @@
...
+* Currently does not work with `--isolation=simple`
+
+## How it works
+The tool marks a timestamp before executing the build.
+After executing the build, the relevant files have their access time compared to the saved timestamp.
+
+1. The tool runs `dnf --assumeno remove ${BuildRequire}` for each field in the SRPM file to get the list of RPMs that would be removed along with the `BuildRequire`.
+2. Then the tool checks the files owned by all the RPMs, if they were accessed during the build.
+3. If no, then the `BuildRequire` is added to the list of removable fields.
+   In the next iteration the `dnf` query is executed with the next `BuildRequire` field together with all the fields that can be removed.
```
xsuchy: Can you add to the bottom of this page the note: Available since version 6.4

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -0,0 +1,278 @@
...
+        buildrequires = set()
+        for srpm in os.scandir(self.srpm_dir):
+            for br in self.do_with_chroot(lambda: self.get_buildrequires(srpm.path)):
+                buildrequires.add(br)
+        self.buildrequires_providers = self.get_buildrequires_providers(sorted(buildrequires))
+
+        # NOTE maybe find a better example file to touch to get an atime?
+        path = os.path.join(self.buildroot.rootdir, "dev", "null")
+        subprocess.run(["touch", path], check = True)
+        self.min_time = os.stat(path).st_atime
```
xsuchy: you can use mockbuild.file_util.touch(fileName)
you can use os.path.getatime(path).
Isn't this equal to time.gmtime()?

mkoncek: Seems like both `os.stat(path).st_atime` and `os.path.getatime(path)` return the same value (float in seconds since the Unix epoch). But `os.path.getatime` sounds better.
`time.gmtime` should not be needed anywhere.

docs/Plugin-Unbreq.md
```diff
@@ -0,0 +1,64 @@
...
+    The exclusion filter `^/usr/share/maven-metadata/` excludes these files from the listing.
+
+## Issues
+* Currently does not work with `--isolation=simple`
+
+## How it works
+The tool marks a timestamp before executing the build.
+After executing the build, the relevant files have their access time compared to the saved timestamp.
+
+1. The tool runs `dnf --assumeno remove ${BuildRequire}` for each field in the SRPM file to get the list of RPMs that would be removed along with the `BuildRequire`.
```
praiskup: Why we are doing this, considering that we do step 2.?  Can you maybe elaborate a bit on "why"?

Can you document the perspective WRT dynamic build dependencies?  Are you analyzing buildroot after dynamic build deps are installed?

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -0,0 +1,282 @@
...
+        result = os.stat(key).st_atime
+        self[key] = result
+        return result
+
+# plugin entry point
+@traceLog()
+def init(plugins, conf, buildroot):
+    Unbreq(plugins, conf, buildroot)
+
+class Unbreq():
```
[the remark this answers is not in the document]
praiskup: the `()` part is not needed

releng/release-notes-next/detect-unused-buildrequires.feature.md
```diff
@@ -0,0 +1,23 @@
...
+```
+INFO: unbreq plugin: BuildRequire asciidoc is needed because file /etc/profile.d/gnupg2.sh was accessed
+INFO: unbreq plugin: BuildRequire bash is needed because file /usr/bin/bash was accessed
+INFO: unbreq plugin: BuildRequire diffutils is needed because file /usr/bin/cmp was accessed
+INFO: unbreq plugin: BuildRequire gcc-c++ is needed because file /usr/bin/c++ was accessed
+INFO: unbreq plugin: BuildRequire make is needed because file /usr/bin/gmake was accessed
+INFO: unbreq plugin: BuildRequire rubygem-asciidoctor is needed because file /usr/bin/asciidoctor was accessed
+WARNING: unbreq plugin: the following BuildRequires were not used:
+        xmlto
+        xxd > 1
```
praiskup: Very nice!

docs/Plugin-Unbreq.md
```diff
@@ -0,0 +1,64 @@
...
+title: Plugin Unbreq
+---
+
+Detector of unused `BuildRequires` of RPM builds.
+
+# Usage
+
+> [!Note]
+> The current implementation requires that the mock chroot is **not** on a filesystem mounted with the `noatime` option.
+> You may need to remount the relevant directory with options `strictatime,lazytime`.
```
praiskup, line 12: Are you detecting these situations and failing hard?

mkoncek: We are detecting them, but failing "soft": https://github.com/mkoncek/mock/blob/unbreq/mock/py/mockbuild/plugins/unbreq.py#L304-L309

docs/Plugin-Unbreq.md
mkoncek: Step 2 works with the results of step 1, maybe it is not clearly written.

I have just tried this plugin with a package which uses dynamic buildrequires and it does not work well; the prebuild hook is called before the buildrequires are installed.

mkoncek:
>Would you mind squashing those commits into one?

I believe there is a button somewhere in the Github interface that does that when accepting a Pull Request...

docs/Plugin-Unbreq.md
mkoncek: @praiskup seems like i want `postdeps` hook and not `prebuild`?

praiskup: I think the part `owned by all the RPMs` confused me.  It's just the set of transitive deps from a specific `BuildRequires` entry.  Maybe change the sentence to `owned by the list of RPMs`?

`postdeps` is called later than `prebuild`, might be a good idea if you want to handle dynamic deps properly.

mock/docs/site-defaults.cfg
```diff
@@ -376,6 +376,12 @@
 #     'cmd_base': ['/usr/bin/rpmautospec', 'process-distgit'],
 # }
 
+### unbreq plugin
+# This plugin is able to detect unused BuildRequires from an RPM build based
+# on file accesses during the build. It is disabled by default.
+# config_opts['plugin_conf']['unbreq_enable'] = False
+# config_opts['plugin_conf']['unbreq_opts']['exclude_accessed_files'] = []
```
praiskup: Slight preference to drop whitespace between `#` and the option;  this way you can easily tell the difference between real config option and the documentation text.

mkoncek: I agree.

praiskup:
Thanks, LGTM. The postdeps vs. prebuild thing can be solved later. While some of the dynamically generated deps could also be detected as useless, the process of calculation them may be the only part of the rpmbuild that is actually using the static BuildRequires.  Quite likely the plugin needs a callback in both hooks - good material for a follow up PR.

praiskup:
I'd like to wrap a new release today (#1633).

mkoncek:
I will swuash the commits then.

PR #1631, opened 2025-09-23
feat: add support for client certificates

scoheb:
- in some rare cases, a mock config can include a yum repo config section that include a ssl cert and key that are used for authentication by dnf
- when running mock-hermetic-repo, the same client cert is needed to download the rpms.
- this commit adds support for the following:

```
--client-cert-for example.com /path/crt.pem /path/key.pem
```

Assisted-by: Cursor

Fixes: #NUMBER

> not deserve a release note.

### DIFF
--- a/mock/py/mock-hermetic-repo.py
+++ b/mock/py/mock-hermetic-repo.py
@@ -16,7 +16,7 @@ import shutil
 import subprocess
 import sys
 
-from urllib.parse import quote
+from urllib.parse import quote, urlparse
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
@@ -46,7 +46,7 @@ def request_with_retry(retries=5, backoff_factor=0.3,
     return session
 
 
-def download_file(url, outputdir):
+def download_file(url, outputdir, client_auth_map):
     """
     Download a single file (pool worker)
     """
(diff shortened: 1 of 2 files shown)


mock/py/mock-hermetic-repo.py
```diff
@@ -59,7 +59,26 @@ def download_file(url, outputdir):
...
+
+        session = request_with_retry()
+        if cert_arg is not None:
+            with session.get(url, stream=True, timeout=60, cert=cert_arg) as response:
+                if response.status_code != 200:
+                    return False
+                with open(file_name, "wb") as fd:
+                    shutil.copyfileobj(response.raw, fd)
+                return True
+        with session.get(url, stream=True, timeout=60) as response:
```
praiskup: Can we have just one `get()` call, and one handling of status_code != 200?

mock/py/mock-hermetic-repo.py
```diff
@@ -121,6 +146,38 @@ def prepare_image(image_specification, bootstrap_data, outputdir):
 def _main():
     options = _argparser().parse_args()
 
+    # Build host -> (cert_path, key_path|None) mapping
+    client_auth_map = {}
+    if getattr(options, "client_cert_for", None):
+        for item in options.client_cert_for:
+            if "=" not in item:
```
praiskup: Can't we have a new option with `nargs='+'`, and avoid this parsing entirely?  With syntax like this?
```
---client-cert-for example.com:5000 certificate-file.crt
---client-cert-for example.com:5000 certificate-file.crt key-file.key
```

mock/py/mock-hermetic-repo.py
scoheb: fixed

mock/py/mock-hermetic-repo.py
scoheb: took a stab at addressing this. lmk

PR #1632, opened 2025-09-25
Enable module_hotfixes flag for local repositories

arogge:
Add `module_hotfixes=true` to repository configurations added by `add_local_repo()` (i.e. when using `--chain --localrepo <dir>`) so `dnf` will consider packages from this repository even when they are filtered out via modularity.

See https://dnf.readthedocs.io/en/latest/modularity.html#hotfix-repositories


praiskup:
No objections here.

PR #1642, opened 2025-10-16
Podman save using id.

matejonnet:
To workaround the error "Digest of source image's manifest would not match destination reference", when image is saved using sha256.

### DIFF
--- a/mock/py/mock-hermetic-repo.py
+++ b/mock/py/mock-hermetic-repo.py
@@ -137,7 +137,7 @@ def prepare_image(image_specification, bootstrap_data, outputdir):
     subprocess.check_output(pull_cmd)
     subprocess.check_output(["podman", "save", "--format=oci-archive", "--quiet",
                              "-o", os.path.join(outputdir, "bootstrap.tar"),
-                             image_specification])
+                             bootstrap_data["id"]])
 
 
 def _main():


mock/py/mock-hermetic-repo.py
```diff
@@ -137,7 +137,7 @@ def prepare_image(image_specification, bootstrap_data, outputdir):
     subprocess.check_output(pull_cmd)
     subprocess.check_output(["podman", "save", "--format=oci-archive", "--quiet",
                              "-o", os.path.join(outputdir, "bootstrap.tar"),
-                             image_specification])
+                             bootstrap_data["id"]])
```
praiskup, line 140: Thank you for the patch!

It's strange this is needed only with the UBI 9 image, but not with Fedora. That said, are we doing something wrong with `podman save`, or the subsequent `podman pull oci-archive:..`?  Is this a buggy OCI image?  Would you mind reporting this against the Podman project?

If this patch works, it works, and the change seems safe.  However, is it possible that `podman pull` could download an image with a different `bootstrap_data["id"]` (e.g., due to a race condition or a long delay between generating the lock file and downloading), leading to a subsequent `podman save` failure (or even worse, different (cached) image being stored)?

praiskup:
Confirmed on slack we can close this one.

Issue #1641, opened 2025-10-16
New nspawn option --suppress-sync=yes causes build failures


bleve:
This option is not harmless.

with `--supress-sync=yes` building knot-3.5.1 tests fail with segfault.

Please remove this optimization.

Here is snipped from failing testsuite from build log. By removing --suppress-sync option testsuite work fine.

I find this option very dangerous.

```
contrib/test_base32hex..........ok
contrib/test_base64.............ok
contrib/test_base64url..........ok
contrib/test_heap...............ok
contrib/test_inet_ntop..........ok
contrib/test_net................ok
contrib/test_net_shortwrite.....ok
contrib/test_qp-trie............ok
contrib/test_qp-cow.............ok
contrib/test_siphash............ok
contrib/test_sockaddr...........ok
contrib/test_string.............ok
contrib/test_strtonum...........ok
contrib/test_time...............ok
contrib/test_toeplitz...........ok
contrib/test_wire_ctx...........ok
libdnssec/test_binary...........ok
libdnssec/test_crypto...........ok
libdnssec/test_key..............ok
libdnssec/test_key_algorithm....ok
libdnssec/test_key_ds...........ok
libdnssec/test_keyid............ok
libdnssec/test_keystore_pkcs11..ok
libdnssec/test_keystore_pkcs8...ok
libdnssec/test_keytag...........ok
libdnssec/test_nsec_bitmap......ok
libdnssec/test_nsec_hash........ok
libdnssec/test_random...........ok
libdnssec/test_sign.............ok
libdnssec/test_sign_der.........ok
libdnssec/test_shared_bignum....ok
libdnssec/test_shared_dname.....ok
libdnssec/test_tsig.............ok
contrib/test_atomic.............ok
contrib/test_spinlock...........ok
knot/test_acl...................FAILED 6 (killed by signal 11, core dumped)
knot/test_changeset.............ok
knot/test_conf..................FAILED 1, 75 (killed by signal 11, core dumped)
knot/test_conf_tools............ok
knot/test_confdb................FAILED 1 (killed by signal 11, core dumped)
knot/test_confio................FAILED 1 (killed by signal 11, core dumped)
knot/test_digest................ok
knot/test_dthreads..............ok
knot/test_fdset.................ok
knot/test_journal...............dubious (killed by signal 11, core dumped)
knot/test_kasp_db...............FAILED 1-2 (killed by signal 11, core dumped)
knot/test_node..................ok
[...]

PR #1644, opened 2025-10-20
make suppress-sync opt-in

xsuchy:
Some tests in %check uses sync() and with this options fails. Until %check can be isolated we need to switch it off by default.

Fixes: #1641

Fixes: #NUMBER

> not deserve a release note.


rriemann:
Thanks for reverting the change. I guess I will only benefit from it at https://copr.fedorainfracloud.org/coprs/rriemann/eu-os/builds/ once mock has a new release. When can we expect such release?

Is there a workaround for COPR meanwhile?

/cc https://github.com/bootc-dev/bootc/issues/1692

praiskup:
See also #1613

PR #1651, opened 2025-11-04
Update Kylin OS configuration and fix `loongarch64`

scaronni:
Images are now kept up to date and all versions for Kylin 10 are now available.

- They are available for all 3 supported architectures (`x86_64`, `aarch64`, `loongarch64`) for both Kylin 10 and 11.
- The `loongarch64` for some reason is not listed on the registry web interface for version 10, but it's there.

Also, fix `forcearch` for `loongarch64`.

- The `loongarch64` architecture is missing from the `target_arch` to `forcearch` mapping, preventing builds for this architecture when executing `mock` on a different architecture.

### DIFF
--- a/mock-core-configs/etc/mock/templates/kylin-10.tpl
+++ b/mock-core-configs/etc/mock/templates/kylin-10.tpl
@@ -4,9 +4,8 @@ config_opts['releasever'] = '10'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 
-# No official up to date image available, the existing ones are all outdated:
-# https://cr.kylinos.cn/zh/image?name=kylin-server
-config_opts['use_bootstrap_image'] = False
+# Registry web interface: https://cr.kylinos.cn/zh/image
+config_opts['bootstrap_image'] = 'cr.kylinos.cn/kylin/kylin-server-platform:v10sp3-2403'
 
 config_opts['dnf.conf'] = """
 [main]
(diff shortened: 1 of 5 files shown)


mock-core-configs/etc/mock/templates/kylin-10.tpl
```diff
@@ -4,9 +4,8 @@ config_opts['releasever'] = '10'
 config_opts['package_manager'] = 'dnf'
 config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 
-# No official up to date image available, the existing ones are all outdated:
-# https://cr.kylinos.cn/zh/image?name=kylin-server
-config_opts['use_bootstrap_image'] = False
+# Registry web interface: https://cr.kylinos.cn/zh/image
+config_opts['bootstrap_image'] = 'cr.kylinos.cn/kylin/kylin-server-platform:v10sp3-2403'
```
praiskup, line 8: Isn't there a better tag? I'm concerned that we'll have to update this reference frequently in the future.

scaronni: Don't know why my comment disappeared, but no, no reasonable tags (like "latest") that can be used. Until a month ago there were just a subset of versions/combinations available.

praiskup: Ack, there's no other option then.

scaronni: They might do that, at one point. I will update if they do.

PR #1658, opened 2025-11-25
unbreq plugin: reword the usage of BuildRequire to BuildRequires

mkoncek:
`label no-release-notes`

This is more of a cosmetic change.
I don't like using the word `BuildRequire` as it is not used anywhere in the `.spec` files. I believe the proper term should be `BuildRequires field` or something like that. Or at the very least the sentence in the log could be interpreted as: "The line BuildRequires ... is not needed because...".


praiskup:
docs only

Issue #1639, opened 2025-10-13
mock --scrub=all doesn’t back up builds


sergiomb2:
The error may have existed for a long time, and I’ve only just noticed it.

When config_opts['backup_on_clean'] = True, mock --scrub=all doesn’t back up the builds, unlike mock --init.

I’m reporting this now as it might be easy for you to resolve the issue. In the meantime, I’ll also look for a solution.

PR #1660, opened 2025-11-26
Back up on scrub

sergiomb2:
Fixes: #1639 

> not deserve a release note.
- make mock --scrub=all correctly back up build directory
- Use 'mv' instead of 'cp' for backing up builds

### DIFF
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -1047,11 +1047,36 @@ class Buildroot(object):
         return util.BindMountedFile(chroot_filename, host_filename)
 
     @traceLog()
+    def backup_build_results(self):
+        """
+        Back up built RPMs if `backup_on_clean` is enabled, before cleaning the chroot and results.
+        """
+        if not self.config['backup_on_clean']:
+            return
+        srcdir = os.path.join(self.basedir, "result")
+        if not os.path.exists(srcdir):
+            return
+        rpms = glob.glob(os.path.join(srcdir, "*rpm"))
+        if len(rpms) == 0:
+            return
+        dstdir = os.path.join(self.config['backup_base_dir'], self.config['root'])
+        file_util.mkdirIfAbsent(dstdir)
+        self.state.state_log.info("backup_on_clean (completed builds): saving with mv %s %s", " ".join(rpms), dstdir)
+        for rpm in rpms:
+            dest_path = os.path.join(dstdir, os.path.basename(rpm))
+            try:
+                os.replace(rpm, dest_path)
+            except OSError as e:
+                self.state.state_log.error("backup_on_clean (completed builds): error moving %s to %s: %s",
+                        rpm, dest_path, e)
+
+    @traceLog()
     def delete(self):
         """
         Deletes the buildroot contents.
         """
         if os.path.exists(self.basedir):
+            self.backup_build_results()
             p = self.make_chroot_path()
             self._lock_buildroot(exclusive=True)
             util.orphansKill(p)
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -56,9 +56,6 @@ class Commands(object):
         self.more_buildreqs = config['more_buildreqs']
         self.cache_alterations = config['cache_alterations']
 
-        self.backup = config['backup_on_clean']
-        self.backup_base_dir = config['backup_base_dir']
-
         # do we allow interactive root shells?
         self.no_root_shells = config['no_root_shells']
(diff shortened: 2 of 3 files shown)


mock/py/mockbuild/buildroot.py
```diff
@@ -1052,6 +1052,23 @@
...
+                    dstdir = os.path.join(self.config['backup_base_dir'], self.config['root'])
+                    file_util.mkdirIfAbsent(dstdir)
+                    rpms = glob.glob(os.path.join(srcdir, "*rpm"))
+                    if rpms:
+                        self.state.state_log.info("backup_results: saving with mv %s %s", " ".join(rpms), dstdir)
+                        for rpm in rpms:
+                            dest_path = os.path.join(dstdir, os.path.basename(rpm))
+                            try:
+                                os.replace(rpm, dest_path)
+                            except Exception as e:
```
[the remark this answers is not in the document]
sergiomb2: why not ?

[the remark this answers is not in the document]
xsuchy: Because it can catch even different errors than the programmer intended. But since it is just a move of the old code to a different place, we can waive this warning.
EDIT: it is not a move of old code. Can you specify the exception(s)? I guess `(OSError, PermissionError)` should be enough.

mock/py/mockbuild/buildroot.py
```diff
@@ -1046,12 +1048,35 @@
...
+        file_util.mkdirIfAbsent(dstdir)
+        rpms = glob.glob(os.path.join(srcdir, "*rpm"))
+        if len(rpms) == 0:
+            return
+        self.state.state_log.info("backup_results: saving with mv %s %s", " ".join(rpms), dstdir)
+        for rpm in rpms:
+            dest_path = os.path.join(dstdir, os.path.basename(rpm))
+            try:
+                os.replace(rpm, dest_path)
+            except Exception as e:
```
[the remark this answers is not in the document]
sergiomb2: We have already reviewed this; no further error details are needed.

mock/py/mockbuild/buildroot.py
```diff
@@ -1046,12 +1048,38 @@ def wrap_host_file(self, filename):
...
     @traceLog()
     def delete(self):
         """
         Deletes the buildroot contents.
         """
         if os.path.exists(self.basedir):
+            # Backup the built RPMs before deleting the buildroot and results, if the backup_on_clean option is enabled
+            if self.backup:
+                self.backup_results()
+
```
praiskup: Style nit: Can you please move the `if self.backup` condition inside the new method, and just do `self.backup_results()`?

sergiomb2: Done. I also improved the function and variable names to make the intent clearer. I chose to use if self.backup_on_clean:, but we could switch to if not self.backup_on_clean: return if you prefer. Let me know if you want that change.

mock/py/mockbuild/buildroot.py
```diff
@@ -1046,12 +1048,36 @@ def wrap_host_file(self, filename):
 
         return util.BindMountedFile(chroot_filename, host_filename)
 
+    @traceLog()
+    def backup_build_results(self):
+        """
+        Back up the built RPMs before deleting the chroot and results, if the `backup_on_clean` option is enabled.
+        """
+        if self.backup_on_clean:
```
praiskup: Yeah, I agree that this would be nice:
```
if not self.config["backup_on_clean"]:
    return
```

mock/py/mockbuild/buildroot.py
```diff
@@ -99,6 +99,8 @@ def __init__(self, config, uid_manager, state, plugins, bootstrap_buildroot=None
         self._lock_file = None
         self.selinux = (not self.config['plugin_conf']['selinux_enable']
                         and util.selinuxEnabled())
+        self.backup_on_clean = config['backup_on_clean']
+        self.backup_base_dir = config['backup_base_dir']
```
praiskup: Are these two needed?  If there are no other readers, just drop them.

sergiomb2: I’ve updated the patch to use self.config directly for the  variables in question:  self.config['backup_on_clean'] and self.config['backup_base_dir'].

mock/py/mockbuild/buildroot.py
sergiomb2: if not self.config["backup_on_clean"]:
    return
is now in the patch

praiskup (approved):

praiskup:
thank you

PR #1662, opened 2025-12-02
Add risc-v fedora chroots

nikromen:


mock-core-configs/etc/mock/templates/fedora-riscv64.tpl
```diff
@@ -0,0 +1,49 @@
...
+
+config_opts['dist'] = 'fc{{ releasever }}'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+
+config_opts['package_manager'] = 'dnf5'
+
+# RISC-V doesn't have bootstrap images yet
+config_opts['use_bootstrap'] = False
+
+config_opts['qemu_user_static'] = '/usr/bin/qemu-riscv64-static'
```
xsuchy: Is this required?

mock-core-configs/etc/mock/templates/fedora-riscv64.tpl
```diff
@@ -0,0 +1,49 @@
...
+config_opts['description'] = 'Fedora {{ releasever }} RISC-V'
+config_opts['chroot_setup_cmd'] = 'install @build'
+
+config_opts['dist'] = 'fc{{ releasever }}'
+config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
+
+config_opts['package_manager'] = 'dnf5'
+
+# RISC-V doesn't have bootstrap images yet
+config_opts['use_bootstrap'] = False
```
xsuchy: should this be `use_bootstrap_image`?

mock-core-configs/etc/mock/templates/fedora-riscv64.tpl
praiskup: I doubt.  We probably want to fix Mock:
https://github.com/rpm-software-management/mock/blob/951ec753ce60f7f746da4ac38765cba5fdac0f6c/mock/py/mockbuild/config.py#L423-L433

praiskup: I noticed that this is pretty painful (recently we added loongarch64).  So please check #1668.

nikromen: thanks for the change pavel!

mock-core-configs/etc/mock/templates/fedora-riscv64.tpl
```diff
@@ -0,0 +1,47 @@
...
+metadata_expire=0
+best=1
+protected_packages=
+user_agent={{ user_agent }}
+
+# repos
+
+[local]
+name=local
+baseurl=https://riscv-koji.fedoraproject.org/repos/f{{ releasever }}-build/latest/$basearch/
```
praiskup: I can not install bootstrap chroot. :-( because I'm on x86
```
 local                                  100% | 898.0   B/s | 784.0   B |  00m01s
>>> Status code: 404 for https://riscv-koji.fedoraproject.org/repos/f43-build/la
>>> Status code: 404 for https://riscv-koji.fedoraproject.org/repos/f43-build/la
>>> Status code: 404 for https://riscv-koji.fedoraproject.org/repos/f43-build/la
>>> Status code: 404 for https://riscv-koji.fedoraproject.org/repos/f43-build/la
```
The thing is that bootstrap chroot is "native" to my host, and $basearch expands to x86_64.

nikromen: hmm it seems that https://github.com/rpm-software-management/mock/pull/1662#discussion_r2598508356 broke this... I think that we really need to use `use_bootstrap` here (it is even used by riscv folks here https://riscv-koji.fedoraproject.org/koji/buildinfo?buildID=45570 and I see my comment `# RISC-V doesn't have bootstrap images yet` was missleading):

```
# Sometimes the rpm/yum/dnf ecosystem on the host machine isn't really
# compatible with the rpm/yum/dnf ecosystem in mock chroot (the system we
# build for).  Typically when host is yum-based and target system is dnf-based.
# Such scenario may cause an error when preparing the target mock buildroot, or
# even worse an unexpected install transaction that is hard to detect.
# Therefore, with `use_bootstrap=True` default, we first install a minimal
# "bootstrap" chroot that only contains rpm/yum/dnf stack, and from that
# "bootstrap" chroot we initialize the target chroot.
#config_opts['use_bootstrap'] = True

# The bootstrap chroot is normally installed using a package manager from host,
# but in some cases even this isn't possible (e.g. installing rather new ZSTD
# compressed dnf.rpm package by 'dnf4_install_command' on epel-7 host).  In such
# case, you may need to have "bootstrap" chroot pre-populated from a container
# image first (where the package manager stack is already pre-installed, so
# mock doesn't have to).
#config_opts['use_bootstrap_image'] = True
```

anyway I updated PR to use `use_bootstrap` again instead and yes the `$basearch` is probably wrong here... is it okay to putting here `riscv64` directly?

xsuchy: > is it okay to putting here riscv64 directly?

Yes. I see no other option. As ian ntermediate step, it is fine.

praiskup: I'd prefer having some Jinja2 "if" condition here.  That should use normal fedora repos for bootstrap (emulation case), and use this repo normally (and $basearch would be OK imo).

nikromen: @praiskup could you please elaborate how to do the if condition based on the bootstrap image? Or do you mean something else? ()

xsuchy: I disagree with you @praiskup I think that would be overengineering. This whole config is just an intermediate step until it becomes fully supported, and then the separate template can be deleted.

praiskup: I just bet the configuration doesn't work, so :shrug: but we apparently are in hurry.

praiskup: > @praiskup could you please elaborate how to do the if condition based on the bootstrap image?

Not “if bootstrap image,” since we don’t use a bootstrap image at all. We initialize two chroots using DNF: the bootstrap chroot and the target chroot.  We need "if" condition for that, well, or disable bootstrap entirely.

I can take a look at how this should be done later, but let’s talk first before we overengineer anything.

xsuchy: We tried that with @nikromen and it works.

praiskup: I'm curious how that is possible:
```
$ rpm -q mock mock-core-configs
mock-6.6.test-1.20251223135058417684.main.5.ged2352a4.fc43.noarch
mock-core-configs-43.3.test-1.20251223135100289918.main.5.ged2352a4.fc43.noarch
```

```
 # /usr/bin/dnf5 --installroot /var/lib/mock/fedora-43-riscv64-bootstrap/root/ --releasever 43 install dnf5 dnf5-plugins --setopt=deltarpm=False --setopt=allow_vendor_change=yes --allowerasing --setopt=tsflags=nocontexts
Updating and loading repositories:
 fedora                                 100% |   1.6 MiB/s |  14.0 MiB |  00m09s
 local                                  100% | 668.4 KiB/s |  14.1 MiB |  00m22s
Repositories loaded.
Failed to resolve the transaction:
Problem 1: cannot install the best candidate for the job
  - package dnf5-5.2.17.0-2.rvre0.fc43.riscv64 from fedora does not have a compatible architecture
  - nothing provides ld-linux-riscv64-lp64d.so.1()(64bit) needed by dnf5-5.2.17.0-2.rvre0.fc43.riscv64 from fedora
  - nothing provides ld-linux-riscv64-lp64d.so.1(GLIBC_2.27)(64bit) needed by dnf5-5.2.17.0-2.rvre0.fc43.riscv64 from fedora
  - nothing provides rtld(GNU_HASH) needed by dnf5-5.2.17.0-2.rvre0.fc43.riscv64 from fedora
  - nothing provides libc.so.6(GLIBC_2.38)(64bit) needed by dnf5-5.2.17.0-2.rvre0.fc43.riscv64 from fedora
...
```

praiskup: Tried #1684 but doesn't seem to help.

PR #1674, opened 2025-12-11
Fix aarch64 configuration for Azure Linux 3

scaronni:
Copy / paste error when the original Azure Linux 3 Pull Request went in.

Didn't notice as we still had the local override for the `.cfg` file before the package was released with the changes.


xsuchy (approved):
+1

Issue #1681, opened 2025-12-23
Get rid of traceLog() entirely


praiskup:
It’s probably a matter of perspective, but it doesn’t really help with debugging—quite the contrary. It has only ever complicated my real-world debugging of Mock by adding an unnecessarily large call stack (try working with pdb and Mock, and you quickly learn that you can’t live without MOCK_TRACE_LOG=false, which was implemented as a result of my long-term despair).

There is also issue #1666, which seems to indicate that this may even slow down Mock execution.

praiskup:
triage: we could just turn it off by default for now, and maybe disable later (in a year or so).

PR #1682, opened 2025-12-31
SBOM generator

atomicturtle:
# Pull Request: Add SBOM Generator Plugin

## Summary
This PR adds a new plugin, `sbom_generator`, which automatically generates a Software Bill of Materials (SBOM) for packages built with Mock.

## How it Works
The plugin runs as a `postbuild` hook within the Mock lifecycle. It performs the following actions:
- **Introspection**: Queries the build chroot for active RPM macros to identify build environment configuration.
- **Metadata Extraction**: Extracts package headers, file lists, SHA-256 hashes, and GPG signatures from binary RPMs and SRPMs.
- **Dependency Resolution**: Maps runtime dependencies and optionally the build toolchain installed in the chroot.
- **Standardization**: Converts the collected data into standardized Package URL (PURL) and CPE identifiers.

## Produced Artifacts
The plugin generates a single output file in the build results directory:
- **`<name>-<version>-<release>.sbom`**: A valid CycloneDX 1.5 JSON file containing:
    - Metadata about the build host and distribution.
    - Components for all built packages, including their specific file lists and hashes.
    - Component entries for source code and patches.
    - A dependency graph linking the packages to their runtime requirements.
    - Build environment hardening status (PIE, RELRO, FORTIFY, etc.).

## Configuration
The following options are available in `config.py` to control the plugin behavior:
- **`generate_sbom`** (default: `True`): Enable or disable SBOM generation.
- **`include_file_components`** (default: `True`): Includes individual file entries and their hashes in the SBOM.
- **`include_file_dependencies`** (default: `False`): Maps explicit dependencies between files and their parent packages in the dependency graph.
- **`include_debug_files`** (default: `False`): Determines if debuginfo files are included in the manifests.
- **`include_man_pages`** (default: `True`): Determines if manual pages and documentation are included.
- **`include_source_dependencies`** (default: `True`): Establishes provenance by linking binary packages back to their source files (SRPMs/patches).
- **`include_toolchain_dependencies`** (default: `False`): Includes every package in the build environment as a dependency for full build-root auditing.

## Release Note
Added a new SBOM generator plugin. It creates a `<name>-<version>-<release>.sbom` file in the build results directory, documenting the package components, hashes, signatures, and build environment hardening status.


mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2024 @@
+# Copyright (C) 2025, Atomicorp, Inc.
+# SPDX-License-Identifier: GPL-2.0-only
```
xsuchy: Would you mind to license it under GPL-2.0-or-later as rest of the code?

xsuchy:
Can you add config option `type` which can have only one option 'cyclonedx`?  But I would love to add later spdx type.

xsuchy:
Please add release notes using https://github.com/rpm-software-management/mock/blob/main/docs/Release-Notes-New-Entry.md

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Yup! No problem

docs/Plugin-SBOM.md
```diff
@@ -0,0 +1,325 @@
...
+## Competitive Advantages
+
+This SBOM generator leverages Mock's unique build environment visibility:
+
+* **Complete Build Toolchain**: Captures every package installed in the build chroot, not just declared dependencies
+* **Build-Time Provenance**: Records the exact build environment, including tool versions and signatures
+* **RPM-Native Intelligence**: Deep integration with RPM metadata, spec files, and package signatures
+* **Reproducible Build Context**: Complete build environment fingerprinting for reproducibility verification
+
+Available since version 6.1.
```
xsuchy: Available since version 6.7.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            return
+
+        state_text = "Generating CycloneDX SBOM for built packages v1.0"
+        self.state.start(state_text)
+
+        try:
+            build_dir = self.buildroot.resultdir
+            # Filter out source RPMs from binary RPM processing
+            rpm_files = [f for f in os.listdir(build_dir) if f.endswith('.rpm') and not f.endswith('.src.rpm')]
+            src_rpm_files = [f for f in os.listdir(build_dir) if f.endswith('.src.rpm')]
```
xsuchy: Can you use os.scandir()? It is much faster. See https://peps.python.org/pep-0471/

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            except Exception:
+                pass
+                
+            print(f"Extracted source files from source RPM: {source_files}")
+        except Exception as e:
+            print(f"Failed to extract source files from source RPM {src_rpm_path}: {e}")
+        
+        return source_files
+
+    def _convert_to_chroot_path(self, host_path):
```
xsuchy: The name is strange. It does not ring bell to me. This is effectively the reverse function of `buildroot.make_chroot_path()`, right?
I would put it to buildroot.py and call it `from_chroot_path()`.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+        """Extracts source files from a source RPM."""
+        import tempfile
+        import shutil
+        import shlex
+        self.buildroot.root_log.debug(f"Extracting source files from source RPM: {src_rpm_path}")
+        source_files = []
+        try:
+            temp_dir = tempfile.mkdtemp(prefix="sbom-srpm-")
+            try:
+                extract_cmd = f"rpm2cpio {shlex.quote(src_rpm_path)} | cpio -idm 2>/dev/null"
```
xsuchy: From manpage of rpm2cpio:

> Note: the cpio(5) format cannot host individual files over 4GB in size, and so this tool is considered obsolete.  Use rpm2archive(1) instead.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            files = result.stdout.splitlines()
+            self.buildroot.root_log.debug(f"Files in RPM {rpm_path}: {files}")
+            return files
+        except subprocess.CalledProcessError as e:
+            self.buildroot.root_log.debug(f"Failed to get file list for {rpm_path}: {e.stderr}")
+            return []
+
+    def get_rpm_file_info(self, rpm_path):
+        """Extracts file hashes, ownership, and permissions from an RPM file using 'rpm -qp --dump'."""
+        cmd = ["rpm", "-qp", "--dump", rpm_path]
```
xsuchy: I would rather see calling python-rpm.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
+# Copyright (C) 2025, Atomicorp, Inc.
```
xsuchy: As vcs-diff-lint warns, this file is quite big. I would love to move to separate file these functions. E.g., get_rpm_file_info(), get_rpm_file_list()...

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+                    file_path = os.path.join(root_dir, file_name)
+                    sha256 = self.hash_file(file_path)
+                    signature = self.get_file_signature(file_path)
+                    source_files.append({
+                        "filename": file_name,
+                        "sha256": sha256,
+                        "digital_signature": signature
+                    })
+            try:
+                shutil.rmtree(temp_dir)
```
xsuchy: This can be problematic (because of access rights, special files,...) for this reason we have `file_util.rmtree()`

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            dependencies = result.stdout.splitlines()
+            self.buildroot.root_log.debug(f"Dependencies for RPM {rpm_path}: {dependencies}")
+            return dependencies
+        except subprocess.CalledProcessError as e:
+            self.buildroot.root_log.debug(f"Failed to get dependencies for {rpm_path}: {e.stderr}")
+            return []
+
+    def get_rpm_signature(self, rpm_path):
+        """Extracts the GPG signature of an RPM file."""
+        cmd = ["rpm", "-qpi", rpm_path]
```
xsuchy: python-rpm?

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+        except subprocess.CalledProcessError as e:
+            self.buildroot.root_log.debug(f"RPM command failed for {rpm_path}: {e.stderr}")
+            return {}
+        except Exception as e:
+            self.buildroot.root_log.debug(f"Failed to extract RPM metadata: {e}")
+            return {}
+
+    def get_rpm_file_list(self, rpm_path):
+        """Extracts the list of files from an RPM file."""
+        cmd = ["rpm", "-qpl", rpm_path]
```
xsuchy: python-rpm?

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+        
+        # Generate CPE
+        cpe = f"cpe:2.3:a:{vendor}:{product}:{version}:*:*:*:*:*:*:*:*"
+        return cpe
+
+    def detect_chroot_distribution(self):
+        """Detects the distribution name inside the chroot by reading /etc/os-release."""
+        try:
+            # Use buildroot's doChroot to cat /etc/os-release
+            cmd = ["cat", "/etc/os-release"]
```
xsuchy: Can you use python-distro? We already use it in `mockbuild.util`. Using https://distro.readthedocs.io/en/latest/#distro.LinuxDistribution you can change the root_dir and query the chroot path.
You can move this function to fileutil.py.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            product = "openssl"
+        elif package_name == "gcc":
+            vendor = "gnu"
+            product = "gcc"
+        elif package_name == "make":
+            vendor = "gnu"
+            product = "make"
+        elif package_name == "gettext":
+            vendor = "gnu"
+            product = "gettext"
```
xsuchy: Uaaaa, this is ugly. Can you make a dict with {package_name: {vendor: X, product: Y}, ....} and look it up?

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            
+            return None
+        except Exception as e:
+            self.buildroot.root_log.debug(f"Failed to check signature for {file_path}: {e}")
+            return None
+
+    def get_iso_timestamp(self):
+        """Returns the current time in ISO 8601 format."""
+        from datetime import datetime
+        return datetime.utcnow().isoformat() + "Z"
```
xsuchy: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

mock/py/mockbuild/plugins/sbom_generator.py
praiskup: I don't think this would work reasonably well; we have to account for bootstrapped use-cases.

praiskup:
@azhuzhu can you give this a brief look?  How much do we overlap with the syft generator?

docs/Plugin-SBOM.md
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: For reference this functions job is  to let it run other tools like rpmspec from the build host, which we need to expand all the macros in the rpm:

For example say the spec declares:
%define gitver  f6d60fd

and you have this:
Source0:  app-%{version}-%{gitver}.tar.gz


I poked around and saw 2 other plugins doing something similar:
ubreq.py 
uses short_path = path[len(self.buildroot.rootdir):] <- this one is almost identical to what I am doing, so this is what I went with to minimize changes

and
rpkg_preprocessor.py
rootdir_prefix = self.buildroot.make_chroot_path()

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
xsuchy: > self.buildroot.make_chroot_path()

This is **reverse** to what you are doing. self.buildroot.make_chroot_path('FOO') is basically self.buildroot.rootdir + '/' + 'FOO'.

ad ubreq - it does not make this code a function.

The code of your function is correct. I just do not like the name.

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Got it, do you want me to keep it local or update buildroot.py?

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Getting rid of this. If the .spec's Vendor field is incorrect, thats a problem on the .spec side. We shouldnt be manipulating that in the output SBOM

mock/py/mockbuild/plugins/sbom_generator.py
xsuchy: Both variants are ok for now. The latter makes a little bit more sense to me.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2025 @@
...
+            sig_valid = signature_info.get("signature_valid", False)
+            properties.append({
+                "name": "mock:signature:valid",
+                "value": str(sig_valid)
+            })
+        
+        return properties
+
+    def parse_spec_file(self, spec_path):
+        """Parses a spec file to extract source and patch files with their hashes and signatures."""
```
xsuchy: BTW you can use https://github.com/packit/specfile/ This is packaged in Fedora and EPEL. And it ease the parsing of specfile a lot.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2067 @@
...
+            
+            # Use os.scandir for better performance
+            with os.scandir(build_dir) as entries:
+                for entry in entries:
+                    if not entry.is_file():
+                        continue
+                    if entry.name.endswith('.rpm') and not entry.name.endswith('.src.rpm'):
+                        rpm_files.append(entry.name)
+                    elif entry.name.endswith('.src.rpm'):
+                        src_rpm_files.append(entry.name)
```
azhuzhu: This can save a duplicate sub-condition
```suggestion
                    if entry.name.endswith('.src.rpm'):
                        src_rpm_files.append(entry.name)
                    elif entry.name.endswith('.rpm'):
                        rpm_files.append(entry.name)
```

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2067 @@
...
+                        # Add license information
+                        if primary_component.get("licenses"):
+                            component_obj["licenses"] = primary_component.get("licenses")
+                        elif primary_rpm_metadata:
+                            license_str = primary_rpm_metadata.get("license")
+                            if license_str and license_str != "(none)":
+                                component_obj["licenses"] = [
+                                    {
+                                        "license": {
+                                            "id": license_str
```
azhuzhu: It looks `license.id` must be one valid string listing in [CycloneDX v1.5 scheme](https://cyclonedx.org/docs/1.5/json/#components_items_licenses_oneOf_i0_items_license_id). So probably you can use SPDX License Expression instead with `licensfedora2spdx` of [license-validate](https://pagure.io/copr/license-validate)

azhuzhu (commented):
> @azhuzhu can you give this a brief look?  How much do we overlap with the syft generator? 

Although both are to generate an SBOM for RPM, this plugin is different from the SBOM generator we have.

* This plugin is based on the mock buildroot for one SRPM with a target arch; Our approach is based on an SRPM with multiple arches. So this doesn't have an SRPM Component as the root.
* There are a list of sources and patches here as component items, but in our generator, we only record source archives as Ancestors of the SRPM component
* This plugin also parse runtime deps, which we didn't do.
* This plugin doesn't use other external tools (like `syft`) to scan the content of (S)RPMS and merge the result in the output but we do.
* Some details of the logic are different

mock/py/mockbuild/plugins/sbom_generator.py
xsuchy: Almost all Fedora packages have a license in the SPDX License Expression format. And almost all RHEL 10 packages. 
I suggest assuming that the license string in RPM **is** in SPDX format. And if not, then it is a bug of that specific package.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2067 @@
...
+            cpe_vendor_default = self.detect_chroot_distribution() or "unknown"
+            import re
+            import datetime
+            
+            for line in output.splitlines():
+                parts = line.split("|", 3)
+                if len(parts) < 3:
+                    continue
+                package_name = parts[0].strip()
+                package_version = parts[1].strip()
```
azhuzhu: Why does this include ${ARCH}, but in other places calling `_generate_purl`, it's f"{VERSION}-{RELEASE}"?

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: done

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2265 @@
...
+        except Exception as e:
+            self.buildroot.root_log.debug(f"Failed to extract RPM metadata: {e}")
+            return {}
+
+    def get_rpm_file_list(self, rpm_path):
+        """Extracts the list of files from an RPM file."""
+        cmd = ["rpm", "-qpl", rpm_path]
+        try:
+            result = subprocess.run(
+                cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True, text=True
```
praiskup: Can you use the built-in tooling, like here?

https://github.com/rpm-software-management/mock/blob/ddb61251aa7186db0451c52be7780b4b606baaf0/mock/py/mockbuild/plugins/buildroot_lock.py#L11

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,2265 @@
...
+        except subprocess.CalledProcessError as e:
+            self.buildroot.root_log.debug(f"Failed to get file list for {rpm_path}: {e.stderr}")
+            return []
+
+    def get_rpm_file_info(self, rpm_path):
+        """Extracts file hashes, ownership, and permissions from an RPM file using rpm-python."""
+        # pylint: disable=no-member
+        file_info = {}
+        try:
+            ts = rpm.TransactionSet()
```
praiskup: This is extremely brittle.  We are using rpm on host, but working with a "target" RPM database -- it's quite likely you can't even read it.

I mean, I think we can document that we don't do DNF/RPM bootstrapping... (that the plugin is broken for targets with incompatible DNF/RPM, but it's a pity)

docs/Plugin-SBOM.md
```diff
@@ -0,0 +1,325 @@
+---
+layout: default
+title: Plugin SBOM Generator
+---
+
+This plugin generates a Software Bill of Materials (SBOM) in CycloneDX format for packages built with Mock. The SBOM provides detailed information about the build environment, source files, and resulting packages, optimized for security use cases.
+
+## Features
+
+* Generates SBOM in CycloneDX 1.5 format (JSON)
```
praiskup: Is SPDX supported now, too?

atomicturtle: Yup! it is as of the last round. Im getting back here, I just had to get some other projects updated ahead of this one. Ill be working in the other changes here soon

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Good catch. You're right—this was inconsistent with how we handle built packages. I've updated the toolchain metadata extraction to separate %{ARCH} from the version string so that all PURLs follow the same standard @version?arch=arch format.

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: I think I got all of these when I shifted to python-rpm, the only place I havent had any success eliminating is  with expanding the spec macros using rpmspec --parse.

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: I shifted this to the internal model rpmautospec and unbreq are using

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Got it, I broke these up into separate files in the plugins/ dir:

mock/py/mockbuild/plugins/sbom_cyclonedx.py 
mock/py/mockbuild/plugins/sbom_spdx.py  
mock/py/mockbuild/plugins/sbom_utils.py

praiskup:
Can you please rebase?

atomicturtle:
Anything else you want me to do here?

praiskup:
Sorry for the delay on the review.  This is going to get much more attention now, thank you again for your work!

@atomicturtle would you mind crating an executable script doing this business, so we could run it without running the whole mock toolchain?  In mock we would just call the script (we can move the script to a separate project if needed).

mock/py/mockbuild/plugins/sbom_utils.py
```diff
@@ -0,0 +1,832 @@
+# -*- coding: utf-8 -*-
```
praiskup: This file is a library file -- should live in toplevel mockbuild module, to not confuse it with plugin.

mock/py/mockbuild/plugins/sbom_utils.py
```diff
@@ -0,0 +1,832 @@
...
+                "/builddir/build/RPMS/noarch",
+                "/builddir/build/SRPMS",
+                "/builddir/build/SOURCES"
+            ]
+            for search_path in search_paths:
+                candidate = os.path.join(search_path, filename)
+                # Verify existence via doChroot
+                cmd = ["ls", candidate]
+                try:
+                    res, _ = self.buildroot.doChroot(cmd, shell=False, returnOutput=True, printOutput=False)
```
praiskup: The `doChroot` method isn't optimal.  It executes RPM in buildroot, which often isn't working well (even if mock builds fine using RPM from bootstrap or host).

You should be using doOutChroot() (check buildroot_lock that does similar things).

praiskup: Also check package_state.py.

mock/py/mockbuild/plugins/sbom_utils.py
atomicturtle: Done

mock/py/mockbuild/plugins/sbom_utils.py
atomicturtle: Done, and done

atomicturtle:
This update changes:

    1. Moved library components out of plugins dir
    
            mock/py/mockbuild/plugins/sbom_cyclonedx.py -> mock/py/mockbuild/sbom_cyclonedx.py
            mock/py/mockbuild/plugins/sbom_spdx.py -> mock/py/mockbuild/sbom_spdx.py
    
    2. Created a new CLI generator mock/py/mock-sbom-generator.py, this can
       be run standalone. This does present some forensic gaps vs running
       inline as it is performing operations outside of the build chain.
    
    3. Added a man page
    
    4. Extends forensic coverage to package level sha-256, RSAHEADER, and
       src.rpm GPG checking.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,351 @@
...
+                "include_debug_files",
+                "include_man_pages",
+                "include_source_dependencies",
+                "include_toolchain_dependencies",
+                "generate_cpe",
+            ):
+                if key in self.conf:
+                    flag = "--" + key.replace("_", "-")
+                    value = "true" if self.conf[key] else "false"
+                    cmd.extend([flag, value])
```
praiskup: Can we please convert those into a command-line directly in Mock config?
Something like this?
```
# config_opts['plugin_conf']['sbom_generator_opts'] = {
#     'generate_sbom': True,
#     'type': 'cyclonedx',  # or 'spdx'
#     'command': '/usr/bin/mock-sbom-generator --type cyclonedx --include-file-components etc...',
```
This would seem more flexible (ie., the generator can live in Mock as long as we feel it scales well, but not necessarily).

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,351 @@
...
+# pylint: enable=invalid-name
+
+DEFAULT_COMMAND = "/usr/bin/mock-sbom-generator"
+# Visible forensic artifact retained in the result directory
+PREBUILD_STATE_FILENAME = "sbom-prebuild.json"
+# Legacy hidden name from earlier builds (still accepted if present)
+LEGACY_PREBUILD_STATE_FILENAME = ".sbom-prebuild.json"
+
+
+@traceLog()
```
praiskup: no need to use traceLog: #1681

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,351 @@
...
+        """Return the primary mock config *file* path (not the config directory)."""
+        config = self.buildroot.config
+        # Prefer the full list of loaded config files; last non-site file is usually
+        # the chroot cfg the user selected.
+        paths = list(config.get("config_paths") or [])
+        for candidate in reversed(paths):
+            if not candidate:
+                continue
+            base = os.path.basename(candidate)
+            if base in ("site-defaults.cfg", "logging.ini"):
```
praiskup, line 266: Are you searching for `config_file`? https://github.com/rpm-software-management/mock/blob/6f7c49f51a732a31084fcf1d589527dd92def3cb/mock/py/mockbuild/config.py#L1064

praiskup: If you attempt to do export a different thing, please fix the docstring.

mock/py/mockbuild/plugins/sbom_generator.py
```diff
@@ -0,0 +1,351 @@
...
+        elif isolation == "nspawn":
+            use_nspawn = True
+        else:
+            # isolation is None or "auto"
+            if use_nspawn is None:
+                use_nspawn = bool(mockbuild.util.USE_NSPAWN)
+            isolation = "nspawn" if use_nspawn else "simple"
+
+        env["isolation"] = str(isolation)
+        env["use_nspawn"] = bool(use_nspawn)
```
praiskup, line 306: I am curious if we want to export the config entirely, something like --debug-config or --debug-config-expanded.

mock/py/mockbuild/sbom_cyclonedx.py
```diff
@@ -0,0 +1,979 @@
+# -*- coding: utf-8 -*-
+# vim:expandtab:autoindent:tabstop=4:shiftwidth=4:filetype=python:textwidth=0:
+# License: GPL2 or later see COPYING
+# Written by Scott R. Shinn <scott@atomicorp.com>
+# Copyright (C) 2026, Atomicorp, Inc.
```
praiskup, line 5: I'd juse SPDX format (like in spec) for licenses, ie GPL-2.0-or-later (but FWIW, I don't think we need these `#` headers at all these days).

mock/py/mock-sbom-generator.py
```diff
@@ -0,0 +1,344 @@
...
+            new_path = os.path.realpath(os.path.join(new_path, relative))
+            if os.path.commonpath((root, new_path)) != root:
+                raise ValueError(f"path escapes buildroot: {path!r}")
+        return new_path
+
+    def from_chroot_path(self, host_path):
+        from mockbuild.util import host_path_to_chroot_path
+        return host_path_to_chroot_path(host_path, self.rootdir)
+
+    def doOutChroot(self, command, *args, **kwargs):  # pylint: disable=invalid-name,unused-argument
```
praiskup, line 92: I meant that you could use the Mock's built-in doOutChroot method, not a re-defined one.

mock/py/mockbuild/sbom_utils.py
```diff
@@ -0,0 +1,1598 @@
...
+        self._signature_cache = {}
+
+    @staticmethod
+    def _rpm_transaction_set_noverify():
+        """TransactionSet that reads headers without signature/digest checks.
+
+        Cryptographic verification is handled separately by verify_rpm_signature.
+        """
+        # pylint: disable=no-member,protected-access
+        ts = rpm.TransactionSet()
```
praiskup, line 148: To get this working well, we need the script to be automatically copied inside the bootstrap chroot, and use the "native RPM".

There are basically two important things the plugin should do in my opinion.  Make sure that the right SBOM tool is well-installed in the bootstrap chroot (or on host, if boostrap=off), and execute it (with the arguments user prefers).

praiskup: I mean, there's no precedent for "copy-pasting-script into a (bootstrap) chroot", but we were discussing this several times ... so it might be the time to do it now?

Btw., we always found a way-around it by running `rpm --root` (doOutChroot) like [this](https://github.com/rpm-software-management/mock/blob/6f7c49f51a732a31084fcf1d589527dd92def3cb/mock/py/mockbuild/plugins/package_state.py#L74)

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: Got it, I changed:

- Removed that programmatic flag-building loop
- command is now a full argv template (default in the plugin, documented in site-defaults.cfg), expanded with %() then shlex.split
- (optionally) - Users can replace command with any generator while still getting Mock path/env injection

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: done

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: _resolve_config_file now prefers config_opts['config_file'] first (the path Mock sets in config.py). The site-defaults.cfg / logging.ini skip of config_paths is only a fallback if that key is missing or the file isn’t there.

mock/py/mockbuild/plugins/sbom_generator.py
atomicturtle: yeah I can see getting that in general, what I was trying to answer here was if mock had rpmbuild_networking enabled or not when the binary was being generated. Im sure there are a bunch of other attack vectors we could isolate by capturing the whole config though

mock/py/mockbuild/sbom_cyclonedx.py
atomicturtle: Got it

mock/py/mockbuild/sbom_utils.py
atomicturtle: Got it (this was a pretty big change)

mock/py/mock-sbom-generator.py
atomicturtle: Got it,  plugin now uses Mock’s real Buildroot.doOutChroot for bootstrap-native mode

praiskup: Do we still need this method defined, then?

I'm struggling to understand how the SBOM generation tool/logic should be called:
a) on the host, b) in the bootstrap (if enabled), or c) inside the chroot?

My understanding is that you prefer a), with certain tools being called inside bootstrap chroot (via doOutChroot(), which falls back to on-host).

mock/py/mockbuild/buildroot.py
```diff
@@ -195,6 +195,10 @@ def make_chroot_path(self, *paths):
             new_path = os.path.join(new_path, path)
         return new_path
 
+    def from_chroot_path(self, host_path):
+        """Convert an absolute host path into the corresponding path inside the build chroot."""
+        return util.host_path_to_chroot_path(host_path, self.rootdir)
```
praiskup: Is this change needed?

mock/docs/site-defaults.cfg
```diff
@@ -302,6 +302,51 @@
...
+#         ' --use-nspawn %(use_nspawn)s'
+#     ),
+#     'type': 'cyclonedx',  # or 'spdx' (also available as %(type)s in command)
+#     'generate_cpe': False,
+#     'include_file_components': True,
+#     'include_file_dependencies': False,
+#     'include_debug_files': False,
+#     'include_man_pages': True,
+#     'include_source_dependencies': True,
+#     'include_toolchain_dependencies': False,
```
praiskup: Imagine there would be other sbom generators available on host in the future.
It would be nice to avoid most of those config options, and modify `command` directly?

mock/py/mock-sbom-generator.py
atomicturtle: Yes, we still need it but only for the standalone CLI:

a) When bootstrap is off (and always for the standalone CLI), the generator runs on the host. Nothing from a Mock chroot is invoked. That makes sense if you don’t trust the chroot, like if networking was allowed and the build pulled in golang deps or similar. Tradeoff: host vs target RPM mismatches, and it can under-report macros / distro quirks (things we score in sbom-auditor, a related but separate project). TL;DR: the SBOM is more trusted, but it has lower forensic fidelity.

b) Bootstrap — Preferred when bootstrap is on: the plugin stages the host-trusted generator into the bootstrap chroot and runs it via Mock’s real Buildroot.doOutChroot (bootstrap-native rpm against the target --root). This is much  more accurate on complex specs that use nested macros and incompatible target RPM DBs. Tradeoff: slightly less trust / more moving parts than pure host. TL;DR: wins on correctness; loses a bit on trust and simplicity. For most Mock builds (bootstrap on, including typical Koji setups), (b) is the right answer; (a) is the fallback / standalone / host-only-trust case.

c) In the target buildroot, we'd never install or run the generator where %build or %check get invoked. 

This duck-typed doOutChroot is only the CLI’s stand-in for (a). On the Mock plugin path with bootstrap enabled, we use real Buildroot.doOutChroot.

mock/py/mockbuild/buildroot.py
atomicturtle: Got it, removed

mock/docs/site-defaults.cfg
atomicturtle: Changed, heres an example:

```
# SPDX instead of CycloneDX
config_opts['plugin_conf']['sbom_generator_opts'] = {
    'generate_sbom': True,
    'command': (
        '/usr/bin/mock-sbom-generator'
        ' --type spdx'
        ' --resultdir %(resultdir)s'
        ' --root %(root)s'
        ' --builddir %(builddir)s'
        ' --include-toolchain-dependencies true'
        ' --online %(online)s'
        ' --rpmbuild-networking %(rpmbuild_networking)s'
        ' --isolation %(isolation)s'
        ' --use-nspawn %(use_nspawn)s'
    ),
}

# Or an external generator later:
# 'command': '/usr/bin/other-sbom-tool --out %(resultdir)s --chroot %(root)s'

```
and this is definitely something going on with the larger E-SBOM mandate, where they have tools analyzing binaries and creating a binary level SBOM analysis. I dont know how I'm going to work those into this yet, but its a good way to leave the design open

PR #1686, opened 2026-01-09
unbreq plugin: log standard outputs of failed commands

mkoncek:
In case that `subprocess.run` fails, the raised exception does not print the outputs of the command. Those outputs are very important for debugging and should be logged.
I was not sure about the exact format:
* one log event per one line of output?
* one log event per stream?
* ERROR / WARNING / INFO - mapping to stdout / stderr?

An alternative I considered was:
```python
if len(process.stdout) != 0:
	getLog().error("stdout:\n%s", process.stdout.strip())
if len(process.stderr) != 0:
	getLog().error("stderr:\n%s", process.stderr.strip())
```

### DIFF
--- a/mock/py/mockbuild/plugins/unbreq.py
+++ b/mock/py/mockbuild/plugins/unbreq.py
@@ -112,6 +112,11 @@ class Unbreq:
 
     @traceLog()
     def check_output(self, command: list[str], *args: Any, expected_returncode = 0, **kwargs: Any) -> str:
+        """
+        Run `command` in the associated chroot. Raise an exception if returned code
+        does not match `expected_returncode`. Additional arguments are passed to
+        the function `mockbuild.util.do_with_status`.
+        """
         # The `--ephemeral` flag is required in order to be able to run `systemd-nspawn` concurrently.
         kwargs["nspawn_args"] = ["--ephemeral", "--bind", self.buildroot.rootdir]
         if self.buildroot.bootstrap_buildroot is not None:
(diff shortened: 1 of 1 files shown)


mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -40,6 +40,31 @@ def __missing__(self, key: str) -> float:
...
+    process = subprocess.run(command, *args, **kwargs)
+    if process.returncode != expected_returncode:
+        for line in process.stdout.splitlines():
+            getLog().error("%s", line)
+        for line in process.stderr.splitlines():
+            getLog().error("%s", line)
+        raise subprocess.CalledProcessError(
+            process.returncode, " ".join(process.args), process.stdout, process.stderr
+        )
+    return process
```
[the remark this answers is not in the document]
mkoncek: Good point but yes, I believe if the plugin can not invoke `/usr/bin/findmnt` then it probably can not behave correctly, so the new behaviour is good. I should remove the questioning comment though.

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -40,6 +40,34 @@ def __missing__(self, key: str) -> float:
             self[key] = result
         return result
 
+@traceLog()
+def _run_subprocess(
+    command: list[str], *args: Any, expected_returncode: int = 0, **kwargs: Any
+) -> subprocess.CompletedProcess:
```
xsuchy: This starts looking like `do_with_status()` https://github.com/rpm-software-management/mock/blob/main/mock/py/mockbuild/util.py#L527

mkoncek: Good point, I will try to see if it could be replaced.

mkoncek:
This PR has grown in size from the initial one, but there should be no more changes.

PR #1697, opened 2026-01-28
mock: disable networking in bootstrap for hermetic builds

praiskup:


mock/py/mock.py
```diff
@@ -812,6 +811,8 @@ def main():
             mount_point = BindMountPoint(srcpath=key_dir, bindpath=chroot_dir)
             bootstrap_buildroot.mounts.add(mount_point)
 
+        util.setup_host_resolv(bootstrap_buildroot_config)
```
xsuchy: Hmm. But plugins and bindmountpoint will not have a functional resolv.conf.

praiskup: The issue is that Mock tries to (unnecessarily) copy a non-existing resolv.conf on host into bootstrap.... (when `podman run --network=none` is being used).

tkopecek: Related Fedora issue https://bugzilla.redhat.com/show_bug.cgi?id=2433808

mock/py/mockbuild/util.py
```diff
@@ -869,7 +870,7 @@ def setup_host_resolv(config_opts):
                 ::1       localhost localhost.localdomain localhost6 localhost6.localdomain6
                 ''')
 
-    if config_opts['isolation'] == 'simple':
+    if not USE_NSPAWN:
```
praiskup, line 873: @tkopecek This turned out to be the most important change - since we use `--isolation=auto` (in container we do a fallback to 'simple'), we shouldn't even execute the rest of the method ...

xsuchy, line 873: Modus operandi of this line was that USE_NSPAWN is global variable. Made in rush. With knowledge that globals are bad. And one day we can replace is with propper variable. But propagating config_opt in some function will be hard. On this place we have config_opts easily accessible.

mock/py/mockbuild/util.py
```diff
@@ -889,7 +890,10 @@ def setup_host_resolv(config_opts):
     os.chmod(resolv_path, 0o644)
 
     if config_opts['use_host_resolv']:
-        shutil.copyfile('/etc/resolv.conf', resolv_path)
+        try:
+            shutil.copyfile('/etc/resolv.conf', resolv_path)
+        except FileNotFoundError:
+            log.warning("Non-existing resolv.conf on host, using an empty one")
```
xsuchy, line 896: ??? so the file does not exist. And then we bindmount the nonexistent file? This works? Should not we at least touch it?

mock/py/mockbuild/util.py
praiskup: We though don't have this specific value in config_opts (the value is result of config.py hack with globals that you described).

mock/py/mockbuild/util.py
praiskup: We do create the file (empty one) for the bindmount a few lines above..

xsuchy: True.

Issue #1705, opened 2026-02-10
Ability to record parsed specfile in chroot


azhuzhu:
specfile extending strongly depends on the macros in chroot. It would be nice to have mock supporting that as an building output.

PR #1706, opened 2026-02-10
expand_spec plugin

azhuzhu:
Fixes: #1705

In some cases, we need information in specfile which are not included in
(S)RPMs' headers, e.g: Source and Patch. Expanding it within the chroot is
the simplest and most correct way, so that we don't have to replicate the
same buildroot later.

In the plugin, specfile is expanded by:
```
rpmspec --parse <specfile>
```

And to diable it, adding this in mock.cfg:
```
config_opts['plugin_conf']['expand_spec_enable'] = False

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -33,7 +33,7 @@ PLUGIN_LIST = ['tmpfs', 'root_cache', 'yum_cache', 'mount', 'bind_mount',
                'lvm_root', 'compress_logs', 'sign', 'pm_request',
                'hw_info', 'procenv', 'showrc', 'rpkg_preprocessor',
                'rpmautospec', 'buildroot_lock', 'export_buildroot_image',
-               'unbreq']
+               'unbreq', 'expand_spec']
 
 def nspawn_supported():
     """Detect some situations where the systemd-nspawn chroot code won't work"""
@@ -257,7 +257,11 @@ def setup_default_config_opts():
         'unbreq_enable': False,
         'unbreq_opts': {
             'exclude_accessed_files': []
-        }
+        },
+        'expand_spec_enable': True,
+        'expand_spec_opts': {
+            'rpmspec_opts': [],
+        },
     }
 
     config_opts['environment'] = {
--- a/mock/py/mockbuild/plugins/expand_spec.py
+++ b/mock/py/mockbuild/plugins/expand_spec.py
@@ -1,0 +1,47 @@
+"""A mock plugin to expand specfile in postdeps phase."""
+
+# our imports
+import os
+from mockbuild.trace_decorator import getLog, traceLog
+
+# pylint: disable=invalid-name
+requires_api_version = "1.1"
+
+
+# plugin entry point
+@traceLog()
+def init(plugins, conf, buildroot):
+    """Register the expand_spec plugin with mock."""
+    ExpandSpec(plugins, conf, buildroot)
+
+
+class ExpandSpec:
+    """Get the runtime rpmspec --parse"""
+    # pylint: disable=too-few-public-methods
+    @traceLog()
+    def __init__(self, plugins, conf, buildroot):
+        """Init the plugin."""
+        self.buildroot = buildroot
+        self.opts = conf
+        self.config = buildroot.config
+        self.logger = getLog()
+
+        # actually run our plugin at this step
+        plugins.add_hook("postdeps", self._postDepsHook)
+        self.logger.info("expand_spec: initialized")
+
+    @traceLog()
+    def _postDepsHook(self):
+        """postdeps hook to expand specfile by ``rpmspec --parse``."""
+        self.logger.info("Executing expand_spec plugin")
+        chroot_spec = self.buildroot.spec
+        expanded_specfile = os.path.join(self.buildroot.resultdir, 'expanded-spec.txt')
+        self.logger.info("Expanding original spec: %s to %s", chroot_spec, expanded_specfile)
+
+        cmd = ["/usr/bin/rpmspec", "--parse", chroot_spec] + self.opts.get("rpmspec_opts", [])
+        output, _ = self.buildroot.doChrootPlugin(cmd, returnOutput=True)
+        with open(expanded_specfile, "w", encoding="utf-8") as o:
+            o.write(output)
+
+        self.buildroot.uid_manager.changeOwner(expanded_specfile, gid=self.config['chrootgid'])
+        getLog().info("Expanded specfile written to: %s", expanded_specfile)
--- a/docs/Plugin-ExpandSpec.md
+++ b/docs/Plugin-ExpandSpec.md
@@ -1,0 +1,27 @@
+---
+layout: default
+title: Plugin ExpandSpec
+---
+
+This plugin expands specfile and save the output to `expanded-spec.txt` in result directory in `postdeps` hook.
+
+It uses to expand the specfile
+```bash
+/usr/bin/rpmspec --parse <spec>
+```
+
+## Configuration
+
+You can disable the plugin using this settings:
+```python
+config_opts['plugin_conf']['expand_spec_enable'] = False
+```
+
+To add extra options in `rpmspec` command, use:
+```python
+config_opts['plugin_conf']['expand_spec_opts']['rpmspec_opts'] = ['--verbose', '-d', 'foo bar']
+```
+
+Available since version 6.7.
+
+This plugin is ENABLED by default.
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -62,6 +62,10 @@ class Commands(object):
         self.private_network = not config['rpmbuild_networking']
         self.rpmbuild_noclean_option = None
 
+        # on-demand buildroot properties
+        # spec path in buildroot
+        self.buildroot.spec = None
+
     @traceLog()
     def clean(self):
         """clean out chroot with extreme prejudice :)"""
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: mock/py/mockbuild/plugins/show_parsed_spec.py [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+        # actually run our plugin at this step
+        plugins.add_hook("pre_srpm_build", self._PreBuildHook)
+
+    @traceLog()
+    def _PreBuildHook(self, host_chroot_spec, host_chroot_sources):
+        getLog().info("enabled show_parsed_spec plugin")
+        rootdir_prefix = self.buildroot.make_chroot_path()
+        chroot_spec = host_chroot_spec.replace(rootdir_prefix, '')
+        chroot_sources = host_chroot_sources.replace(rootdir_prefix, '')
+        parsed_specfile = self.buildroot.resultdir + '/' + os.path.basename(chroot_spec)[:-4] + 'parsed.spec'
```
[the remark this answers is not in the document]
praiskup: Can we use a static name like `parsed-spec.txt`?

mock/py/mockbuild/config.py
```diff
@@ -257,7 +257,11 @@ def setup_default_config_opts():
         'unbreq_enable': False,
         'unbreq_opts': {
             'exclude_accessed_files': []
-        }
+        },
+        'show_parsed_spec_enable': True,
+        'show_parsed_spec_opts': {
```
praiskup: Nitpick: "show_" prefix evokes that it is going to be printed on stdout or something like that;  I'd maybe just use "parse_spec" or "expand_spec".

mock/py/mockbuild/plugins/show_parsed_spec.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, plugins, conf, buildroot):
+        self.buildroot = buildroot
+        self.opts = conf
+        self.config = buildroot.config
+
+        # actually run our plugin at this step
+        plugins.add_hook("pre_srpm_build", self._PreBuildHook)
+
+    @traceLog()
+    def _PreBuildHook(self, host_chroot_spec, host_chroot_sources):
```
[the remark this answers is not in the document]
praiskup: Have you considered moving it post build, post deps, or others?  Any pros and cons?

This feature seems actually so convenient.  Does it deserve its own command line option?

This comment is mostly my curiosity, and attempt to start a discussion; not a call for PR update.

praiskup:
I :heart:  the approach (both that we lock rpm parser into the chroot for security reasons, and it actually expands the specfile _the only right way_).  Thank you for working on this!

mock/py/mockbuild/plugins/show_parsed_spec.py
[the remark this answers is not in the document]
azhuzhu: I firstly tried `prebuild` and the checked other available hooks and finally decided to use `pre_srpm_build`. The biggest pro of `pre_srpm_build` is that `host_chroot_spec` has been provided, so we don't have to search it again. And another pro is that the output is with `buildsrpm` which is more general that binary rpm build.

And adding `rpmspec_opts` is only to enable verbose output, or in case the possibility to add/modify MACROs...

mock/py/mockbuild/plugins/show_parsed_spec.py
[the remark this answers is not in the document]
azhuzhu: updated  to  `expanded-spec.txt`

mock/py/mockbuild/plugins/show_parsed_spec.py
```diff
@@ -0,0 +1,45 @@
...
+        getLog().info("enabled show_parsed_spec plugin")
+        rootdir_prefix = self.buildroot.make_chroot_path()
+        chroot_spec = host_chroot_spec.replace(rootdir_prefix, '')
+        chroot_sources = host_chroot_sources.replace(rootdir_prefix, '')
+        parsed_specfile = self.buildroot.resultdir + '/' + os.path.basename(chroot_spec)[:-4] + 'parsed.spec'
+
+        cmd = ["/usr/bin/rpmspec", "--parse", chroot_spec] + self.opts.get("rpmspec_opts", [])
+        output, _ = self.buildroot.doChrootPlugin(cmd, cwd=chroot_sources, returnOutput=True)
+        with open(parsed_specfile, "w", encoding="utf-8") as o:
+            o.write(output)
```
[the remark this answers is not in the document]
azhuzhu: This seems a kind of AI hallucination

mock/py/mockbuild/config.py
azhuzhu: updated to expand_spec

praiskup:
Thank you for the update.

> seems initialized, but...

I wanted to say that the plugin doesnt really generate any output  file for me (for fedora-rawhide-x86_64).  Am i doing something wrong?

mock/py/mockbuild/plugins/expand_spec.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @traceLog()
+    def _PreSRPMBuildHook(self, host_chroot_spec, host_chroot_sources):
+        """pre_srpm_build hook to expand specfile by ``rpmspec --parse``."""
+        self.logger.info("Executing expand_spec plugin")
+        rootdir_prefix = self.buildroot.make_chroot_path()
+        chroot_spec = host_chroot_spec.replace(rootdir_prefix, '')
+        chroot_sources = host_chroot_sources.replace(rootdir_prefix, '')
+        expanded_specfile = os.path.join(self.buildroot.resultdir, 'expanded-spec.txt')
+        self.logger.debug("Expanding original spec: %s to %s", chroot_spec, expanded_specfile)
```
praiskup: please s/debug/info/

azhuzhu: updated

praiskup:
> hmmm... it doesn't look like pre_srpm_build is a part of rebuild from SRPM

Exactly, see #1714 - I forgot to ping :-( sorry

mock/py/mockbuild/plugins/expand_spec.py
```diff
@@ -0,0 +1,48 @@
...
+        self.config = buildroot.config
+        self.logger = getLog()
+
+        # actually run our plugin at this step
+        plugins.add_hook("pre_srpm_build", self._PreSRPMBuildHook)
+        self.logger.info("expand_spec: initialized")
+
+    @traceLog()
+    def _PreSRPMBuildHook(self, host_chroot_spec, host_chroot_sources):
+        """pre_srpm_build hook to expand specfile by ``rpmspec --parse``."""
```
praiskup: See this: https://github.com/rpm-software-management/mock/blob/077ae471c014da668d2c1c62190848a0636716d7/mock/py/mockbuild/backend.py#L772-L774

If dynamic build deps are employed, mock (re)generates source.rpm a bit later (just so rpm -qpR *.src.rpm contains also those dynamic entries).

Nah.  We expand the specfile on multiple places within mock, and it's not clear now .... when _you_ need to expand it?

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -371,7 +371,7 @@ def _PostYumHook(self) -> None:
                     self.get_buildrequires(srpm.path)
 
     @traceLog()
-    def _PostDepsHook(self) -> None:
+    def _PostDepsHook(self, spec: str) -> None:  # pylint: disable=unused-argument, invalid-name
```
praiskup: Just use `_spec`

praiskup: Or like `_PostDepsHook(self, *_, **__)` ... if we want to be able to add more aruguments in the future.

mock/py/mockbuild/plugins/buildroot_lock.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, plugins, conf, buildroot):
         self.inst_done = False
         plugins.add_hook("postdeps", self.produce_lockfile)
 
-    def produce_lockfile(self):
+    def produce_lockfile(self, spec):  # pylint: disable=unused-argument
```
praiskup: use `_spec` (linter would be happy, no need to `disable=unused-argument`).

mock/py/mockbuild/plugins/export_buildroot_image.py
```diff
@@ -55,7 +55,7 @@ def _produce_image_as_root(self):
                  "oci-archive:" + tarball])
         self.do(["buildah", "rm", name])
 
-    def produce_buildroot_image(self):
+    def produce_buildroot_image(self, spec):  # pylint: disable=unused-argument
```
praiskup: Same here, `_spec`.

mock/py/mockbuild/plugins/package_state.py
```diff
@@ -63,7 +63,7 @@ def _availablePostYumHook(self):
                 self.state.finish("Outputting list of available packages")
 
     @traceLog()
-    def _installedPreBuildHook(self):
+    def _postDepsHook(self, spec):  # pylint: disable=invalid-name, unused-argument
```
praiskup: and here

mock/py/mockbuild/plugins/expand_spec.py
azhuzhu: It's not important for my usecase so far, because only Sources and Patches are what I want to expand.
And I think specfile is immutable in all stages and the macros should be unchanged after `buildrequrires` are installed. So it theoretically could be at any time after dependencies preparation.

Maybe they are all cases, if we have dynamic requires and/or rpmautospec involved?

praiskup (approved):

Issue #1667, opened 2025-12-08
Adapt Fedora Change: Dropping of cert.pem file


rezib:
I use latest version of mock from Fedora rawhide:

```console
[root@rpm ~]# rpm -qa | grep mock
mock-filesystem-6.5-1.fc44.noarch
mock-core-configs-43.3-1.fc44.noarch
mock-6.5-1.fc44.noarch
```

Creating the build environment for _opensuse-leap-16.0_ fails with SSL error:

```console
[root@rpm ~]# mock --init --root=opensuse-leap-16.0-x86_64
INFO: mock.py version 6.5 starting (python version = 3.14.0, NVR = mock-6.5-1.fc44), args: /usr/libexec/mock/mock --init --root=opensuse-leap-16.0-x86_64
Start(bootstrap): init plugins
INFO: selinux disabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux disabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start: clean chroot
Finish: clean chroot
Mock Version: 6.5
INFO: Mock Version: 6.5
Start(bootstrap): chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: /etc/mock/templates/opensuse-leap-16.0.tpl newer than root cache; cache will be rebuilt
INFO: enabled package manager cache
Start(bootstrap): cleaning package manager metadata
Finish(bootstrap): cleaning package manager metadata
INFO: Package manager dnf4 detected and used (fallback)
Start(bootstrap): installing dnf tooling
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
[…]
Finish(bootstrap): chroot init
Start: chroot init
INFO: calling preinit hooks
INFO: enabled root cache
INFO: enabled package manager cache
Start: cleaning package manager metadata
Finish: cleaning package manager metadata
INFO: enabled HW Info plugin
INFO: Package manager dnf4 detected and used (direct choice)
Start: installing minimal buildroot with dnf
Failed to set locale, defaulting to C
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
openSUSE Leap 16.0 - x86_64 - OSS                                                                                                                                                                              0.0  B/s |   0  B     00:00    
Errors during downloading metadata for repository 'opensuse-leap-oss':
[...]

PR #1711, opened 2026-02-12
Adapt to Changes/droppingOfCertPemFile

praiskup:
The change suggests us to use different bundle location:

| needs to preferably use the defaults of the library or if they must, | use the /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem file | instead.

Fixes: #1667


xsuchy (approved):
+1

PR #1724, opened 2026-03-09
Fix the tool check

tkopecek:
Previous code incorrectly looked to the host machine instead of bootstrap environment.


mock/py/mockbuild/installed_packages.py
```diff
@@ -116,7 +116,7 @@ def query_packages_location(packages, chrootpath=None,
     if 'yum' in dnf_cmd:
         # yum (yum-utils must be installed)
         repoquery_cmd = '/usr/bin/repoquery'
-        if not os.path.exists(repoquery_cmd):
+        if executor(["ls", "repoquery_cmd"]).strip() != repoquery_cmd:
```
praiskup: Can we drop the check entirely?

PR #1725, opened 2026-03-10
remove dependency on python3-dnf

xsuchy:
This module is going to be removed.

### DIFF
--- a/mock-core-configs/mock-core-configs.spec
+++ b/mock-core-configs/mock-core-configs.spec
@@ -29,11 +29,14 @@ Requires:   mock-filesystem
 
 Requires(post): coreutils
 # to detect correct default.cfg
-Requires(post): python3-dnf
-Requires(post): python3-hawkey
+# python3-libdnf5 - Fedora and RHEL10+
+# python3-dnf and python3-hawkey - older systems
+Requires(post): (python3-libdnf5 or (python3-dnf and python3-hawkey))
 Requires(post): system-release
 Requires(post): python3
+%if 0%{?mageia}
 Requires(post): sed
+%endif
 
 %description
 Mock configuration files which allow you to create chroots for Alma Linux,
@@ -99,7 +102,8 @@ fi
 if [ -s /etc/mageia-release ]; then
     mock_arch=$(sed -n '/^$/!{$ s/.* \(\w*\)$/\1/p}' /etc/mageia-release)
 else
-    mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
+    # the second command with hawkey can be removed when we stop supporting RHEL10
+    mock_arch=$(%{python3} -c 'import libdnf5 ; base = libdnf5.base.Base(); base.setup(); print(base.get_vars().get_value("basearch"))' 2>/dev/null || %{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
 fi
 
 cfg=unknown-distro
--- a/releng/release-notes-next/remove-dependency-python3-dnf.feature.md
+++ b/releng/release-notes-next/remove-dependency-python3-dnf.feature.md
@@ -1,0 +1,4 @@
+Previously we used python3-dnf to detect system architecture to guess default mock's config.
+As python3-dnf is going to be removed we now try to use function from python3-libdnf5. And if it fails we
+try to use the old function from python3-dnf.
+This should preserve functionality on both modern and old systems.


mock-core-configs/mock-core-configs.spec
```diff
@@ -99,7 +98,7 @@ fi
 if [ -s /etc/mageia-release ]; then
     mock_arch=$(sed -n '/^$/!{$ s/.* \(\w*\)$/\1/p}' /etc/mageia-release)
 else
-    mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
+    mock_arch=$(rpm --eval "%%{_arch}")
```
Conan-Kudo: This doesn't do the same thing. We need the basearch.

praiskup: ? 
```
python3 -c 'import libdnf5 ; base = libdnf5.base.Base(); base.setup(); print(base.get_vars().get_value("basearch"))'
```

xsuchy: This is better. I will update PR.

mock-core-configs/mock-core-configs.spec
```diff
@@ -99,7 +105,8 @@ fi
 if [ -s /etc/mageia-release ]; then
     mock_arch=$(sed -n '/^$/!{$ s/.* \(\w*\)$/\1/p}' /etc/mageia-release)
 else
-    mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
+    # the second command with hawkey can be removed when we stop supporting RHEL9
```
Conan-Kudo: Don't you mean when we stop supporting RHEL 10?

Conan-Kudo (changes requested):

mock-core-configs/mock-core-configs.spec
xsuchy: RHEL 10 has libdnf5 and python3-libdnf5, but... ops that is just internal tooling. Correct I will update the comment.

mock-core-configs/mock-core-configs.spec
```diff
@@ -29,11 +29,17 @@ Requires:   mock-filesystem
 
 Requires(post): coreutils
 # to detect correct default.cfg
+%if 0%{?fedora} || 0%{?rhel} > 9
+Requires(post): python3-libdnf5
```
praiskup:
```
Red Hat Universal Base Image 10 (RPMs) - BaseOS                                                                                                                                                             698 kB/s | 179 kB     00:00    
Red Hat Universal Base Image 10 (RPMs) - AppStream                                                                                                                                                          1.4 MB/s | 385 kB     00:00    
Red Hat Universal Base Image 10 (RPMs) - CodeReady Builder                                                                                                                                                   97 kB/s |  18 kB     00:00    
No match for argument: python3-libdnf5
Error: Unable to find a match: python3-libdnf5
```

mock-core-configs/mock-core-configs.spec
```diff
@@ -99,7 +105,8 @@ fi
 if [ -s /etc/mageia-release ]; then
     mock_arch=$(sed -n '/^$/!{$ s/.* \(\w*\)$/\1/p}' /etc/mageia-release)
 else
-    mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
+    # the second command with hawkey can be removed when we stop supporting RHEL10
+    mock_arch=$(%{python3} -c 'import libdnf5 ; base = libdnf5.base.Base(); base.setup(); print(base.get_vars().get_value("basearch"))' 2>/dev/null || %{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
```
praiskup, line 106: WDYT about a "dumb" RPM build-time condition?
```
%if 0%{?fedora} >= 42 || 0%{rhel} >= 10
mock_arch=$(%{python3} -c 'import libdnf5 ; base = libdnf5.base.Base(); base.setup(); print(base.get_vars().get_value("basearch"))'
%else
mock_arch=$(%{python3} -c "import dnf.rpm; import hawkey; print(dnf.rpm.basearch(hawkey.detect_arch()))")
%endif
```

Conan-Kudo: You don't know what distributions have it dnf5 vs dnf4 though.

mock-core-configs/mock-core-configs.spec
xsuchy: updated.

mock-core-configs/mock-core-configs.spec
xsuchy: +1 to @Conan-Kudo  It may start failing on Magiea, Suse and whatever. And it will require fix. What I put there is more bulletproof.

Conan-Kudo: It is kind of crappy that getting the basearch now requires setting up a base object though. We need to test to make sure that doesn't break while dnf itself is running.

praiskup: You say bulletproof, bu I can not help myself: #563.

praiskup: I'm not against merging this if you folks think it's a good idea!  But the fact that we do `aaa || bbb` and `aaa` may fail for many weird unexpected reasons (potentially not raising exceptions?), and if, `bbb` is likely failing too, I really find this brittle.  We are mixing up "build-time" & "installation-time" with a thing that should IMVHO really be moved to "run-time".

praiskup: > make sure that doesn't break while dnf itself is running

Reminds me a note from the packaging workshop that RPM/DNF is not reentrant (though this isn’t actually new here—we’ve been re-entering RPM for years in Mock's scriptlet).  Anyway, great point.

mock-core-configs/mock-core-configs.spec
```diff
@@ -29,11 +29,17 @@ Requires:   mock-filesystem
 
 Requires(post): coreutils
 # to detect correct default.cfg
+%if 0%{?fedora} || 0%{?rhel} > 10
+Requires(post): python3-libdnf5
+%else
 Requires(post): python3-dnf
 Requires(post): python3-hawkey
+%endif
```
Conan-Kudo: This should be a rich dep instead.

`Requires(post): (python3-libdnf5 or (python3-dnf and python3-hawkey))`

praiskup: Are rich deps RHEL 8+?

Conan-Kudo: Yes, they were introduced in RPM 4.14.

releng/release-notes-next/remove-dependency-python3-dnf.feature.md
```diff
@@ -0,0 +1,3 @@
+Previously we used python3-dnf to detect system architecture to guess default mock's config.
+As python3-dnf is going to be removed and python3-libdnf5 does not have this function we
+call directly rpm command to retrive the correct string.
```
Conan-Kudo: This release note is completely wrong now.

xsuchy: Updated.

mock-core-configs/mock-core-configs.spec
xsuchy: I tested it and you are right.
I updated this PR.

PR #1728, opened 2026-03-23
Decode percent-escaped file:// repo paths before bootstrap bind mount

Mr-Tao:
## Summary

When Mock scans package-manager config for host-local repositories to bind-mount into the bootstrap chroot, it currently strips the `file:///` prefix and uses the remaining string as a filesystem path directly.

That breaks valid file URIs containing percent-escaped path characters, for example paths generated by `Path(...).as_uri()`:

  - host path: `/tmp/workspace/job@2/results/bootstrap-repo`
  - file URI: `file:///tmp/workspace/job%402/results/bootstrap-repo`

Mock then checks `/tmp/workspace/job%402/results/bootstrap-repo` on the host, does not find it, skips the bind mount, and the bootstrap package manager later fails to read the local repository.

## Fix

Decode the local filesystem path with `urllib.parse.unquote()` after variable expansion and before the existence checks / bind mount setup.

This keeps existing behavior for plain absolute paths and unescaped `file://` URLs, while making percent-escaped file URIs work correctly.

## Tests

Added a regression test that uses `Path(repo_directory).as_uri()` for a path containing `@` and verifies that the repository is bind-mounted into the bootstrap chroot.

## Reproducer

A config like this should now work correctly:

```ini
[fedora]
baseurl = file:///tmp/workspace/job%402/results/bootstrap-repo
```

### DIFF
--- a/mock/py/mockbuild/package_manager.py
+++ b/mock/py/mockbuild/package_manager.py
@@ -8,6 +8,7 @@ import shutil
 import sys
 import time
 import re
+from urllib.parse import unquote
 from textwrap import dedent
 from configparser import ConfigParser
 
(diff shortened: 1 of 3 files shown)


praiskup:
Looks good, thank you!

Can you please provide a Towncrier snippet?

PR #1733, opened 2026-04-06
feat: add OCI platform support for x86_64 sub-architecture containers

### LINKED ISSUE #1732: x86_64_v2 configs disable bootstrap images due to missing OCI platform support
## Problem

Mock's Podman integration doesn't pass `--platform` when pulling container images. For standard architectures this isn't an issue — the registry returns the correct image. But for x86_64 sub-architecture variants (x86_64_v2, x86_64_v3, x86_64_v4), `podman pull` fetches the base amd64 image instead of the variant.

As a workaround, the AlmaLinux 10 and AlmaLinux Kitten 10 templates currently set `use_bootstrap_image = False` for x86_64_v2.

## Expected behavior

x86_64_v2/v3/v4 builds should use container-based bootstrap like every other architecture. `podman pull` should request the correct OCI image variant via `--platform linux/amd64/v2` (or v3/v4), and the architecture check should accept variant images on a matching base host.

## Affected configs

- `almalinux-10-x86_64_v2`
- `almalinux-kitten-10-x86_64_v2`
- Any future config using x86_64_v2, x86_64_v3, x86_64_v4 targets

andrewlukoshko:
Add oci_platform_map config dict mapping x86_64_v2/v3/v4 to OCI platform strings (linux/amd64/v2, v3, v4).

When target_arch has a mapping, podman pull receives --platform so the correct image variant is fetched from multi-arch manifests.

Update podman_check_native_image_architecture() with a keyword-only oci_platform parameter for variant-tolerant arch comparison — an amd64/v2 image is accepted on an x86_64 host.

Includes 11 unit tests covering platform injection, variant tolerance, backward compatibility, and error paths.

Fixes: #1732

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -773,8 +773,14 @@ def main():
         # Enforce host-native repo architecture for bootstrap chroot (unless
         # bootstrap_forcearch=True, which should never be the case).  This
         # decision affects condPersonality() for DNF calls!
+        #
+        # Exception: sub-architecture variants (e.g. x86_64_v2) listed in
+        # oci_platform_map run natively on the host — keep the target's
+        # repo_arch so the bootstrap uses the correct sub-arch repos.
         host_arch = config_opts["host_arch"]
-        if config_opts["use_bootstrap_image"]:
+        target_arch = config_opts["target_arch"]
+        if config_opts["use_bootstrap_image"] \
+                and target_arch not in config_opts.get('oci_platform_map', {}):
             # with bootstrap image, bootstrap is always native
             bootstrap_buildroot_config['repo_arch'] = config_opts['repo_arch_map'].get(host_arch, host_arch)
         elif host_arch not in config_opts.get("legal_host_arches", []) \
--- a/mock-core-configs/etc/mock/templates/almalinux-10.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-10.tpl
@@ -9,7 +9,6 @@ config_opts['bootstrap_image'] = 'quay.io/almalinuxorg/almalinux:10'
 # deal with special handling for x86_64_v2 variant
 config_opts['mirrorlist_arch'] = "{% if repo_arch == 'x86_64_v2' %}?arch=x86_64_v2{% endif %}"
 config_opts['baseurl_arch'] = "{% if repo_arch == 'x86_64_v2' %}x86_64_v2{% else %}$basearch{% endif %}"
-config_opts['use_bootstrap_image'] = "{% if repo_arch == 'x86_64_v2' %}False{% else %}True{% endif %}"
 
 config_opts['dnf.conf'] = """
 [main]
--- a/mock-core-configs/etc/mock/templates/almalinux-kitten-10.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-kitten-10.tpl
@@ -9,7 +9,6 @@ config_opts['bootstrap_image'] = 'quay.io/almalinuxorg/almalinux:10-kitten'
 # deal with special handling for x86_64_v2 variant
 config_opts['mirrorlist_arch'] = "{% if repo_arch == 'x86_64_v2' %}?arch=x86_64_v2{% endif %}"
 config_opts['baseurl_arch'] = "{% if repo_arch == 'x86_64_v2' %}x86_64_v2{% else %}$basearch{% endif %}"
-config_opts['use_bootstrap_image'] = "{% if repo_arch == 'x86_64_v2' %}False{% else %}True{% endif %}"
 
 config_opts['dnf.conf'] = """
 [main]
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -429,6 +429,13 @@ def setup_default_config_opts():
         'i686': 'i386',
     }
 
+    # mapping from target_arch to OCI --platform string for podman pull
+    config_opts['oci_platform_map'] = {
+        'x86_64_v2': 'linux/amd64/v2',
+        'x86_64_v3': 'linux/amd64/v3',
+        'x86_64_v4': 'linux/amd64/v4',
+    }
+
     config_opts["recursion_limit"] = 5000
 
     config_opts["calculatedeps"] = None
--- a/mock/py/mockbuild/podman.py
+++ b/mock/py/mockbuild/podman.py
@@ -117,7 +117,12 @@ class Podman:
         """ pull the latest image, return True if successful """
         logger = getLog()
         logger.info("Pulling image: %s", self.image)
-        cmd = [self.podman_binary, "pull", self.image]
+        cmd = [self.podman_binary, "pull"]
+        target_arch = self.buildroot.config.get('target_arch', '')
+        oci_platform = self.buildroot.config.get('oci_platform_map', {}).get(target_arch)
+        if oci_platform:
+            cmd += ["--platform", oci_platform]
+        cmd.append(self.image)
 
         res = subprocess.run(cmd, env=self.buildroot.env,
                              stdout=subprocess.PIPE, stderr=subprocess.PIPE,
--- a/mock/tests/test_podman.py
+++ b/mock/tests/test_podman.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+"""Unit tests for mockbuild.podman — OCI platform pull and architecture check."""
+
+import subprocess
+from unittest.mock import MagicMock, patch
+
+from mockbuild.podman import (
+    Podman,
+    podman_check_native_image_architecture,
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _make_buildroot(target_arch, oci_platform_map=None):
+    """Return a minimal mock buildroot with the given config values."""
+    config = {"target_arch": target_arch}
+    if oci_platform_map is not None:
+        config["oci_platform_map"] = oci_platform_map
+    br = MagicMock()
+    br.config = config
+    br.env = {}
+    return br
+
+
+def _make_podman(buildroot, image="registry.example.com/image:latest"):
+    """Instantiate Podman while bypassing the os.path.exists check."""
+    with patch("os.path.exists", return_value=True):
+        return Podman(buildroot, image)
+
+
+# ===================================================================
+# R002 — pull_image injects --platform when mapping matches
+# ===================================================================
+
+class TestPullImagePlatform:
+    """pull_image() must inject --platform from oci_platform_map."""
+
+    @patch("mockbuild.podman.subprocess.run")
+    def test_platform_injected_when_mapped(self, mock_run):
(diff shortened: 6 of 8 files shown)


Conan-Kudo (approved):

mock/py/mock.py
```diff
@@ -773,8 +773,14 @@ def main():
...
         # decision affects condPersonality() for DNF calls!
+        #
+        # Exception: sub-architecture variants (e.g. x86_64_v2) listed in
+        # oci_platform_map run natively on the host — keep the target's
+        # repo_arch so the bootstrap uses the correct sub-arch repos.
         host_arch = config_opts["host_arch"]
-        if config_opts["use_bootstrap_image"]:
+        target_arch = config_opts["target_arch"]
+        if config_opts["use_bootstrap_image"] \
+                and target_arch not in config_opts.get('oci_platform_map', {}):
```
praiskup, line 783: Late to the party :(  Is this really correct?

The point is to use "the native" bootstrap tooling, for installing cross-arch chroot.  IOW, if the old code works for cross-arch use-cases, why do we have to change it for cross-sub-arch?

What would happen if we haven't done this?

Conan-Kudo: We would not be able to fetch the correct image. The way subarches work is especially annoying, since there is no detection support in the OCI tooling currently. You have to "know" and pull the right one.

praiskup:
Can you please help me with the reproducer of the problem that needed this patch?

andrewlukoshko:
@praiskup if your build host CPU is not x86_64_v3 compatible then you can't build for AlmaLinux-10-x86_64_v2 using regular AlmaLinux 10 x86_64 bootstrap image.

mock/py/mock.py
praiskup: Why do we need a specific (or "correct") sub-architecture *in bootstrap*, especially if we say "bootstrap_ready=true"?  We don't need that for cross-compilations either.

praiskup:
>  if your build host CPU is not x86_64_v3 compatible

I probably don't have access to such a box :shrug: but

>  then you can't build for AlmaLinux-10-x86_64_v2 

Can you elaborate on why?

andrewlukoshko:
@praiskup 
> I probably don't have access to such a box

you could start KVM machine and set pre-2013 CPU model like Nehalem.

> Can you elaborate on why?

Because glibc, libdnf and everything else in "regular" x86_64 bootstrap container are built for x86_64-v3 and can't run on pre-2013 CPUs. So we're getting into weird situation where you need new CPU to build package for old CPU.
This pull request fixes this. So basically you can build for AlmaLinux-10-x86_64_v2 on any EL8 or EL9 system on older CPUs.

praiskup:
> Because glibc, libdnf and everything else in bootstrap container are built for x86_64-v3 and can't run on pre-2013 CPUs.

Do I get it right that Podman is broken?  That it downloads invalid variant of the image by default?

andrewlukoshko:
@praiskup I can't say it's BROKEN because it's correct behavior, but it by default pulls image for current arch which is "x86_64", so it translated into "linux/amd64" platform. So this PR adds `--platform=linux/amd64/v2` **only if x86_64_v2 target arch is selected**, the same for **v3** and **v4**.
And nothing happens if we operate with regular arches like x86_64, aarch64, ppc64le, etc.

praiskup:
I am probably still missing something :-(

> but it by default pulls image for current arch which is "x86_64", so it translated into "linux/amd64" platform.

That is exactly what Mock expects. The only requirement is that Mock can execute dnf/rpm commands inside a container started from this image. For example, if Mock is running on an x86_64-v3 host, the bootstrap image can (should!) also be v3, even if the build target is v2.  There's no point in enforcing v2, same as if you (cross)compile for s390x you don't force Podman to pull s390x bootstrap image.

andrewlukoshko:
@praiskup Now imagine you're running mock on old (x86_64-v2 only) EL9 host and try to build package for AlmaLinux 10 x86_64_v2.

So your host's arch is x86_64, Podman downloads regular AlmaLinux 10 x86_64 (linux/amd64) container image and it requires x86_64-v3, so dnf from it can't run on your x86_64-v2 host.

So **if your host is x86_64-v3 compatible - you're fine even without this PR**. But without this PR you're forcing users to use x86_64-v3 hosts even if they just want to build for x86_64-v2. This doesn't make sense.
And you cannot disable using bootstrap image completely as a workaround because EL9's rpm/dnf have no idea about x86_64_v2 arch.

Conan-Kudo:
@praiskup The problem is that RHEL did the wrong thing and marked x86_64-v3 as x86_64. This means that _nothing_ for automatically detecting this works properly. Not in RPM, not in libsolv, and certainly not in Mock.

For AlmaLinux, the added complication is that we have x86_64-v2 supported with the correct architecture name, and we use the variants correctly, but unless you force it, users cannot bootstrap or use Mock correctly on those platforms.

This pull request effectively fixes this by working around the breakage caused by x86_64-v3 calling itself x86_64 in RHEL.

Conan-Kudo:
> > Because glibc, libdnf and everything else in bootstrap container are built for x86_64-v3 and can't run on pre-2013 CPUs.
> 
> Do I get it right that Podman is broken? That it downloads invalid variant of the image by default?

Yes, this is correct, because the images are not defined correctly for Podman to do the right thing.

praiskup:
How I read your comments: On an x86_64_v2 host, `podman pull` downloads v3 images, even though there are v2 alternatives. Why are we fixing Mock instead of Podman for the v2 architecture?

Conan-Kudo:
Because Podman can't know better. `amd64` is supposed to be valid, but the wrong content is included.

praiskup:
How comes podman doesn't know it was compiled for v2, to use adequate `--platform`?  Instead, everybody needs to use --platform explicitly?

Conan-Kudo:
Because it's not yet specified how to handle it in OCI: https://github.com/containers/podman/discussions/15256

praiskup:
> Because it's not yet specified how to handle it in OCI

There is the `--platform` option that Podman (built on v2 system) can use.
Why we can't patch Podman on Alma v2?

Conan-Kudo:
Because it doesn't work anywhere else? The _point_ of Mock is that things are supposed to work cross-distro. It's just silly if we make a fix that only works in AlmaLinux when most people would be using Fedora or some other distribution to build packages targeting AlmaLinux.

praiskup:
You mean anywhere else **on x86_64_v2**?  If we fixed Podman on v2 to select proper images, no other clients have to make any hacks, or?

This patch is just an ugly (and hopefully temporary) workaround for a broken Podman **there, on x86_64_v2**.

Not just **we in Mock**, but anybody on **x86_64_v2** needs to think about v2 image selection explicitly, and *that's the real problem*.  The point I'm trying to make is that we can't patch the world (including Mock) just to work-around a Podman bug.

Conan-Kudo:
Then get the Red Hat container team to fix it in Podman. Which requires getting it specified in OCI. It was _Mock's choice_ to hang its hat on the OCI ecosystem. None of us are able to fix anything in the OCI space. We can happily drop the forced platform flag stuff once Skopeo and Podman are fixed, but until it is, this is the way it needs to work.

Conan-Kudo:
Also, you forget that this would be required even if Podman had auto-detection, because Podman's detection would be based on the detected CPU. So if you wanted to force v2 proper builds on a v3 system, you'd need the override _anyway_.

praiskup:
> you'd need the override anyway

That is clearly what I’m failing to explain. Mock’s only requirement is that `podman pull` does the 'right thing' automatically. Since Mock [downloads the 'default'](https://github.com/rpm-software-management/mock/blob/e438e45d27d8e4516eee4392ac49fd87f560e283/mock/py/mockbuild/podman.py#L62-L91) x86_64 bootstrap images even for s390x builds, it feels very inconsistent that we now have to manually handle these x86_64_v2 nuances.

Conan-Kudo:
Technically, we need to handle it for x86_64 variants and RISC-V variants, though we have no RISC-V anything to force that for us yet.

In the context of native architecture, we need flags for RHEL chroots to block them on incompatible CPUs that are impossible to detect from the image itself.

Something like [x86-64-level](https://github.com/HenrikBengtsson/x86-64-level) would be something we could depend on for environments where you can't ask the link loader for this information, and then we can reliably detect x86_64 levels and propagate them for bootstrap image selection or valid x86_64 chroots.

praiskup:
> Technically, we need to handle it for x86_64 variants and RISC-V variants, though we have no RISC-V anything to force that for us yet.

We don't have to: Mock should always just ask the tooling (Podman/OCI) to download native variant of the image.  If we really have to work-around Podman's missing feature or a bug, there's the `bootstrap_image` option for that (specific image tag might be set in an architecture-specific configuration file, instead of "once for all" in the template).

praiskup:
@Conan-Kudo I am sorry to say this, but 5771350906735850065b880ffa6b4c64cb000be7 shouldn't get into the next release, so I tend to revert.  Is it OK to specify bootstrap_image value per-architecture in almalinux specs?  What would be the image names?

Conan-Kudo:
> @Conan-Kudo I am sorry to say this, but [5771350](https://github.com/rpm-software-management/mock/commit/5771350906735850065b880ffa6b4c64cb000be7) shouldn't get into the next release, so I tend to revert. Is it OK to specify bootstrap_image value per-architecture in almalinux specs? What would be the image names?

It is not okay to revert, because you haven't actually come up with a real solution. You can't use image names alone to pull it, or otherwise we would have done that. You cannot select architecture variants manually by using a label. The `--platform` argument has to be passed to force it.

praiskup:
The real solution is to fix Podman so it handles the selection appropriately. The cross-arch logic is incorrect and goes against how Mock bootstrapping normally works. This patch shouldn't have been merged in the first place :-( It would be nice to give us a chance to review it next time (with a 24h+ grace period). I can't release it in its current state.

> You can't use image names alone to pull it

Could you please elaborate on why this is impossible? If a system with architecture `<foo>` runs `podman pull <multi-platform-image-reference>` and doesn't download the correct (available!) platform variant, that’s a bug. This serious problem needs to be solved in Podman itself, not worked around in every single tool that uses it. Besides, the affected distribution (Alma only?) already has image tags and image names to work around the issue (separate the images), rather than patching all the tooling.

Conan-Kudo:
AlmaLinux is the only RPM distribution right now, yes (but not the only distribution period), but my precondition at this point is that if you want this out, then Podman needs to do the right thing by default. It currently cannot. And you cannot use an image reference to request an arch variant either, as it's only possible to specify platform variants with `--platform`.

I don't diagree that we want to prefer using native for bootstrap, but we're stuck between a rock and a hard place right now because of how broken things are right now.

praiskup:
> And you cannot use an image reference to request an arch variant either

We two can't change this ourselves, maybe @andrewlukoshko?  As it's AlmaLinux's policy on how they tag `x86_v<x>` images. Frankly, from their perspective, it also seems much more user-friendly to simply (temporarily) document _please pull almalinux:10_v2 if you're on v2_, rather than teaching users to use the --platform flag explicitly.

PR #1734, opened 2026-04-09
configs: set bootstrap_image_ready=True for AlmaLinux configs

andrewlukoshko:
Set bootstrap_image_ready = True for almalinux-8, almalinux-9, almalinux-10,
and almalinux-kitten-10 templates.

AlmaLinux container images now have python3-dnf-plugins-core available
by default, so there is no need to update the bootstrap chroot after
extracting the image.

See AlmaLinux/container-images#88


Conan-Kudo (approved):

Conan-Kudo (changes requested):
Please add a towncrier changelog snippet.

Conan-Kudo (approved):

xsuchy (approved):
+1

PR #1736, opened 2026-04-15
Add config for Fedora 44 version of RISC-V port

hrw:
As we started building Fedora 44 it would be good to have config available for it.


praiskup (approved):

xsuchy (approved):
+1

hrw:
Will there be a new version of `mock-core-configs` with it?

xsuchy:
Yes, we are trying to stick to about 2 month cadence. And it's about the time.

hrw:
Flock is coming, there will be a talk about risc-v port and mock will be mentioned. With info how to use it on any architecture.

So having f44 config would be nice.

I missed that lack as I had a copy of config done very early.

Issue #1243, opened 2023-10-12
Better CPU/MEM monitoring during the build (sampling?)


praiskup:
Sometimes happens that some (a bit more complicated) Mock build fails in Copr with "Killed" reason.  So I take a look at the [hw_info.log.gz](https://download.copr-dev.fedorainfracloud.org/results/@copr/powerful-test/fedora-39-aarch64/02915634-kernel/hw_info.log.gz) and it doesn't help -> the problem is that we take the info before we actually start the build.

It would be awesome to have support for periodic sampling, perhaps within the same module.  So we could analyze post-mortem how the memory/swap/storage behaved.  Is this worth doing in Mock, or should we do this on Copr builder?

Conan-Kudo:
Would something like [`below`](https://fedoramagazine.org/below-a-time-traveling-resource-monitor/) work for this?

cc: @davide125

praiskup:
Nice!  That is exactly the tool we need.  This looks like a tool for Copr builders, rather than Mock.  WDYT?

Conan-Kudo:
I think it could go either way, depending on how you want to collect, present, and archive the data. A Mock plugin would allow for people to have this information locally when they're trying to figure out build performance bottlenecks. Copr, as you can guess, would benefit from it for gathering data about builder performance.

stewartsmith:
With `systemd-nspawn` we should be able to get statistics out of `systemd-cgtop` (or what backs it... even just the raw cgroup data from the kernel) and thus get a decent idea of actual usage?

In the back of my mind it had been kicking around as an idea to try and more accurately size koji builders - we tend to brute force things a bit, and there's certainly opportunities for non-trivial cost savings in the future.

PR #1748, opened 2026-04-25
RFC: Add a system monitor plugin

yaneti:
Based on the kernel memcg cgroup2 memory accounting and its systemd-nspawn intergration

When enabled produces output like:
`MOCKMA: Total Memory Peak 52309.39 MiB | Top process: RSS:10975.77 MiB [/usr/libexec/gcc/x86_64-redhat-linux/16/cc1plus -quiet -I /build... ` 
<full command line> at the end of the build.log

With assistance from Gemini AI

### DIFF
--- a/docs/Plugin-SystemMonitor.md
+++ b/docs/Plugin-SystemMonitor.md
@@ -1,0 +1,25 @@
+---
+layout: default
+title: Plugin SystemMonitor
+---
+
+This plugin activates per-interval collection of various statistics
+based on the kernels cgroupv2 controllers during the build phase and
+dumps a json file 'system_monitor.json' with the collected statistics
+in the result dir
+
+Currently dumped statistics include total maximum memory usage for the build
+and the process with maximum memory RSS
+
+The plugin requires the use of systemd-nspawn as build container runner
+
+## Configuration
+
+The module is disabled by default and needs to be activated by:
+
+    config_opts['plugin_conf']['system_monitor_enable'] = True
+
+The following sub-options may be specified:
+
+    # the interval between statistics collection runs in seconds, default 2
+    config_opts['plugin_conf']['system_monitor_opts']['interval'] = 10
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -397,6 +397,11 @@
 # certain file paths being accessed.
 #config_opts['plugin_conf']['unbreq_opts']['exclude_accessed_files'] = []
 
+### system_monitor plugin disabled by default
+# config_opts['plugin_conf']['system_monitor_enable'] = False
+## statistics collection interval in seconds
+# config_opts['plugin_conf']['system_monitor_opts']['interval'] = 2
+
 #############################################################################
 #
 # environment for chroot
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -33,7 +33,7 @@ PLUGIN_LIST = ['tmpfs', 'root_cache', 'yum_cache', 'mount', 'bind_mount',
                'lvm_root', 'compress_logs', 'sign', 'pm_request',
                'hw_info', 'procenv', 'showrc', 'rpkg_preprocessor',
                'rpmautospec', 'buildroot_lock', 'export_buildroot_image',
-               'unbreq', 'expand_spec']
+               'unbreq', 'expand_spec', 'system_monitor']
 
 def nspawn_supported():
     """Detect some situations where the systemd-nspawn chroot code won't work"""
@@ -264,6 +264,10 @@ def setup_default_config_opts():
         'expand_spec_opts': {
             'rpmspec_opts': [],
         },
+        'system_monitor_enable': False,
+        'system_monitor_opts': {
+            'interval' : 2
+        }
     }
 
     config_opts['environment'] = {
--- a/mock/py/mockbuild/plugins/system_monitor.py
+++ b/mock/py/mockbuild/plugins/system_monitor.py
@@ -1,0 +1,171 @@
+# License: GPL2 or later see COPYING
+"""
+Track various system statistics during the build phase.
+- Report maximum allocated memory in total and for the "biggest" process
+
+The plugin requires systemd-nspawn as build container runner
+"""
+
+import threading
+import os
+import json
+import backoff
+
+from mockbuild.trace_decorator import getLog
+from mockbuild.util import get_machinectl_uuid, _safe_check_output, USE_NSPAWN
+
+requires_api_version = "1.1"
+run_in_bootstrap = False
+
+def init(plugins, conf, buildroot):
+    """ Plugin entry point """
+    SystemMonitor(plugins, conf, buildroot)
+
+class SystemMonitor:
+    """ Main plugin class """
+
+    def get_top_process_info(self, scope_path):
+        """Finds the process in the cgroup with the highest RSS.
+
+        Args:
+            scope_path (str): Path to the cgroup scope directory.
+
+        Returns:
+            tuple[int, str]: A tuple containing the RSS of the top process in bytes
+                and its command line.
+        """
+        procs_path = os.path.join(scope_path, "cgroup.procs")
+        max_rss = 0
+        max_cmdline = ""
+
+        try:
+            if not os.path.exists(procs_path):
+                return (0, "unknown")
+            with open(procs_path, 'r', encoding="utf-8") as f:
+                pids = f.read().split()
+
+            for pid in pids:
+                try:
+                    with open(f"/proc/{pid}/statm", 'r', encoding="utf-8") as sm:
+                        data = sm.read().split()
+                        if not data:
+                            continue
+                        # RSS in pages * PAGE_SIZE bytes
+                        rss_bytes = int(data[1]) * os.sysconf('SC_PAGE_SIZE')
+
+                    if rss_bytes > max_rss:
+                        max_rss = rss_bytes
+                        with open(f"/proc/{pid}/cmdline", 'r', encoding="utf-8") as cmd:
+                            max_cmdline = cmd.read().replace('\0', ' ').strip()
+                        if not max_cmdline:
+                            max_cmdline = f"{pid}"
+
+                except (FileNotFoundError, ProcessLookupError, IndexError):
+                    continue
+        except OSError as e:
+            getLog().error("SYSMON: Error: %s", e)
+
+        return (max_rss, max_cmdline)
+
+    @backoff.on_predicate(backoff.constant, jitter=None, interval=2, max_time=120)
+    def get_machine_id(self, buildroot):
+        """ Retry getting machine id until nspawn starts """
+        return get_machinectl_uuid(buildroot.make_chroot_path())
+
+    def sysmon_thread(self, buildroot, interval):
+        """ Main monitoring thread """
+        current_peak = 0
+
+        machine_id = self.get_machine_id(buildroot)
+        if machine_id is None:
+            getLog().error("SYSMON: Failed to get nspawn container machine_id")
+            return
+
+        getLog().debug("SYSMON: Collecting data from machine_id: %s", machine_id)
+
+        # The unit name depends on systemd version
+        ustr = _safe_check_output(["/bin/machinectl", "show", "--property=Unit", f"{machine_id}"])
+        if isinstance(ustr, bytes):
+            ustr = ustr.decode("utf-8")
+        machine_id_unit = ustr.rstrip().split('=')[1]
+        scope_dir = f"/sys/fs/cgroup/machine.slice/{machine_id_unit}"
+        peak_file = os.path.join(scope_dir, "memory.peak")
+
+        while not self.sysmon_stop_event.is_set():
+            max_status = "Current"
+            pid_status = "Current"
+
+            try:
+                if os.path.exists(peak_file):
+                    with open(peak_file, 'r', encoding="utf-8") as f:
+                        current_peak = int(f.read().strip())
+
+                if current_peak > self.max_memory_peak:
+                    max_status = "NEW"
+                    self.max_memory_peak = current_peak
+
+                top = self.get_top_process_info(f"{scope_dir}/payload")
+                if top[0] > self.top_rss[0]:
+                    pid_status = "NEW"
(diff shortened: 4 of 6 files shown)


mock/py/mockbuild/plugins/memory_accounting.py
```diff
@@ -0,0 +1,140 @@
...
+        except (IOError, ValueError):
+            getLog().debug(f"MOCKMA: memory.peak missing {scope_dir}")
+            pass
+
+        if ma_stop_event.wait(timeout=interval):
+            break
+
+
+
+class MemoryAccounting(object):
```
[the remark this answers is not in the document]
yaneti: I coudn't make the thread run with `ma_check_proc_mem` being a class method, for some reason. Didn't dig deeper..

davidlt:
I was thinking about this more, and two things that popped into mind:
1. The current format is not machine friendly to parse this information thus harder to write tools to ingest it.
2. Related to the 1st one. Sometimes our `build.log` goes into gigabytes thus finding this information again seems to be expensive.

It would be nicer to have this as a separate file with maybe a JSON content. In that case we could have a tooling that could check historical host + this output to figure out future scheduling opportunities.

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -0,0 +1,140 @@
...
+                    continue
+        except Exception as e:
+            getLog().error("MOCKMA: Error: %s", e)
+        return top_cmdline, max_rss
+
+    def ma_check_proc_mem(self, buildroot, interval):
+        current_peak = 0
+
+        # wait for nspawn to start
+        time.sleep(5)
```
xsuchy: Can this be faster by pooling something?

xsuchy:
In general, I welcome this work.

I agree that separate file would be better. Plus one to JSON.

Please add documentation. Take a look at https://rpm-software-management.github.io/mock/Plugin-PackageState which you can mimic.

Add configuration option to enable/disable it. Before this mature, it should be disabled by default.

Reviews from Claude:

  1. Thread safety — class-level mutable state (bug)                                                                                                                                         
  
  Lines 16-19 define tracking variables as class attributes, not instance attributes. This means all instances share the same state. If two builds run concurrently (or a plugin is          
  re-instantiated), they'll clobber each other's data. These should be initialized in __init__:

```                                                                                                                                                                                             
  # Current (shared across all instances):
  class MemoryAccounting(object):                                                                                                                                                            
      max_memory_peak = 0                                                                                                                                                                    
      top_rss = 0                                                                                                                                                                            
      ...                                                                                                                                                                                    
                                                                                                                                                                                             
  # Should be:                                                                                                                                                                               
  def __init__(self, plugins, buildroot):                                                                                                                                                    
      self.max_memory_peak = 0                                                                                                                                                               
      self.top_rss = 0                                                                                                                                                                       
      self.top_rss_cmd = ""                                                                                                                                                                  
      self.ma_stop_event = threading.Event()                                                                                                                                                 
      self.ma_timer_thread = None
```

  2. No thread.join() in _on_post_build (bug)                                                                                                                                                
                                                                                                                                                                                             
  _on_post_build sets the stop event but never calls self.ma_timer_thread.join(). The monitoring thread could still be running when the results are logged, leading to a race where the final
   values are read while still being updated. Also, if mock exits quickly, the dangling thread could prevent clean shutdown. The thread should also be set as daemon=True as a safety net.
                                                                                                                                                                                             
  3. Missing use_nspawn guard (bug)

  The plugin unconditionally injects --property=MemoryAccounting=on into nspawn_args and calls machinectl. This will fail (or be meaningless) when mock is configured with use_nspawn = False
   (chroot mode). The plugin should check self.config.get('use_nspawn') and either skip activation or log a warning.
My note: we should probably completly skip this plug if nspawn is not used.

                                                                                                                                                                                             
  4. get_machinectl_uuid can return None (bug)                                                                                                                                               
  
  If get_machinectl_uuid() returns None (documented as "we should never get here" in util.py but possible), the subsequent machinectl show call will fail with an unhelpful error. Should    
  check for None and bail with a clear log message.

  7. Missing run_in_bootstrap declaration                                                                                                                                                    
                  
  Most plugins declare run_in_bootstrap = False at module level. This plugin omits it.  

  8. init() ignores conf parameter                                                                                                                                                           
                  
  The init() function receives conf but doesn't pass it to the constructor. Other plugins (hw_info, procenv, ccache, etc.) store conf as self.*_opts for plugin-specific configuration. The  
  interval config is fetched in _on_pre_build via a roundabout self.config['plugin_conf']['memory_accounting_opts'] path — using conf directly would be cleaner and consistent.
                                                                                                                                                                                             
  9. Missing license header

  All other plugin files include a license header (# License: GPL2 or later see COPYING). 


  Style / minor                                                                                                                                                                              
                  
  - Three blank lines (lines 107-109) before _on_pre_build — should be two per PEP 8 (one within a class).
  - if (isinstance(ustr, bytes)) — parentheses around the condition are unnecessary and inconsistent (though util.py has the same pattern).                                                  
  - "unknown" returned for errors — consider using the process PID as fallback when cmdline can't be read (some kernel threads have empty cmdlines).

xsuchy:
And please add release notes entry https://github.com/rpm-software-management/mock/blob/main/docs/Release-Notes-New-Entry.md

yaneti:
Tried to address most of the issues.
Since @davidlt mentioned that cgroup2 controller based statistic can be gotten for many more things than max memory perhaps now 's the time to rename the thing to   build_stats / nspawn_stats / mock_stats / whatever_stats
and make it a placeholder for  other per-interval collection from the build process.
This would all for collecting data on memory pressure, cpu load, io load etc...

I'd like to leave the naming to whomever feels inspired ;)

mock/py/mockbuild/plugins/system_monitor.py
yaneti: I don't really know..ideas ?

xsuchy:
+1 to cpu sampling. That was already suggested in https://github.com/rpm-software-management/mock/issues/1243

xsuchy:
Naming ideas:
* system-monitor
* system-accounting
* system-top

davidlt:
I also mentioned on Matrix that there seems to be statistics for swap usage too. That seems to be ` memory.swap.peak` from cgroupsv2.

yaneti:
So its renamed.
The json format is completely made up and I am not sure exactly what to collect for swap or cpu statistics , so that'll have to be added by someone else.

xsuchy:
@yaneti, can you please fix the lint warnings?

yaneti:
> @yaneti, can you please fix the lint warnings?

I tried, but my python foo is weak.

The current 5 second wait for the nspawn to start actually breaks with %generate_buildrequires.. so it still needs solution for how to figure when it has started.

yaneti:
Ugh, this last version catches the prep nspawn, not the build one...

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -0,0 +1,167 @@
...
+        self.max_memory_peak = 0
+        self.top_rss = (0, "")
+        self.sysmon_timer_thread = None
+        self.sysmon_stop_event = threading.Event()
+        self.buildroot = buildroot
+        self.system_monitor_opts = conf
+        self.config = buildroot.config
+
+        if self.config.get('use_nspawn', True):
+            getLog().warning("SYSMON: build is not using nspawn. Statistics will not be available")
```
praiskup, line 166: Can you use `util.USE_NSPAWN` instead?  This option is obsoleted and probably doesn't work properly.

praiskup: the logic seems inverted

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -0,0 +1,167 @@
...
+        getLog().debug("SYSMON: Monitoring thread started via callback.")
+
+    def _on_postbuild(self):
+        self.sysmon_stop_event.set()
+        self.sysmon_timer_thread.join()
+        getLog().info(
+            "SYSMON: Total Memory Peak %.2f MiB | Top process: RSS:%.2f MiB [%s]",
+            self.max_memory_peak / 1048576, self.top_rss[0] / 1048576, self.top_rss[1]
+        )
+        out_file = self.buildroot.resultdir + '/system_monior.json'
```
praiskup: typo

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -0,0 +1,167 @@
...
+            tuple[int, str]: A tuple containing the RSS of the top process in bytes
+                and its command line.
+        """
+        procs_path = os.path.join(scope_path, "cgroup.procs")
+        max_rss = 0
+        max_cmdline = ""
+
+        try:
+            if not os.path.exists(procs_path):
+                return "unknown", 0
```
praiskup: (str, int) returned vs. (int, str) on 68?

releng/release-notes-next/system-monitor-plugin.feature.md
```diff
@@ -0,0 +1 @@
+New system_memory plugin for collecting various statistics in the build phase
```
praiskup: typo: system_monitor

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def sysmon_thread(self, buildroot, interval):
+        """ Main monitoring thread """
+        current_peak = 0
+        machine_id = None
+
+        # wait for nspawn to start
+        while machine_id is None:
+            time.sleep(2)
+            machine_id = get_machinectl_uuid(buildroot.make_chroot_path())
```
praiskup: Please use backoff library with some reasonable limit (like 2 minutes).

mock/py/mockbuild/config.py
```diff
@@ -33,7 +33,7 @@
                'lvm_root', 'compress_logs', 'sign', 'pm_request',
                'hw_info', 'procenv', 'showrc', 'rpkg_preprocessor',
                'rpmautospec', 'buildroot_lock', 'export_buildroot_image',
-               'unbreq', 'expand_spec']
+               'unbreq', 'expand_spec','system_monitor']
```
praiskup: missing space

mock/py/mockbuild/plugins/system_monitor.py
praiskup: bot suggests: Should check self.sysmon_stop_event in this loop.

mock/py/mockbuild/config.py
```diff
@@ -264,6 +264,7 @@ def setup_default_config_opts():
         'expand_spec_opts': {
             'rpmspec_opts': [],
         },
+        'system_monitor_enable': False,
```
praiskup, line 267: please also define default configuration for the plugin here

mock/py/mockbuild/plugins/system_monitor.py
praiskup: os.path.join() please

docs/Plugin-SystemMonitor.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: default
+title: Plugin SystemMonitor
+---
+
+This plugin activates per-interval collection of various statistics
+based on the kernels cgroupv2 controllers during the build phase and
+dumps a json file 'systemd-monitor.json' with the collected statitics
```
praiskup: typo: statistics

docs/Plugin-SystemMonitor.md
```diff
@@ -0,0 +1,25 @@
...
+in the result dir
+
+Currently dumped statistics include total maximum memory usage for the build
+and the process with maximum memory RSS
+
+The plugin requires the use for systemd-nspawn as build container runner
```
praiskup: s/use for/ use of/

mock/py/mockbuild/plugins/system_monitor.py
xsuchy: You can find example how we use backoff e.g., in `mock/py/mockbuild/file_downloader.py`

yaneti: I guess sysmon_stop_event can be in the on_predicate lambda .... if thats what the bot means..

yaneti: lest leave that for later . The backoff approach is in current code.

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -0,0 +1,171 @@
+# License: GPL2 or later see COPYING
+"""
+Track various system statistics during the build phase.
+- Report maximum allocated memory in total and for the "biggest" process
+
+The plugin requires systemd-nspawn as build container runner
+"""
+
+import threading
+import time
```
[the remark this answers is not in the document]
xsuchy: This should be fixed (removed)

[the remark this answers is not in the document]
xsuchy: This too.

docs/Plugin-SystemMonitor.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Configuration
+
+The module is disabled by default and needs to be activated by:
+
+    config_opts['plugin_conf']['system_monitor_enable'] = True
+
+The following sub-options may be specified:
+
+    # the interval between statistics collection runs , default 2
```
praiskup: s/2/seconds/

mock/docs/site-defaults.cfg
```diff
@@ -397,6 +397,11 @@
 # certain file paths being accessed.
 #config_opts['plugin_conf']['unbreq_opts']['exclude_accessed_files'] = []
 
+### system_monitor plugin disabled by default
+# config_opts['plugin_conf']['system_monitor_enable'] = False
+## statistics collection interval
+# config_opts['plugin_conf']['system_monitor_opts']['interval'] = 2
```
praiskup, line 403: Please document that the unit is seconds.

praiskup:
Can you please squash the commit into a single one?  otherwise LGTM

xsuchy:
Thank you for your patience and for addressing all our comments.

yaneti:
Thanks for _your_ patience :smile:

PR #1752, opened 2026-05-12
fix: reload uidManager with chrootuid/chrootgid from config

### LINKED ISSUE #1731: config_opts['chrootuid'] with --isolation=simple stopped working
See RHBZ 2454782 for more info.
https://bugzilla.redhat.com/show_bug.cgi?id=2454782

praiskup:
The uidManager was initialized before config was loaded, so it always used the calling user's UID/GID for privilege dropping.  When config specifies a different chrootuid/chrootgid, the uidManager needs to be reloaded with those values — otherwise operations run under the wrong identity, leading to permission errors in the chroot.

Move uidManager initialization after command_parse() (which doesn't need to drop privileges) and reload it with chrootuid/chrootgid from config_opts right after config is loaded (loaded with dropped privileges).

Fixes: #1731

Assisted-By: Claude Code (claude-opus-4-6)

### DIFF
--- a/mock/py/mock.py
+++ b/mock/py/mock.py
@@ -675,20 +675,6 @@ def main():
     # initial sanity check for correct invocation method
     rootcheck()
 
-    # drop unprivileged to parse args, etc.
-    #   uidManager saves current real uid/gid which are unprivileged (callers)
-    #   due to suid helper, our current effective uid is 0
-    #   also supports being run by sudo
-    #
-    #   setuid wrapper has real uid = unpriv,  effective uid = 0
-    #   sudo sets real/effective = 0, and sets env vars
-    #   setuid wrapper clears environment, so there wont be any conflict between these two
-    uidManager = mockbuild.uid.setup_uid_manager()
-
-    # go unpriv only when root to make --help etc work for non-mock users
-    if os.geteuid() == 0:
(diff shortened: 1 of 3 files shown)


mock/py/mock.py
```diff
@@ -683,12 +683,6 @@ def main():
     #   setuid wrapper has real uid = unpriv,  effective uid = 0
     #   sudo sets real/effective = 0, and sets env vars
     #   setuid wrapper clears environment, so there wont be any conflict between these two
```
xsuchy: These comments should be moved too.

mock/py/mock.py
```diff
@@ -704,9 +698,14 @@ def main():
     if options.hermetic_build:
         options.chroot = "hermetic-build"
 
+    uidManager = mockbuild.uid.setup_uid_manager()
     config_opts = uidManager.run_in_subprocess_without_privileges(
             config.load_config, config_path, options.chroot)
 
+    uidManager = mockbuild.uid.reload_uidmanager(uidManager, config_opts)
```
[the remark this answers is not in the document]
praiskup: fixed

mock/py/mock.py
praiskup: fixed

xsuchy (approved):
+1

PR #1754, opened 2026-05-31
plugins: system_monitor: track max swap usage

yaneti:


mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -104,15 +108,26 @@ def sysmon_thread(self, buildroot, interval):
...
                 top = self.get_top_process_info(f"{scope_dir}/payload")
                 if top[0] > self.top_rss[0]:
                     pid_status = "NEW"
                     self.top_rss = top
 
-                if f"{max_status}{pid_status}" != "CurrentCurrent":
+                if f"{max_status}{swap_status}{pid_status}" != "CurrentCurrentCurrent":
                     getLog().debug(
-                        "SYSMON: %s PEAK %.2f MiB | SYSMON: %s Top Process: RSS:%.2f MiB [%s]",
+                        "SYSMON: %s PEAK %.2f MiB | "
```
xsuchy: We now have two peaks, the second is labeled SWAP peak, but this one is now just peak. Can this be "MEMORY PEAK"?

yaneti: sorry , I messed up the strings again. Reworded bot the string and the variable

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -89,30 +91,43 @@ def sysmon_thread(self, buildroot, interval):
             ustr = ustr.decode("utf-8")
         machine_id_unit = ustr.rstrip().split('=')[1]
         scope_dir = f"/sys/fs/cgroup/machine.slice/{machine_id_unit}"
-        peak_file = os.path.join(scope_dir, "memory.peak")
+        memory_peak_file = os.path.join(scope_dir, "memory.peak")
+        swap_peak_file = os.path.join(scope_dir, "memory.swap.peak")
```
praiskup, line 95: Just a comment (or a wish for the future): The most often in Copr, we are curious about the system memory / swap consumption - so it would be nice to have a tool for monitoring that (i.e, peak is fine, but since we are sampling...

praiskup: I mean, here we still monitor the slice, not the system swap consumption.

praiskup (approved):
LGTM

mock/py/mockbuild/plugins/system_monitor.py
yaneti: Indeed, there are more things that could be sampled for more insight...
For me the whole thing started with being annoyed that a single process in the build process could overflow the available memory on a slow riscv64 machine, and not knowing exactly which one.  

Hopefully it would be easier now for people to just add interesting things to the collection and output.

xsuchy (approved):
+1

PR #1755, opened 2026-06-02
unbreq plugin: disable SRPM caching

mkoncek:
This is an actual logic error which causes the plugin to miss BuildRequires fields in case multiple runs of generate BuildRequires are executed.


xsuchy (approved):
+1

PR #1756, opened 2026-06-03
Add pivot_root_chroot option to enable user namespaces

tkopecek:
New `pivot_root_chroot` configuration option uses `pivot_root(2)` inside a private mount namespace to atomically swap the mount-namespace root to the chroot path. This satisfies the kernel check that blocks `unshare(CLONE_NEWUSER)` after plain `chroot()`, allowing tools like pasta/passt that require user namespaces to work inside mock buildroots. This is the same approach used by bubblewrap and other container runtimes. Requires `CAP_SYS_ADMIN` (available in `--privileged` containers); falls back to plain `chroot()` on failure.
Set `config_opts['pivot_root_chroot'] = True` to enable.

Requirements:
- CAP_SYS_ADMIN capability (provided by --privileged containers)
- Already in a mount namespace (Podman/Docker containers)

Assisted-By: Claude Sonnet 4.5

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -77,6 +77,12 @@
 # 'chroot').
 #config_opts['isolation'] = 'auto'
 
+# For "simple" there is also an option to chroot whole process to a new
+# directory. It enables user namespaces inside the build environment, thus e.g.
+# allowing use of pasta/passt (local) networking in build or tests. It is a
+# preview feature disabled by default.
+#config_opts['pivot_root_chroot'] = False
+
 # If you're using isolation='nspawn', then by default networking will be turned
 # off for rpmbuild.  This helps ensure more reproducible builds.
 #config_opts['rpmbuild_networking'] = False
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -93,6 +93,7 @@ def setup_default_config_opts():
     config_opts['online'] = True
     config_opts['isolation'] = None
     config_opts['use_nspawn'] = None
+    config_opts['pivot_root_chroot'] = False
     config_opts['rpmbuild_networking'] = False
     config_opts['nspawn_args'] = ['--capability=cap_ipc_lock']
     # FIXME disable as default because of https://github.com/rpm-software-management/mock/issues/1641
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -40,13 +40,21 @@ from .trace_decorator import getLog, traceLog
 from .uid import setresuid
 from pyroute2 import IPRoute
 
-_libc = ctypes.cdll.LoadLibrary(None)
+_libc = ctypes.CDLL(None, use_errno=True)
 _libc.personality.argtypes = [ctypes.c_ulong]
 _libc.personality.restype = ctypes.c_int
 _libc.unshare.argtypes = [ctypes.c_int]
 _libc.unshare.restype = ctypes.c_int
+_libc.mount.argtypes = [ctypes.c_char_p, ctypes.c_char_p,
+                         ctypes.c_char_p, ctypes.c_ulong,
+                         ctypes.c_void_p]
+_libc.mount.restype = ctypes.c_int
+_libc.umount2.argtypes = [ctypes.c_char_p, ctypes.c_int]
+_libc.umount2.restype = ctypes.c_int
 _libc.sethostname.argtypes = [ctypes.c_char_p, ctypes.c_int]
 _libc.sethostname.restype = ctypes.c_int
+_libc.pivot_root.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
+_libc.pivot_root.restype = ctypes.c_int
 
 # See linux/include/sched.h
 CLONE_NEWNS = 0x00020000
@@ -54,6 +62,14 @@ CLONE_NEWUTS = 0x04000000
 CLONE_NEWPID = 0x20000000
 CLONE_NEWNET = 0x40000000
 CLONE_NEWIPC = 0x08000000
+
+# Mount flags used by condChrootPivotRoot
+# See libmount/libmount.h (libmount-devel)
+MS_BIND    = 0x1000       # 4096: Mount existing tree elsewhere as well
+MS_REC     = 0x4000       # 16384: Recursive loopback
+MS_PRIVATE = 0x40000      # 262144: Make private
+# See sys/mount.h
+MNT_DETACH = 0x00000002   # 2: Just detach from tree
 
 # taken from sys/personality.h
 PER_LINUX32 = 0x0008
@@ -310,6 +326,65 @@ def condChroot(chrootPath):
         setresuid(saved['ruid'], saved['euid'])
 
 
+def _do_pivot_root(chrootPath):
+    """Perform pivot_root(2) chroot.  Caller must already be running as root."""
+    chroot_bytes = chrootPath.encode('utf-8')
+
+    # New mount namespace so our mount changes don't affect the parent
+    if _libc.unshare(CLONE_NEWNS) != 0:
+        raise OSError(ctypes.get_errno(), "unshare(CLONE_NEWNS)")
+
+    # Make all existing mounts private to prevent propagation
+    if _libc.mount(b"none", b"/", None, MS_REC | MS_PRIVATE, None) != 0:
+        raise OSError(ctypes.get_errno(), "mount --make-rprivate /")
+
+    # Bind-mount the chroot dir onto itself so it becomes a mount point
+    # (pivot_root requires new_root to be a mount point)
+    if _libc.mount(chroot_bytes, chroot_bytes, None,
+                    MS_BIND | MS_REC, None) != 0:
+        raise OSError(ctypes.get_errno(), "bind mount " + chrootPath)
+
+    # Directory where the old root will be stashed
+    put_old = os.path.join(chrootPath, ".pivot_old")
+    file_util.mkdirIfAbsent(put_old)
+    put_old_bytes = put_old.encode('utf-8')
+
+    if _libc.pivot_root(chroot_bytes, put_old_bytes) != 0:
+        raise OSError(ctypes.get_errno(), "pivot_root")
+
+    os.chdir("/")
+
+    # Detach the old root — it is no longer needed
+    if _libc.umount2(b"/.pivot_old", MNT_DETACH) != 0:
+        raise OSError(ctypes.get_errno(), "umount2 /.pivot_old")
+
+    os.rmdir("/.pivot_old")
+
+
+def condChrootPivotRoot(chrootPath):
+    """
+    Chroot using pivot_root(2) to allow user namespaces afterward.
+
+    The kernel blocks unshare(CLONE_NEWUSER) after plain chroot() because
+    the process root no longer matches the mount namespace root.  By using
+    pivot_root(2) inside a private mount namespace we atomically swap the
+    mount-namespace root to chrootPath, satisfying the kernel check.
+
+    This is the same approach used by bubblewrap and other container runtimes.
+
+    Requires CAP_SYS_ADMIN (available in --privileged containers).
+    """
+    if chrootPath is None:
+        return
+
+    saved = {"ruid": os.getuid(), "euid": os.geteuid()}
+    setresuid(0, 0, 0)
+    try:
+        _do_pivot_root(chrootPath)
+    finally:
+        setresuid(saved['ruid'], saved['euid'])
+
+
 def condChdir(cwd):
     if cwd is not None:
         os.chdir(cwd)
@@ -526,7 +601,7 @@ def do(*args, **kargs):
 def do_with_status(command, shell=False, chrootPath=None, cwd=None, timeout=0, raiseExc=True,
                    returnOutput=0, uid=None, gid=None, user=None, personality=None,
                    printOutput=False, env=None, pty=False, nspawn_args=None, unshare_net=False,
-                   returnStderr=True, *_, **kargs):
+                   returnStderr=True, pivot_root_chroot=False, *_, **kargs):
     logger = kargs.get("logger", getLog())
     if timeout == 0:
         timeout = _OPS_TIMEOUT
@@ -536,8 +611,12 @@ def do_with_status(command, shell=False, chrootPath=None, cwd=None, timeout=0, r
         lead_pty, sub_pty = os.openpty()
         resize_pty(sub_pty)
         reader = os.fdopen(lead_pty, 'rb')
+    if pivot_root_chroot and chrootPath and USE_NSPAWN:
+        logger.warning(
+            "pivot_root_chroot=True ignored when using systemd-nspawn")
     preexec = ChildPreExec(personality, chrootPath, cwd, uid, gid,
-                           unshare_ipc=bool(chrootPath), unshare_net=unshare_net)
+                           unshare_ipc=bool(chrootPath), unshare_net=unshare_net,
+                           pivot_root_chroot=pivot_root_chroot)
     if env is None:
         env = clean_env()
     stdout = None
@@ -616,11 +695,13 @@ def do_with_status(command, shell=False, chrootPath=None, cwd=None, timeout=0, r
 class ChildPreExec(object):
     def __init__(self, personality, chrootPath, cwd, uid, gid, env=None,
                  shell=False, unshare_ipc=False, unshare_net=False,
-                 no_setsid=False):
+                 no_setsid=False, pivot_root_chroot=False):
         """
         Params:
         - no_setsid - assure we don't call os.setsid(), as the process we run
             calls that itself
+        - pivot_root_chroot - use pivot_root(2) for chroot to enable
+            user namespaces afterward (requires CAP_SYS_ADMIN)
         """
         self.personality = personality
         self.chrootPath = chrootPath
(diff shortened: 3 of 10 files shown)


mock/py/mockbuild/util.py
```diff
@@ -310,6 +334,89 @@
...
+        try:
+            os.rmdir("/.pivot_old")
+        except OSError:
+            pass
+
+    except OSError as e:
+        if bind_mount_created:
+            try:
+                _libc.umount2(chrootBytes, MNT_DETACH)
+            except Exception:
```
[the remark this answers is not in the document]
tkopecek: It is not a real catch - it is trying to recover what is possible after already failed pivot_root

mock/tests/test_mount_pivot_chroot.py
```diff
@@ -0,0 +1,244 @@
...
+import ctypes
+import pytest
+import tempfile
+import shutil
+
+# Add the py directory to the path to import the local version
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'py'))
+
+# Import constants from mockbuild.util
+from mockbuild.util import (
```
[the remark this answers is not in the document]
tkopecek: Depends on sys.path.insert which should be after base modules are imported to not inject something else.

mock/tests/test_mount_pivot_chroot.py
```diff
@@ -0,0 +1,244 @@
...
+
+import os
+import sys
+import ctypes
+import pytest
+import tempfile
+import shutil
+
+# Add the py directory to the path to import the local version
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'py'))
```
xsuchy: Why is this here?

tkopecek: Ah, you're right - I had it there as it mixed with installed mock. /I'll drop that/. Bigger issue is that tests must run with CAP_SYS_ADMIN (so, under the root or in privileged container). Do you've some other tests dealing with such requirements? Otherwise it is just skipped :-(

mock/py/mockbuild/util.py
```diff
@@ -310,6 +334,89 @@ def condChroot(chrootPath):
...
+                pass
+        log.warning("pivot_root chroot failed (%s), falling back to plain chroot", e)
+        setresuid(saved['ruid'], saved['euid'])
+        condChroot(chrootPath)
+        return
+
+    setresuid(saved['ruid'], saved['euid'])
```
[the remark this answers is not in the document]
tkopecek: should be fixed now

mock/py/mockbuild/util.py
```diff
@@ -644,9 +755,14 @@ def __call__(self, *args, **kargs):
...
         if not USE_NSPAWN or not self.chrootPath:
-            condChroot(self.chrootPath)
+            if self.pivot_root_chroot:
+                condChrootPivotRoot(self.chrootPath)
+            else:
+                condChroot(self.chrootPath)
             condDropPrivs(self.uid, self.gid)
             condChdir(self.cwd)
+        elif self.pivot_root_chroot and self.chrootPath:
+            getLog().warning("pivot_root_chroot=True ignored when using systemd-nspawn")
```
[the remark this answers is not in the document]
tkopecek: should be fixed now

mock/tests/test_mount_pivot_chroot.py
praiskup: we have behave / "old" testsuite, those are run in (real host's) root mode ... but these are not yet? transitively running unit-tests.  That can be changed though ... (you can, or I can take a look, if you make the test "skipped" inside of a non-privileged environment)

mock/py/mockbuild/util.py
```diff
@@ -310,6 +334,95 @@ def condChroot(chrootPath):
...
+    This is the same approach used by bubblewrap and other container runtimes.
+
+    Requires CAP_SYS_ADMIN (available in --privileged containers).
+    Falls back to plain chroot() on failure.
+    """
+    if chrootPath is None:
+        return
+
+    saved = {"ruid": os.getuid(), "euid": os.geteuid()}
+    setresuid(0, 0, 0)
```
praiskup, line 381: Can we use uid_manger.elevate_privileges()?

mock/py/mockbuild/util.py
```diff
@@ -55,6 +62,23 @@
...
+MNT_DETACH = 0x00000002   # 2: Just detach from tree
+
+# Syscall number for pivot_root(2) — architecture dependent
+_SYS_PIVOT_ROOT = {
+    'x86_64': 155, 'i386': 217, 'i686': 217,
+    'aarch64': 41, 'armv7l': 218,
+    'ppc64le': 203, 'ppc64': 203,
+    's390x': 217, 's390': 217,
+    'riscv64': 41, 'loongarch64': 41,
+}.get(os.uname().machine)
```
praiskup: ouch ... these numbers ... can't we do this instead?
```python
try:
    pivot_root_syscall = libc.pivot_root
except AttributeError:
    .... ??? something ???...
```

mock/py/mockbuild/util.py
tkopecek: updated

mock/py/mockbuild/util.py
tkopecek: I've been looking for them so heavily that I've ignored the simple solution :-D

tkopecek:
ping @praiskup

mock/py/mockbuild/config.py
```diff
@@ -93,6 +93,7 @@ def setup_default_config_opts():
     config_opts['online'] = True
     config_opts['isolation'] = None
     config_opts['use_nspawn'] = None
+    config_opts['pivot_root_chroot'] = False
```
praiskup, line 96: please add this into site-defaults.cfg doc file

mock/docs/site-defaults.cfg
```diff
@@ -77,6 +77,12 @@
 # 'chroot').
 #config_opts['isolation'] = 'auto'
 
+# For "simple" there is also an option to chroot whole process to new directory.
+# It enables user namespaces inside the build environment, thus e.g. allowing
+# use pasta/passt (local) networking in build or tests. It can break some other
+# usecases, so it is not turn on by default.
```
praiskup: What other use-cases?  Some example / link would be nice.
I suppose we are going to turn this on in Konflux?

mock/py/mockbuild/util.py
```diff
@@ -301,13 +318,109 @@ def sethostname(hostname):
...
+        # Make all existing mounts private to prevent propagation
+        if _libc.mount(b"none", b"/", None, MS_REC | MS_PRIVATE, None) != 0:
+            raise OSError(ctypes.get_errno(), "mount --make-rprivate /")
+
+        # Bind-mount the chroot dir onto itself so it becomes a mount point
+        # (pivot_root requires new_root to be a mount point)
+        if _libc.mount(chroot_bytes, chroot_bytes, None,
+                       MS_BIND | MS_REC, None) != 0:
+            raise OSError(ctypes.get_errno(), "bind mount " + chrootPath)
+        pivoted = True
```
praiskup: this one should be set True after the libc.pivot_root() call

mock/py/mockbuild/util.py
```diff
@@ -301,13 +318,109 @@ def sethostname(hostname):
...
+    """Perform pivot_root(2) chroot.  Caller must already be running as root."""
+    chroot_bytes = chrootPath.encode('utf-8')
+    pivoted = False
+    bind_mount_created = False
+
+    try:
+        if not os.path.isdir(chrootPath):
+            raise OSError(errno.ENOENT, f"chroot path does not exist: {chrootPath}")
+
+        if not getattr(_libc, 'pivot_root'):
```
praiskup: is this dead code..?  (line 57 would raise traceback)

mock/py/mockbuild/util.py
```diff
@@ -301,13 +318,109 @@ def sethostname(hostname):
...
+
+def _do_pivot_root(chrootPath):
+    """Perform pivot_root(2) chroot.  Caller must already be running as root."""
+    chroot_bytes = chrootPath.encode('utf-8')
+    pivoted = False
+    bind_mount_created = False
+
+    try:
+        if not os.path.isdir(chrootPath):
+            raise OSError(errno.ENOENT, f"chroot path does not exist: {chrootPath}")
```
praiskup: I'm not fan of "isdir" checks (TOCTOU risks), what if we let it die later?

mock/py/mockbuild/util.py
```diff
@@ -301,13 +318,109 @@ def sethostname(hostname):
...
+        # (pivot_root requires new_root to be a mount point)
+        if _libc.mount(chroot_bytes, chroot_bytes, None,
+                       MS_BIND | MS_REC, None) != 0:
+            raise OSError(ctypes.get_errno(), "bind mount " + chrootPath)
+        pivoted = True
+        bind_mount_created = True
+
+        # Directory where the old root will be stashed
+        put_old = os.path.join(chrootPath, ".pivot_old")
+        os.makedirs(put_old, exist_ok=True)
```
praiskup: mkdirIfAbsent

mock/py/mockbuild/util.py
```diff
@@ -40,13 +40,22 @@
...
 _libc.personality.restype = ctypes.c_int
 _libc.unshare.argtypes = [ctypes.c_int]
 _libc.unshare.restype = ctypes.c_int
+_libc.mount.argtypes = [ctypes.c_char_p, ctypes.c_char_p,
+                         ctypes.c_char_p, ctypes.c_ulong,
+                         ctypes.c_void_p]
+_libc.mount.restype = ctypes.c_int
+_libc.umount2.argtypes = [ctypes.c_char_p, ctypes.c_int]
+_libc.umount2.restype = ctypes.c_int
+_libc.syscall.restype = ctypes.c_long
```
praiskup: is syscall needed?

praiskup (commented):
Otherwise seems good, thank you for the testsuite

mock/docs/site-defaults.cfg
praiskup: Or is this going to be enabled on a per-package basis?

tkopecek: - virt-v2v is spawning qemu machines during test phase which needs to communicate over network. They can do so in hermetic mode if userspace pasta can be instantiated.
- I would like to turn it on for whole konflux, but maybe some other packages will fail on this, so we will need to tune it per package.

mock/py/mockbuild/util.py
tkopecek: leftover, dropping

mock/py/mockbuild/util.py
```diff
@@ -799,7 +917,9 @@ def doshell(chrootPath=None, environ=None, uid=None, gid=None, cmd=None,
             cwd=None,
             nspawn_args=None,
             unshare_ipc=True,
-            unshare_net=False):
+            unshare_net=False,
+            pivot_root_chroot=False,
+            uid_manager=None):
```
tkopecek: @praiskup Any idea, what to about complaining pylint here (Too many arguments (11/10))?

praiskup: adjust pylintrc to 15? :-) ... or just `# pylint: disable=...`

mock/docs/site-defaults.cfg
praiskup: If we don't know about any real failures, I'd avoid the "speculative tone" sentence to not make people scared?  Or just change it to something like "a preview feature, disabled by default"?

mock/py/mockbuild/util.py
```diff
@@ -301,13 +317,102 @@ def sethostname(hostname):
...
+
+    except OSError as e:
+        if bind_mount_created:
+            try:
+                _libc.umount2(chroot_bytes, MNT_DETACH)
+            except Exception:  # pylint: disable=broad-exception-caught
+                pass
+        if pivoted:
+            raise RuntimeError(f"pivot_root succeeded but cleanup failed: {e}") from e
+        # Fall back to plain chroot (already running as root)
```
praiskup: I'm confused.  Do we really want to call chroot() after pivot_root()?

tkopecek: If it was already pivoted, we're doomed. Otherwise, it could be a way (as most builds don't care). But to have cleaner approach, we can fail in such case and don't fallback (and force user to turn that feature off).

tkopecek: Probably depends on how similar situation are handled elsewhere in the code.

mock/py/mockbuild/util.py
```diff
@@ -301,13 +317,102 @@ def sethostname(hostname):
...
+            raise OSError(ctypes.get_errno(), "unshare(CLONE_NEWNS)")
+
+        # Make all existing mounts private to prevent propagation
+        if _libc.mount(b"none", b"/", None, MS_REC | MS_PRIVATE, None) != 0:
+            raise OSError(ctypes.get_errno(), "mount --make-rprivate /")
+
+        # Bind-mount the chroot dir onto itself so it becomes a mount point
+        # (pivot_root requires new_root to be a mount point)
+        if _libc.mount(chroot_bytes, chroot_bytes, None,
+                       MS_BIND | MS_REC, None) != 0:
```
praiskup: is MS_REC correct here?

mock/py/mockbuild/util.py
```diff
@@ -301,13 +317,102 @@ def sethostname(hostname):
...
-def condChroot(chrootPath):
+def condChroot(chrootPath, uid_manager=None):
     if chrootPath is not None:
-        saved = {"ruid": os.getuid(), "euid": os.geteuid()}
-        setresuid(0, 0, 0)
+        if uid_manager:
+            with uid_manager.elevated_privileges():
+                os.chdir(chrootPath)
+                os.chroot(chrootPath)
+        else:
```
praiskup: This seems to be a dead code now.

mock/py/mockbuild/util.py
```diff
@@ -301,13 +317,102 @@ def sethostname(hostname):
...
+
+        os.chdir("/")
+
+        # Detach the old root — it is no longer needed
+        if _libc.umount2(b"/.pivot_old", MNT_DETACH) != 0:
+            raise OSError(ctypes.get_errno(), "umount2 /.pivot_old")
+        try:
+            os.rmdir("/.pivot_old")
+        except OSError:
+            pass
```
praiskup: I don't think we need this exception handler;  if it doesn't exist, the situation is so weird so we should let it bomb on us.

mock/py/mockbuild/util.py
```diff
@@ -301,13 +317,102 @@ def sethostname(hostname):
...
+
+    This is the same approach used by bubblewrap and other container runtimes.
+
+    Requires CAP_SYS_ADMIN (available in --privileged containers).
+    Falls back to plain chroot() on failure.
+    """
+    if chrootPath is None:
+        return
+
+    if uid_manager:
```
praiskup: same here, uid_manager is always set

mock/py/mockbuild/util.py
tkopecek: It is overriding potential MS_SHARED which is not allowed by pivot_root (which would fail)

mock/py/mockbuild/util.py
tkopecek: dropped

mock/py/mockbuild/util.py
tkopecek: dropped (plus similar code in condChrootPivotRoot). Question if not to drop these functions completely and make it inline?

tkopecek: But maybe it is more readable this way.

mock/py/mockbuild/util.py
tkopecek: Reverted - it is not always. Many calls are coming from non-broot calls (plugins, SCM, etc.) with unset uid_manager.

mock/py/mockbuild/util.py
praiskup: This is a can of worms, and I was wrong that uid_manager is the right choice :-( sorry - if we can't be sure uid_manager is available, can we just keep the old (non-uid_manager) pattern?  Ditto for condChrootPivot().

mock/py/mockbuild/util.py
praiskup: Nah, sorry for the confusion.  Can we just keep just one way of raising privileges here, the manual "setresuid(0, 0, 0)" pattern?

mock/py/mockbuild/util.py
```diff
@@ -616,11 +697,14 @@ def do_with_status(command, shell=False, chrootPath=None, cwd=None, timeout=0, r
...
                  shell=False, unshare_ipc=False, unshare_net=False,
-                 no_setsid=False):
+                 no_setsid=False, pivot_root_chroot=False, uid_manager=None):
         """
         Params:
         - no_setsid - assure we don't call os.setsid(), as the process we run
             calls that itself
+        - pivot_root_chroot - use pivot_root(2) for chroot to enable
+            user namespaces afterward (requires CAP_SYS_ADMIN)
+        - uid_manager - UidManager instance for privilege handling
```
praiskup: nit: this is now unused

mock/docs/site-defaults.cfg
```diff
@@ -77,6 +77,12 @@
 # 'chroot').
 #config_opts['isolation'] = 'auto'
 
+# For "simple" there is also an option to chroot whole process to a new
+# directory. It enables user namespaces inside the build environment, thus e.g.
+# allowing use pasta/passt (local) networking in build or tests. It is a
+# preview feature disabled by default.
+# config_opts['pivot_root_chroot'] = False
```
praiskup: nit: we tend to drop the whitespace here, to optically highlight commented-out configuration

praiskup:
LGTM, can you please squash?  Or I can, just please tell.

tkopecek:
squashed and rebased

praiskup (approved):
LGTM, thank you for the great work.

PR #1757, opened 2026-06-03
unbreq plugin: do not treat not installed packages as unused

mkoncek:


xsuchy (approved):
+1

mock/py/mockbuild/plugins/unbreq.py
```diff
@@ -284,6 +284,8 @@ def resolve_buildrequires(self) -> None:
                     "unbreq plugin: BuildRequires '%s' is needed because file %s was accessed",
                     br, path
                 )
+            elif not providers:
+                getLog().info("unbreq plugin: BuildRequires '%s' is not installed", br)
```
praiskup: Could we mention "rich deps" here in the log wording?  Initially I was very curious how this would be possible...

mkoncek: Right, I forgot how they are named.

praiskup (approved):

PR #1762, opened 2026-06-11
Fixes #1761: Add Fedora Rawhide RISC-V

### LINKED ISSUE #1761: Add Fedora Rawhide RISC-V
Add Fedora Rawhide RISC-V to mock-core-configs

jasonmontleon:


mock-core-configs/etc/mock/fedora-rawhide-riscv64.cfg
```diff
@@ -0,0 +1,4 @@
+config_opts['target_arch'] = 'riscv64'
+config_opts['legal_host_arches'] = ('riscv64',)
```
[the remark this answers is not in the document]
jasonmontleon: This runs counter to the other fedora-rawhide configs for other architectures.

jasonmontleon:
I adjusted the existing template name and configs so that it would be more in sync with the other fedora-branched/fedora-rawhide template names. If it's not desired I can switch it back, just let me know.

mock-core-configs/etc/mock/templates/fedora-branched-riscv64.tpl
```diff
@@ -34,7 +34,7 @@ user_agent={{ user_agent }}
 name=local
 baseurl=https://riscv-koji.fedoraproject.org/repos/f{{ releasever }}-build/latest/riscv64/
 gpgcheck=0
-enabled=1
+enabled=0
```
praiskup, line 37: If `local` is disabled, where are the RPMs coming from?

praiskup:
Thank you for the progress, Jason.

mock-core-configs/etc/mock/templates/fedora-branched-riscv64.tpl
davidlt: There are two repos: dist-repos (distribution repositories) with Fedora COMPS and Fedora Koji working repos (minimal COMPS). So `local` repos point to `fXY-build` which is Koji internal tag and these are "working repos". Simply put using those repos is like building in Koji directly, it doesn't contain Fedora full COMPS.

So the question if:
- Should stable/branched released use dist-repos (stable repos) with full Fedora COMPS?
- Should Rawhide (not-branched) use Koji direct repos (since typically those don't have distro repos available until later in time).

praiskup: > Should stable/branched released use dist-repos (stable repos) with full Fedora COMPS?

If they are available already, please, yes!  Otherwise, we need to continue building from Koji (it doesn't make sense to have a non-working Mock config IMO).

> Should Rawhide (not-branched) use Koji direct repos (since typically those don't have distro repos available until later in time).

I'd say the same rules apply to both Fedora Rawhide and branched.

davidlt: We typically generate dist-repos quite late Fedora XY cycle. Usually that's after it gets branched. Initial months  are only Koji working repos.

In this particular case I generated dist-repos for Rawhide, but it's almost identical to F44 (branched). It should be stable. I probably not gonna upgrade it until Rawhide/F45 gets into a good shape. Thus anyone targeting their work Rawhide/F45 should use Koji working repos (local).

praiskup: Ok, thank you for the explanation.

Seems there's `fedora` repo which has `enabled=1`.  This seems fine to me.

mock-core-configs/etc/mock/templates/fedora-branched-riscv64.tpl
```diff
@@ -1,7 +1,7 @@
 config_opts['root'] = 'fedora-{{ releasever }}-{{ target_arch }}'
 
 config_opts['description'] = 'Fedora {{ releasever }} RISC-V'
-config_opts['chroot_setup_cmd'] = 'install @build'
+config_opts['chroot_setup_cmd'] = 'install buildsys-build'
```
praiskup: I'm getting a build failure:
```
ERROR: Exception(/tmp/quick-package/dummy-pkg-20260619_1740-1.src.rpm) Config(fedora-44-riscv64) 0 minutes 14 seconds
INFO: Results and/or logs in: /var/lib/mock/fedora-44-riscv64/result
ERROR: Command failed: 
 # /usr/bin/dnf5 --installroot /var/lib/mock/fedora-44-riscv64/root/ --releasever 44 install buildsys-build --setopt=deltarpm=False --setopt=allow_vendor_change=yes --allowerasing --forcearch riscv64 --setopt=tsflags=nocontexts
Updating and loading repositories:
 fedora                                 100% |   1.8 MiB/s |  14.3 MiB |  00m08s
Repositories loaded.
Failed to resolve the transaction:
No match for argument: buildsys-build
You can try to add to command line:
  --skip-unavailable to skip unavailable packages
```

davidlt: Looking at https://riscv-koji.fedoraproject.org/repos-dist/f44/latest/groups/comps.xml it should be OK:
```
<group>
<id>buildsys-build</id>
<name>Buildsystem building group</name>
```
Ah, isn't it missing `@` for the group? Thus it attempted to install non-existing package instead.

jasonmontleon: It was the `@` missing. I fixed the template, tested successfully, and have pushed it. I tested rawhide successfully again as well.

praiskup:
Thank you.

Issue #684, opened 2020-12-25
Optional support for polkit (additionally to usermode)?


robert-scheck:
I wonder if there is a chance for optional polkit support (at least additionally to existing usermode support) at mock?

xsuchy:
I tried that - not so long time ago. I hit the lack of documentation and IIRC lack of some features. After discussion with package maintainer of polkit, I decided to stay with usermode.

If anyone want to try to investigate current situation, file PR and discuss potentially missing features with polkit developers - feel free to do it.

robert-scheck:
So what are the missing features at polkit?

xsuchy:
IIRC we need to handle ENV. Propagate some, filter out some, set some.

PR #1763, opened 2026-06-12
Include support for polkit

rajibade:
This is only applicable if the PR to support vars in polkit is accepted.

@xsuchy Per your previous attempt, what didn't get addressed to allow for this support? does the changes in the polkit PR help?

Polkit feature request: https://github.com/polkit-org/polkit/pull/672
Addresses: https://github.com/rpm-software-management/mock/issues/684


xsuchy:
AFAIK consolehelper still does not support passing the env vars and other things we have in consolehelper configs. 
But I am not against having it as option during the build of RPM - with disabled by default.

I will do a full review later next week.

rajibade:
Yeah that was what I was hoping to fix with this. I put up the PR to the polkit maintainers, that should help with passing env vars. check above. is there any other thing aside from the vars? and yes, this PR adds it to the spec file disabled by default.

rajibade:
@xsuchy any chance to take a look at this yet?

xsuchy:
I have no objections. 
Can I only ask you to edit the release notes files to be more verbose: state this is experimental and likely miss features that consolehelper has. It is disabled by default and how it can be enabled (if somebody wants to try it).

rajibade:
Okay thanks. Yeah I can. I'm hopeful that the PR to the polkit maintainers should put it on par with consolehelper, but happy to do more testing with it later when that's out

mock/mock.spec
```diff
@@ -31,13 +32,19 @@ URL: https://github.com/rpm-software-management/mock/
 BuildArch: noarch
 Requires: tar
 Requires: pigz
+%if %{without polkit}
```
praiskup, line 35: Are you sure you want to have this disabled by default?

mock/mock.spec
```diff
@@ -1,5 +1,6 @@
 %bcond_with lint
 %bcond_without tests
+%bcond_with polkit
```
praiskup, line 3: Are you sure you want to have this disabled by default?

xsuchy: I think this is wanted. At least for me this is a MUST in this state of polkit.

praiskup:
OK, looks good to me anyway. Thank you for the patch!

mock/mock.spec
rajibade: Yeah I agree on keeping this disabled. the polkit support isn't merged yet

rajibade:
Aw didn't get to update the release-notes on this. sorry.

PR #1764, opened 2026-06-20
configs: opensuse: Forcibly disable repo_gpgcheck for now

Conan-Kudo:
openSUSE Tumbleweed's dnf5 package defaults to pkg_gpgcheck and repo_gpgcheck being enabled since openSUSE and OBS repositories have PGP signatures for repository metadata by convention.

However, COPR does not yet produce metadata signatures, so until COPR does, this needs to be turned off.

### DIFF
--- a/releng/release-notes-next/disable-repo_gpgcheck-for-opensuse.config
+++ b/releng/release-notes-next/disable-repo_gpgcheck-for-opensuse.config
@@ -1,0 +1,2 @@
+Forcibly disable repo_gpgcheck for openSUSE chroots for now,
+since COPR does not yet sign its own produced repositories.
--- a/mock-core-configs/etc/mock/templates/opensuse-leap-15.6.tpl
+++ b/mock-core-configs/etc/mock/templates/opensuse-leap-15.6.tpl
@@ -24,6 +24,7 @@ logfile=/var/log/yum.log
 retries=20
 obsoletes=1
 gpgcheck=0
+repo_gpgcheck=0
 assumeyes=1
 syslog_ident=mock
 syslog_device=
--- a/mock-core-configs/etc/mock/templates/opensuse-leap-16.0.tpl
+++ b/mock-core-configs/etc/mock/templates/opensuse-leap-16.0.tpl
@@ -26,6 +26,7 @@ logfile=/var/log/yum.log
 retries=20
 obsoletes=1
 gpgcheck=0
+repo_gpgcheck=0
 assumeyes=1
 syslog_ident=mock
 syslog_device=
--- a/mock-core-configs/etc/mock/templates/opensuse-tumbleweed.tpl
+++ b/mock-core-configs/etc/mock/templates/opensuse-tumbleweed.tpl
@@ -24,6 +24,7 @@ logfile=/var/log/yum.log
 retries=20
 obsoletes=1
 gpgcheck=0
+repo_gpgcheck=0
 assumeyes=1
 syslog_ident=mock
 syslog_device=


Conan-Kudo:
This will address https://github.com/fedora-copr/copr/issues/4352 until https://github.com/fedora-copr/copr/issues/2644 is implemented.

releng/release-notes-next/disable-repo_gpgcheck-for-opensuse.config
```diff
@@ -0,0 +1,2 @@
+Forcibly disable repo_gpgcheck for openSUSE chroots for now,
+since COPR does not yet sign its own produced repositories.
```
praiskup, line 2: Seems like repo_gpgcheck=1 should be set for OpenSuse repos, though?

praiskup: Or is there a reason to keep repo_gpgcheck=false "globally"?

praiskup (changes requested):

releng/release-notes-next/disable-repo_gpgcheck-for-opensuse.config
Conan-Kudo: Leaving it enabled would only work for repos from the SUSE ecosystem like OBS. And enabling it with dnf4 causes issues, so globally disabling it by default for now is just safer.

praiskup: > Leaving it enabled would only work for repos from the SUSE ecosystem like OBS.

Which is the case, or?

> And enabling it with dnf4 causes issues

Which is not the case, because bootstrap is ON?

I mean, the change LGTM (if you think it's the right thing to do), but the comment here doesn't align with the change.  Note that Mock != Copr, and many Mock users wouldn't care about the COPR reasoning here (my bet is that you care about the COPR use-case, this is rather a Copr issue).

Conan-Kudo: > > Leaving it enabled would only work for repos from the SUSE ecosystem like OBS.
> 
> Which is the case, or?
> 

Multi-distro RPM repositories rarely have it, and it is less likely that folks using Mock for building packages for the SUSE ecosystem would have it either, since they are likely from the Red Hat world. That said, we could turn it on for the core openSUSE repos at least in Tumbleweed.

> > And enabling it with dnf4 causes issues
> 
> Which is not the case, because bootstrap is ON?
> 

Not for Leap, because Leap uses dnf4 in its container currently.

> I mean, the change LGTM (if you think it's the right thing to do), but the comment here doesn't align with the change. Note that Mock != Copr, and many Mock users wouldn't care about the COPR reasoning here (my bet is that you care about the COPR use-case, this is rather a Copr issue).

It basically affects any build system that uses Mock's bootstrap chroot mode (so everything _but_ Koji). It's more than COPR. Offhand, I believe ABF is another system that would be equally affected.

praiskup: Btw., Fedora Koji uses bootstrap (but it doesn't build openSUSE..).

OK, indeed this seems like a desired change; just please better document why we think a normal user don't have those (rather then "this is because of Copr").  If this was because of Copr, it would be better & easier to change  `[copr_base]` [there](https://github.com/fedora-copr/copr/issues/new/choose).

Conan-Kudo: There's already a COPR ticket, if we can change COPR itself, then we can avoid this until someone else reports a problem.

praiskup:
Ok, proposing a Copr fix: https://github.com/fedora-copr/copr/pull/4359

This should mean that all Copr repositories consumed from the same copr will have repo_gpgcheck disabled.  Other users shouldn't be affected at all (unless they misconfigure same as Copr did).

PR #1766, opened 2026-06-26
Add support for isolated %check phase

### LINKED ISSUE #1352: RFE: make easier to build systems to download artifacts after %install and before %check phase
The apoXZalypse just started, but we can learn a thing from that: trust of whole upstream tarball is not granted. We can minimize such attacks in future when we run tests in isolation without ability to affect final artifacts.

Now we run all phases of rpmbuild. Including %check and then offers the artifacts for next use. And I am looking at Koji, Copr and other build system, not just cli-user.

What we can change is that we can run %build, %install - pause build system can fetch artifacts - and run %check and report status.

praiskup:
Add support for running %check as a separate rpmbuild phase using rpmbuild -bk --short-circuit (rpm >= 6.0.91).  This allows Mock to protect built artifacts (RPMS/, SRPMS/) under a tmpfs overlay during %check, preventing rogue test scripts from modifying or deleting packages.

New config option `separate_check` with three modes:

- 'off' (default): monolithic build, %check runs inline
- 'best_effort': use separate %check if rpmbuild supports -bk, otherwise fall back to inline
- 'on': require separate %check, fail if unsupported

Fixes: #1352
Relates: https://github.com/rpm-software-management/rpm/issues/3015

## Summary by CodeRabbit

* **New Features**
  * Added `--separate-check MODE` / `separate_check` to run `%check` as a separate `rpmbuild` phase (default: `off`), with `best_effort` and `enforce` modes.
  * `%check` runs in isolation to reduce the chance of it altering already-produced RPM artifacts.
* **Documentation**
  * Updated the `mock(1)` man page and default configuration docs with option details, valid modes, and behavior.
* **Bug Fixes**
  * Added a dedicated result code for invalid/unsupported option values (exit code `7`).


mock/py/mockbuild/backend.py
```diff
@@ -712,13 +722,62 @@ def rebuild_installed_srpm(self, spec_path, timeout):
...
+                "separate_check='enforce' requires rpmbuild with -bk support"
+                " (rpm >= 6.0.91)")
+
+        if separate_check == 'best_effort' and not supports_bk:
+            self.buildroot.build_log.warning(
+                "rpmbuild does not support -bk, falling back to"
+                " monolithic build with %check inline")
+            return False
+
+        return True
```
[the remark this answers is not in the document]
praiskup: wrong value now raises exception

mock/py/mockbuild/backend.py
```diff
@@ -712,13 +722,62 @@ def rebuild_installed_srpm(self, spec_path, timeout):
...
+                    path=artifact_dir, filetype='tmpfs',
+                    options='size=1m'))
+        getLog().info("Protecting built artifacts in RPMS/ and SRPMS/ with tmpfs overlay")
+        for mp in mounts:
+            mp.mount()
+        try:
+            yield
+        finally:
+            for mp in mounts:
+                mp.umount()
```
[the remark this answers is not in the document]
praiskup: fixed by directly mounting, and later unmounting only those that were actually mounted

praiskup:
> mock/py/mockbuild/backend.py:792:4: R0915[too-many-statements]: Commands.rebuild_package: Too many statements (58/50)

I'm going to ignore this one, unless a human reviewer wants me to do something about it.

mock/docs/mock.1
```diff
@@ -383,6 +383,24 @@ Do not clean chroot before building a package.
...
+\fB\-\-separate\-check\fR \fIMODE\fR
+Run %check as a separate rpmbuild phase (rpmbuild \-bk \-\-short\-circuit)
+after the main build finishes.  This is a security measure \(em built RPM
+artifacts cannot be modified by %check scripts.  During the %check phase,
+the RPMS/ and SRPMS/ directories are hidden under a tmpfs overlay.
+Requires rpm >= 6.0.91.  Valid values for \fIMODE\fR:
+.RS
+.TP
+.B best_effort
+Use separate %check when rpmbuild supports \-bk, fall back to inline otherwise.
```
xsuchy: What is "inline". I do not parse that sentence.

mock/py/mockbuild/backend.py
```diff
@@ -665,29 +666,37 @@ def install_srpm(self, srpm_path):
...
     @property
     def _rpmbuild_noclean_option(self):
         """
-        Detect and cache if rpmbuild in buildroot supports the --noclean
-        option.  Return "--noclean" string if supported, otherwise return an
-        empty string.
+        Return "--noclean" if rpmbuild supports it, otherwise "".
 
-        TODO: Remove this method once nobody is building for RHEL 6.
+        TODO: Remove --noclean detection once nobody is building for RHEL 6.
```
xsuchy: Are we building for RHEL 6?
Edit: Ah, this is for build targets (chroots).

mock/py/mock.py
```diff
@@ -262,6 +262,15 @@ def command_parse():
                            " (i.e. --install). Only useful when using tmpfs plugin.")
     parser.add_option("--nocheck", action="store_false", dest="check",
                       default=True, help="pass --nocheck to rpmbuild to skip 'make check' tests")
+    parser.add_option("--separate-check",
+                      choices=['best_effort', 'enforce', 'off'],
+                      default=None, dest="separate_check",
```
xsuchy, line 267: site-default.cfg says that the default is off.

xsuchy:
Man page uses "enforce" consistently but PR description body says 'on'. Not a code issue, just PR description is slightly off from actual implementation.

mock/py/mockbuild/backend.py
```diff
@@ -712,13 +721,80 @@ def rebuild_installed_srpm(self, spec_path, timeout):
...
+            return False
+
+        if separate_check not in ('best_effort', 'enforce'):
+            raise Error(
+                f"Invalid separate_check value: {separate_check}.  "
+                "Valid values are 'off', 'best_effort', 'enforce'")
+
+        supports_bk = self._rpmbuild_supports_bk
+        if separate_check == 'enforce' and not supports_bk:
+            raise Error(
```
xsuchy: Instead of this, can we have:
```
log.error(...)
sys.exit(NUMBER)
```
and define it in the list of exit codes we already have?

mock/py/mockbuild/backend.py
```diff
@@ -735,12 +811,15 @@ def rebuild_package(self, spec_path, timeout, check, dynamic_buildrequires):
         if additional_opts == ['']:
             additional_opts = []
 
-        def get_command(mode, checkdeps=False):
+        def get_command(mode, checkdeps=False, with_check=False):
             nodeps_opt = [] if checkdeps else ['--nodeps']
+            # check_opt contains --nocheck or equivalent; drop it when we
```
xsuchy: This is ugly. Using variable from outter scope. Can we pass it as a parameter, please?

mock/py/mockbuild/backend.py
```diff
@@ -712,13 +721,80 @@ def rebuild_installed_srpm(self, spec_path, timeout):
                                                                       self.buildroot.builddir)))
         return results[0]
 
+    def _resolve_separate_check(self, check):
+        """
```
xsuchy: IIRC then if config_opts['check'] = False and separate_check = best_effort then the check is run. While it should not.
For better readibility I would add at the top
return False if not check
and then you do need to check it again.

mock/py/mockbuild/backend.py
```diff
@@ -735,12 +811,15 @@ def rebuild_package(self, spec_path, timeout, check, dynamic_buildrequires):
         if additional_opts == ['']:
             additional_opts = []
 
-        def get_command(mode, checkdeps=False):
+        def get_command(mode, checkdeps=False, with_check=False):
             nodeps_opt = [] if checkdeps else ['--nodeps']
+            # check_opt contains --nocheck or equivalent; drop it when we
+            # explicitly want to run %check (separate check phase).
+            effective_check_opt = [] if with_check else check_opt
```
xsuchy: With --nocheck  and separate_check in anything else than off - it will run the check.

xsuchy:
I have two main issues with this PR:

== variable naming

During the review I was VERY confused what is the semantics of every `check` or `with_check` variable. (separate_check was good). I would prefer better naming. Even if that will mean more variables and a bit more code.

== TMPFS and design

Your implementation is (simplified):

* Call rpmbuild without %check
* mount tmpfs over /RPMS/ and /SRPMS/
* call rpmbuild to finish the %check phase
* copy the results

This seems complicated to me. And I see possible attack vectors like umounting tmpfs in rogue %check script and modify binary artifacts.

What I would prefer is:

* Call rpmbuild without %check
* copy the results (copy_build_results())
* call rpmbuild to finish the %check phase

This way the results are already outside of the containers in %check phase so the attack would be MUCH harder. Yes, the code will be trickier as you have to touch `Commands.build()`. But the design will be better. IMO.

mock/docs/mock.1
praiskup: ack, I read as "in normal place", going to try to rephrase

mock/py/mock.py
praiskup: Yeah, off s equivalent to None.

mock/py/mockbuild/backend.py
praiskup: should we define a special exception for this?

praiskup:
> == variable naming

I'll try to do something about it.

> This seems complicated to me. 

I found it quite trivial and safe.

>  And I see possible attack vectors like umounting tmpfs in rogue %check script and modify binary artifacts.

Doubt, can you double-check?  At that point, any reasonable rpmbuild is run non-privileged (not sure if we want to avoid foot-shots, i.e. building under root).
I doubt scriplets can cause harm here either.

praiskup:
> copy the results (copy_build_results())

This seems much, much more complicated (I had a version that did a backup, did chown and chmod for security reasons, and later restored the backed-up permissions, it was *soo ugly*).

praiskup:
> This way the results are already outside of the containers.

Hmm, I see what you mean now 🤔 We aren't always in a container... But since we're not root, it shouldn't be an issue even with --isolation=simple.

Since  `copy_build_results ` is done in later phase, I'll have to do some more refactoring that I probably won't be proud of. So please double-check first and confirm you want this.

mock/py/mockbuild/backend.py
praiskup: Agreed.   Let me refactor the old code.

mock/py/mockbuild/backend.py
praiskup: I don't think the logic is wrong, but let's try to refactor so it is more obvious.

mock/py/mock.py
xsuchy: I know. My point was on unifying the terminology.

mock/py/mockbuild/backend.py
xsuchy: Dunno. `exit()` is fine, special exception handled in mock.py is fine. I care about the result - different exit code.

mock/docs/mock.1
praiskup: fixed

mock/py/mock.py
praiskup: switched to `off` for consistency

mock/py/mockbuild/backend.py
praiskup: :no_good: exit in backend.py isn't optimal

I personally don't need to define new exception+exit, but if you think we need a new exit status we need to modify `exception.py` -- so please suggest the number, and exception name.

Or is BadCmdline OK?

mock/py/mockbuild/backend.py
praiskup: Hm, previously I wanted to make as small change as possible.  I now moved the rpmbuild related stuff into the RpmBuild class...

praiskup: (sorry for rather large change)

mock/py/mockbuild/backend.py
praiskup: I don't think this was the right interpretation, but the current flow is much more obvious, PTAL

mock/py/mockbuild/backend.py
xsuchy: Number 7 is free https://github.com/rpm-software-management/mock/blob/main/mock/py/mockbuild/exception.py#L26 
BadSeparateCheck sounds better - as this is not a general error about the command line

praiskup:
@xsuchy if you can please take one more look ... I'm going to rebase & squash once we have an acceptable variant

mock/py/mockbuild/backend.py
```diff
@@ -27,6 +27,126 @@
 from .rebuild import do_rebuild
 
 
+class RpmBuild:
+    """Thin wrapper for running rpmbuild inside a Mock chroot."""
```
xsuchy: I LOVE this. 
At the same time - this file is already huge and I would like to see it smaller. Can you move this class to a separate file?

mock/py/mockbuild/buildroot.py
```diff
@@ -1079,7 +1079,25 @@ def wrap_host_file(self, filename):
...
+        try:
+            for subdir in ('RPMS', 'SRPMS'):
+                artifact_dir = os.path.join(chroot_builddir, subdir)
+                mount = mounts.FileSystemMountPoint(
+                    path=artifact_dir, filetype='tmpfs',
+                    options='size=1m')
+                mount.mount()
+                mounted.append(mount)
+            yield
+        finally:
```
xsuchy, line 1098: Ouch. This is so hackish. This method is used in "with", so I would expect regular `__enter__()` and `__exit__()`

mock/py/mockbuild/backend.py
```diff
@@ -756,18 +823,12 @@ def get_command(mode, checkdeps=False):
...
-                                            user=self.buildroot.chrootuser,
-                                            unshare_net=self.private_network, raiseExc=False,
-                                            printOutput=self.config['print_main_output'])
+                (output, returncode) = rpmbuild.run_build(
+                    br_mode, raiseExc=False)
                 if returncode > 0 and returncode != 11:
                     # we treat exit status 11 as success, as well as exit
                     # status 0, see issue#434
-                    raise BuildError("Command failed: \n # %s\n%s" % (command, output))
+                    raise BuildError("Command failed: \n # %s\n%s" % (br_mode, output))
```
xsuchy: br_mode is ['-br'] — user sees ['-br'] instead of full rpmbuild command string. Harder to debug.

mock/py/mockbuild/config.py
```diff
@@ -681,6 +682,9 @@ def set_config_opts_per_cmdline(config_opts, options, args):
         config_opts['short_circuit'] = options.short_circuit
         config_opts['clean'] = False
 
+    if options.separate_check:
```
xsuchy, line 685: --separate-check argparse default breaks config file settings.

  In mock.py:264, default='off'. Since 'off' is truthy, config.py:685's guard if options.separate_check: is always True — any config file setting (config_opts['separate_check'] = 
  'best_effort') gets overridden to 'off' by the CLI default. Should be default=None.

mock/docs/mock.1
```diff
@@ -383,6 +383,24 @@ Do not clean chroot before building a package.
...
+\fB\-\-separate\-check\fR \fIMODE\fR
+Run %check as a separate rpmbuild phase (rpmbuild \-bk \-\-short\-circuit)
+after the main build finishes.  This is a security measure \(em built RPM
+artifacts cannot be modified by %check scripts.  During the %check phase,
+the RPMS/ and SRPMS/ directories are hidden under a tmpfs overlay.
+Requires rpm >= 6.0.91.  Valid values for \fIMODE\fR:
+.RS
+.TP
+.B best_effort
+Use separate %check when rpmbuild supports \-bk, fall back to to normal order otherwise.
```
xsuchy: typo: "fall back to to normal order" (double "to").

mock/py/mockbuild/buildroot.py
praiskup: It is a context manager.

mock/py/mockbuild/backend.py
praiskup: I do not intend to fix this, sorry.  That would make the code even more complicated ... or bigger (moving the 0 vs 11 exit code handling down to the RpmBuild).

If anyone cares, there's --verbose output / --trace, etc...

mock/py/mockbuild/config.py
praiskup: ?? please elaborate

mock/py/mockbuild/buildroot.py
praiskup: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

mock/py/mockbuild/config.py
praiskup: Ah, I see (no matter the config value, argparse always wins with the default).  Going back to None (against the previous remark, though)

mock/docs/mock.1
praiskup: fixed

mock/py/mockbuild/backend.py
praiskup: moved

mock/py/mockbuild/buildroot.py
xsuchy: Ouch, so this is "supported". I am used to https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers

mock/py/mockbuild/backend.py
xsuchy: Can I suggest:
Create attribute  `last_command` in the class `rpmbuild.RpmBuild`. 
In the `RpmBuild.run()` store the `command` to last_command.
And here just pass `rpmbuild.last_command`.

mock/py/mockbuild/config.py
xsuchy: Or you can leave it in argparse. And remove the first setting on line 77. 
Hmm, when I look at it for second time, I actually like the current code and take back my comment regarding this line.

mock/py/mockbuild/backend.py
praiskup: Seems OK, thank you for the suggestion.

mock/py/mockbuild/backend.py
praiskup: BadSeparateCheck seems weird, what about BadOption?  ## error in command-line or config ption.

praiskup:
> `_resolve_check_flags()` runs (and logs via `getLog().info`/`build_log.warning`) on _every_ `RpmBuild(...)` construction.

Valid point, fixing.

praiskup:
> In the non-nspawn path, build() has already dropped to the build user before run_separate_check() reaches this context manager, so /bin/mount -t tmpfs can fail with EPERM and break separate_check. Wrap the mount/umount pair with uid_manager.becomeUser(0, 0) / restorePrivs(), like nuke_rpm_db() does.

Valid.

mock/py/mockbuild/backend.py
praiskup: Fixed.

mock/py/mockbuild/buildroot.py
```diff
@@ -1079,7 +1079,25 @@ def wrap_host_file(self, filename):
...
+            for mp in mounted:
+                mp.umount()
```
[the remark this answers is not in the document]
praiskup: Fixed

mock/py/mockbuild/buildroot.py
```diff
@@ -1079,7 +1079,25 @@ def wrap_host_file(self, filename):
...
+        getLog().info("Protecting built artifacts in RPMS/ and SRPMS/ with tmpfs overlay")
+        mounted = []
+        try:
+            for subdir in ('RPMS', 'SRPMS'):
+                artifact_dir = os.path.join(chroot_builddir, subdir)
+                mount = mounts.FileSystemMountPoint(
+                    path=artifact_dir, filetype='tmpfs',
+                    options='size=1m')
+                mount.mount()
+                mounted.append(mount)
```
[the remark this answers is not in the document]
praiskup: fixed

xsuchy (approved):
+1

praiskup:
:facepalm: meh not squashed

PR #1767, opened 2026-07-01
Add cgroup device isolation for build environments

jinzihao:
Enforce a strict cgroup device allowlist (DevicePolicy=closed) on every nspawn invocation by wrapping it in a transient systemd-run --scope with --property=DeviceAllow=... flags, and passing --keep-unit so nspawn stays inside the pre-created restricted scope.

Without this, the container's cgroup inherits 'a *:* rwm' from its parent, meaning root inside the container (e.g. after a privilege escalation exploit) can mknod and open arbitrary host hardware: raw disks (/dev/sda), physical memory (/dev/mem), the KVM hypervisor (/dev/kvm), GPUs, etc.

The allowlist mirrors exactly what mock already deliberately bind-mounts into the chroot — all virtual/interface devices that legitimate builds need:

  - /dev/null, /dev/zero, /dev/full, /dev/random, /dev/urandom  (pseudo)
  - /dev/tty, /dev/console, /dev/ptmx, /dev/pts/*               (terminals)
  - /dev/fuse                 (FUSE mounts: libguestfs, squashfuse, etc.)
  - /dev/btrfs-control        (btrfs-progs test suite)
  - /dev/mapper/control       (lvm2, device-mapper tools)
  - /dev/loop-control         (loop device allocator)
  - block-loop                (/dev/loop0-N: filesystem image tests)
  - block-device-mapper       (/dev/mapper/*: virtual LVM volumes)

Real hardware devices (/dev/mem, /dev/kmem, /dev/kvm, /dev/sd*, /dev/dri/*, /dev/input/*, /dev/net/tun) are deliberately omitted. No legitimate package %check needs them; a build accessing real host hardware is a strong signal of a malicious or severely misconfigured package.

This is analogous to network isolation: rather than blocking all devices, we carve out the specific virtual interfaces builds legitimately need while making real hardware unreachable even to root.

Controlled by config_opts['device_isolation'] (default False). Set to True to enable this additional isolation for builds that do not need raw hardware.

Assisted-By: Qoder

Fixes: #NUMBER

> not deserve a release note.

### DIFF
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -81,6 +81,7 @@ personality_defs = {
 
 USE_NSPAWN = False
 USE_NSPAWN_SECCOMP = False
+_DEVICE_ISOLATION = False
 
 _NSPAWN_HELP_OUTPUT = None
 
@@ -518,6 +519,64 @@ def do(*args, **kargs):
 # logger =
 # output = [1|0]
 # chrootPath
+def _device_isolation_scope_args():
+    """
+    Build the systemd-run --scope arguments for device isolation.
+
+    Instead of racing to write cgroup device policy after nspawn starts
+    (which leaves a window where the payload has full device access),
+    we pre-create a restricted scope using systemd-run.  nspawn then
+    runs inside this scope with --keep-unit, so the payload process is
+    born into an already-restricted cgroup — zero race window.
+
+    On cgroup v1, systemd's DevicePolicy=closed is enforced via BPF_DEVCG
+    (eBPF device control) when the scope is created by systemd-run, but
+    NOT when nspawn creates its own scope via D-Bus.  This approach
+    eliminates that inconsistency.
+
+    On cgroup v2, systemd enforces DevicePolicy natively via eBPF in
+    all cases, so this is equally effective.
+
+    Returns a list of arguments to prepend to the nspawn command,
+    e.g. ['systemd-run', '--scope', '--property=DevicePolicy=closed', ...]
+    or an empty list if device isolation is disabled.
+    """
+    if not _DEVICE_ISOLATION:
+        return []
+
+    args = ['systemd-run', '--scope']
+
+    # DevicePolicy=closed denies all devices by default
+    args.append('--property=DevicePolicy=closed')
+
+    # --- Pseudo-devices: char major 1 ---
+    # Enumerate individually to exclude /dev/mem (1:1) and /dev/kmem (1:2)
+    args.append('--property=DeviceAllow=/dev/null rwm')
+    args.append('--property=DeviceAllow=/dev/zero rwm')
+    args.append('--property=DeviceAllow=/dev/full rwm')
+    args.append('--property=DeviceAllow=/dev/random rwm')
+    args.append('--property=DeviceAllow=/dev/urandom rwm')
+
+    # --- Terminal devices ---
+    args.append('--property=DeviceAllow=/dev/tty rwm')
+    args.append('--property=DeviceAllow=/dev/console rwm')
+    args.append('--property=DeviceAllow=/dev/ptmx rwm')
+    args.append('--property=DeviceAllow=char-pts rw')
+
+    # --- Filesystem interface devices ---
+    args.append('--property=DeviceAllow=/dev/fuse rwm')
+    args.append('--property=DeviceAllow=/dev/btrfs-control rwm')
+    args.append('--property=DeviceAllow=/dev/mapper/control rwm')
+    args.append('--property=DeviceAllow=/dev/loop-control rwm')
+
+    # --- Loop block devices ---
+    args.append('--property=DeviceAllow=block-loop rwm')
+
+    # --- Device-mapper block devices ---
+    args.append('--property=DeviceAllow=block-device-mapper rwm')
+
+    return args
+
 #
 # The "Not-as-complicated" version
 #
@@ -659,8 +718,10 @@ def setup_operations_timeout(config_opts):
 def set_use_nspawn(value, config_opts):
     global USE_NSPAWN
     global USE_NSPAWN_SECCOMP
+    global _DEVICE_ISOLATION
     USE_NSPAWN = value
     USE_NSPAWN_SECCOMP = config_opts["seccomp"]
+    _DEVICE_ISOLATION = config_opts.get('device_isolation', False)
 
 
 class BindMountedFile(str):
@@ -793,7 +854,21 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None,
     if isinstance(cmd, str):
         cmd = ['/bin/sh', '-c', cmd]
 
-    return nspawn_argv + cmd
+    full_cmd = nspawn_argv + cmd
+
+    # When device isolation is active, wrap nspawn in a systemd-run scope
+    # with DevicePolicy=closed.  This pre-creates a restricted scope so
+    # nspawn's payload process is born into an already-restricted cgroup —
+    # no race window where the process has full device access.
+    # nspawn must use --keep-unit to stay in our pre-created scope instead
+    # of creating its own.
+    if _DEVICE_ISOLATION:
+        scope_args = _device_isolation_scope_args()
+        if scope_args:
+            # Insert --keep-unit after the nspawn binary path
+            full_cmd[0:1] = scope_args + full_cmd[:1] + ['--keep-unit']
+
+    return full_cmd
 
 def doshell(chrootPath=None, environ=None, uid=None, gid=None, cmd=None,
             cwd=None,
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -718,7 +718,7 @@ def setup_operations_timeout(config_opts):
 def set_use_nspawn(value, config_opts):
     global USE_NSPAWN
     global USE_NSPAWN_SECCOMP
-    global _DEVICE_ISOLATION
+    global _DEVICE_ISOLATION  # pylint: disable=global-statement
     USE_NSPAWN = value
     USE_NSPAWN_SECCOMP = config_opts["seccomp"]
     _DEVICE_ISOLATION = config_opts.get('device_isolation', False)
--- a/mock/py/mockbuild/util.py
+++ b/mock/py/mockbuild/util.py
@@ -866,7 +866,7 @@ def _prepare_nspawn_command(chrootPath, user, cmd, nspawn_args=None, env=None,
         scope_args = _device_isolation_scope_args()
         if scope_args:
             # Insert --keep-unit after the nspawn binary path
-            full_cmd[0:1] = scope_args + full_cmd[:1] + ['--keep-unit']
+            full_cmd = scope_args + [full_cmd[0], '--keep-unit'] + full_cmd[1:]
 
     return full_cmd
 
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -80,6 +80,15 @@
 # If you're using isolation='nspawn', then by default networking will be turned
 # off for rpmbuild.  This helps ensure more reproducible builds.
 #config_opts['rpmbuild_networking'] = False
+
+# Device isolation.  When enabled, mock wraps nspawn in a systemd-run scope
+# with DevicePolicy=closed so that root inside the container cannot open real
+# host hardware devices (/dev/mem, /dev/kvm, physical disks, GPUs, etc.).  Only
+# virtual/interface devices that legitimate builds need are allowed
+# (/dev/null, /dev/urandom, /dev/fuse, /dev/loop*, /dev/mapper/*, etc.).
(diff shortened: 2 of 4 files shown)


praiskup:
> Real hardware devices (/dev/mem, /dev/kmem, /dev/kvm, /dev/sd*, /dev/dri/, /dev/input/, /dev/net/tun) are deliberately omitted. No legitimate package %check needs them; a build accessing real host hardware is a strong signal of a malicious or severely misconfigured package.

My bet there _are_ legitimate packages that are actually using some those.  Btw., why do you mention `%check` here?  Separate #1766 for the %check details.

Can you start a discussion on Fedora Devel list about this before we start considering this change?
I'm not really in favor of merging this (for the same reasons like in #1768).

praiskup:
See also #1756 and #569.

jinzihao:
> My bet there _are_ legitimate packages that are actually using some those.

I also share this concern with you. So we tested device and network isolation over a Linux distro's full set (1000+) of packages, and observed no compatibility problems. It would be utterly irresponsible if I submit untested changes here which may break everyone's build system.

> Btw., why do you mention `%check` here? Separate #1766 for the %check details.

Because we noticed for the few packages that require device access during build, it mostly occurs during %check phase. It's just a pattern we identified. Nothing more than that.

jinzihao:
> See also #1756

I see its expected behavior is "pivot_root_chroot is silently ignored with nspawn isolation". My proposed changes only work under nspawn isolation, so I guess they won't interfere with each other.

> and #569.

I don't feel it's right to automatically mount /dev/kvm for every build environment according to the principle of least privilege. It might be OK if it's mounted only for Koji's image creation tasks, as they enforce a much weaker isolation than regular build tasks.

Conan-Kudo:
Keep in mind, Mock is deliberately not a security boundary. To turn it into one, we need Mock to spawn VMs for builds instead of using chroots or nspawn.

jinzihao:
> Keep in mind, Mock is deliberately not a security boundary.

Unfortunately, it has been evolving in such a direction since over a decade ago...

> To turn it into one, we need Mock to spawn VMs for builds instead of using chroots or nspawn.

This one is disputable - a VM is definitely not the only thing that's able to establish a security boundary.

jinzihao:
> My bet there _are_ legitimate packages that are actually using some those.

I realized I need to further clarify this - while we did a lot of testing, the golden stardard is definitely not how many packages we've tested on. The real reason is, if any package uses a non-whitelisted device, it almost means exploiting an undefined behavior. Currently, you can never see non-whitelisted devices under /dev - you need to use some hacks to reach those host-side devices. It's very unlikely to have legitimate use cases apart from an illegitimate container escape.

Anyway, this is an optional feature, and we definitely won't make this default behavior for now, so it's very unlikely to break a lot of things.

Conan-Kudo:
> > Keep in mind, Mock is deliberately not a security boundary.
> 
> Unfortunately, it has been evolving in such a direction since over a decade ago...
> 

No, it hasn't. We didn't start using nspawn for security, we started using it to make build environment construction more reliable.

We even explicitly document that Mock isn't intended to provide any kind of security.

> > To turn it into one, we need Mock to spawn VMs for builds instead of using chroots or nspawn.
> 
> This one is disputable - a VM is definitely not the only thing that's able to establish a security boundary.

It is the only _reliable_ boundary for creating a hermetic environment.

jinzihao:
> No, it hasn't. We didn't start using nspawn for security, we started using it to make build environment construction more reliable.

From [docs/Why-Mock.md](https://github.com/rpm-software-management/mock/blob/main/docs/Why-Mock.md): 

> Additionally, building RPM packages is inherently dangerous because specfiles are written in a Turing-complete language. Building an untrusted package can easily lead to a compromised system. This is also true for self-written packages. One small bug like rm -rf %{?nonexisting}/ can wipe your entire system if you are unwise enough to run rpmbuild as root. Or at least your user home directory.
> 
> Mock does many parts of the build process inside a systemd-nspawn container (or at least [chroot](https://en.wikipedia.org/wiki/Chroot)) to avoid such disasters as much as possible.

I feel at least we have to admit there're some design inconsistencies so far...

Conan-Kudo:
Also, we use Mock to do image builds now. There is currently an attempt to try to fix the breakage with nspawn for running image builds in #1749. Legitimately, we may want `/dev/kvm` for that as well.

jinzihao:
> Also, we use Mock to do image builds now. There is currently an attempt to try to fix the breakage with nspawn for running image builds in #1749. Legitimately, we may want `/dev/kvm` for that as well.

Not meant to offend anybody, but why not bind mount / into the container? It'll solve every problem?

Conan-Kudo:
> > Also, we use Mock to do image builds now. There is currently an attempt to try to fix the breakage with nspawn for running image builds in #1749. Legitimately, we may want `/dev/kvm` for that as well.
> 
> Not meant to offend anybody, but why not bind mount / into the container? It'll solve every problem?

Because we still don't want `rm -rf *` to bork the host. Just because it's not a security boundary doesn't mean we want it to allow breaking all the things.

jinzihao:
> > > Also, we use Mock to do image builds now. There is currently an attempt to try to fix the breakage with nspawn for running image builds in #1749. Legitimately, we may want `/dev/kvm` for that as well.
> > 
> > 
> > Not meant to offend anybody, but why not bind mount / into the container? It'll solve every problem?
> 
> Because we still don't want `rm -rf *` to bork the host. Just because it's not a security boundary doesn't mean we want it to allow breaking all the things.

Well, if you only need _safety_ rather than _security_, I don't see why you won't stay with chroot, but opt to get along with all those restrictions of nspawn.

xsuchy:
> I don't see why you won't stay with chroot, but opt to get along with all those restrictions of nspawn.

Because in simple chroot the safety isolation is harder. I.e. the network isolation was done just by deleting resolv.conf, but if any test uses fixed IP address it reached the internet. It is harder to measure resource consumption....

jinzihao:
I guess you're referring to an availability problem, i.e., the A in the [CIA triad](https://en.wikipedia.org/wiki/CIA_triad). I believe currently mock and koji are nowhere close to providing an availability guarantee, which would require robust resource isolation for CPU, memory, disk, etc.. I feel we do strive to ensure confidentiality and integrity, while leaving availability as a non-goal.

mock/py/mockbuild/util.py
```diff
@@ -518,6 +519,64 @@ def do(*args, **kargs):
...
 # chrootPath
+def _device_isolation_scope_args():
+    """
+    Build the systemd-run --scope arguments for device isolation.
+
+    Instead of racing to write cgroup device policy after nspawn starts
+    (which leaves a window where the payload has full device access),
+    we pre-create a restricted scope using systemd-run.  nspawn then
+    runs inside this scope with --keep-unit, so the payload process is
+    born into an already-restricted cgroup — zero race window.
```
praiskup, line 530: I don't think I follow.  Can you describe what happens if we keep using systemd-nspawn, with the same set of arguments?

jinzihao: We indeed implemented that way at first, i.e., using systemd-nspawn with the same set of arguments. However, we tested it and found isolation is not enforced at all on systems using cgroup v1 (likely due to nspawn creating a scope without passing down the device isolation properties, or having them overwritten/shadowed at some stage), while `systemd-run --scope` does work as intended.

BTW, you can see mentioning of a race window - we initially tried to fix it for cgroup v1 by applying device isolation policy immediately after invoking `systemd-nspawn`, but that would leave a race window. Therefore, we switched to `systemd-run --scope`.

praiskup: I'm not sure this answers my question. What is actually happening in nspawn? Does nspawn accept those arguments and ignore them? If it's a bug, could you report it? Wrapping systemd-nspawn with systemd-run seems like a terrible approach...

The note about the race condition window is very confusing. You explained why a workaround doesn't work, but not why the proper solution (running nspawn with the correct set of arguments) fails.

If we fix nspawn (assuming it's broken), we won't need any changes in Mock—since you can already configure custom nspawn arguments there.

jinzihao: > What is actually happening in nspawn? Does nspawn accept those arguments and ignore them?

This in fact touches systemd intricacies deep behind `systemd-nspawn`. When you use plain `systemd-nspawn` without `--keep-unit`, it delegates scope creation to `systemd-machined`, which always creates a scope with `Delegate=yes`, which, with cgroup v1, sets `devices.allow` to `a`, so the devices controller can be handed off to the systemd within the nspawn container. Pretty complicated.

In short, `systemd-nspawn` without `--keep-unit` creates a systemd scope for you. The scope by design allows all devices, which unfortunately shadows the `--property=DevicePolicy=closed` passed from command line. On the contrary, if you create a scope with `systemd-run --scope` and use `--keep-unit` to make `systemd-nspawn` take that scope, you get a scope exactly as you specified - there won't be some opaque mechanisms mangling the scope.

> The note about the race condition window is very confusing.

I agree the comments are indeed confusing - there are intermediate findings no longer relevant to the current version of code. I can clean up them afterwards.

praiskup: > which unfortunately shadows the --property=DevicePolicy=closed passed from command line. 

Is that a bug to be fixed?  Can you report it?

> with cgroup v1

Our Fedora workers are on v2. Could you update, too?

Considering that:
a) you can easily wrap Mock in a container (for real, reasonable isolation),
b) most builders today should already be on v2,
c) nspawn in Mock doesn't guarantee isolation (and won't in the future), and
d) we're likely dealing with an nspawn bug anyway (otherwise see `nspawn_args` config),

I am not for merging.

jinzihao: > Is that a bug to be fixed? Can you report it?

It's not a bug. `systemd-nspawn` simply does not guarantee every `--property=` would be fully passed down. However, they do have a `--keep-unit` option for that. It's fairly consistent.

> Our Fedora workers are on v2.

Mock is itself a standalone piece of software, with use cases beyond building packages for Fedora. If you insist it's only for Fedora, please add an explicit warning for that assumption.

> most builders today should already be on v2.

Please do the survey.

> you can easily wrap Mock in a container (for real, reasonable isolation)

It's not that easy until someone actually implements that for Koji. For the same reason, even if nspawn worked as you expected, writing the correct `nspawn_args` is not easy until you write that into mock's documentation.

(By way of analogy: "You can easily write memory-safe C programs." Theoretically, but not in practice. To make it easy in practice, some guys even created a new language named Rust.)

praiskup: > It's not a bug. systemd-nspawn simply does not guarantee every --property= would be fully passed down.

Is this documented/said somewhere?

> If you insist it's only for Fedora

Not really!  Sounds like we shouldn't bother with cgroups v1 in 2026 in new code (anyone can upgrade builders to EL9+ at least). 
 
> It's not that easy until someone actually implements that for Koji.

Such an effort would seem reasonable IMO, instead of trying to make Mock isolate.

> By way of analogy:

That's not a good analogy for our case.  If you can contribute this patch, it should be trivial to copy-paste the nspawn arguments into your local mock's configuration.  Without pushing us to review & maintain code that we don't need to have upstream.

jinzihao: > Is this documented/said somewhere?

They cannot guarantee a full `--property` pass-through because some `--property` (e.g., `DevicePolicy`) conflicts with its internal working.

> ... review & maintain code that we don't need to have upstream.

As mock selectively exposes a list of devices to the isolated environment, there's a clear intention: things inside the environment shouldn't be able to access devices beyond those. The current implementation is broken, and the "device isolation" is in fact a bugfix -  it should have been the default behavior from day 1. As we forgot to do that on day 1, we add an option today, so the user can switch on the expected behavior in the future.

> anyone can upgrade builders to EL9+ at least
> Such an effort would seem reasonable IMO, instead of trying to make Mock isolate.

Many options out there: let the users find out if they can upgrade to EL9+ and figure out the correct `nspawn_args`, let `systemd-nspawn` change its behavior for cgroup v1, let Koji add support for a non-mock container backend, etc.. Let's face it: the single most reasonable option is to do that in mock. (BTW, I'd love to see Koji supports a non-mock container backend, can you petition for that?)

> If you can contribute this patch, it should be trivial to copy-paste the nspawn arguments into your local mock's configuration.

I've already secured our internal system. However, the whole point of creating a pull request is to help others secure their systems. That's the spirit of open source.

praiskup:
> They cannot guarantee a full --property pass-through because some --property (e.g., DevicePolicy) conflicts with its internal working.

OK, but still - citation needed.

> However, the whole point of creating a pull request is to help others secure their systems.

Thank you for the PR! It won't be forgotten—anyone is welcome to pick it up and use it in their system. If enough Mock users speak up in favor of this feature, we can always reconsider.

Until then, I'm closing this issue. To be clear, it is an intentional design choice that Mock does not provide isolation. Security and isolation must be handled at a different level (with or without this patch, Mock should not be assumed to be safe/isolated). Because of this, the added value for current users is minimal, and the additional complexity (another command wrapper) isn't worth it.

PR #1768, opened 2026-07-01
Add NAT network isolation for build environments

jinzihao:
Implement optional NAT-based network isolation for mock build chroots using Linux network namespaces, veth pairs, and iptables/ip6tables MASQUERADE rules. Builds are executed inside isolated network namespaces with controlled outbound internet access via NAT.

Key components:
- SubnetPool: file-locked allocator that subdivides configurable CIDR blocks into /30 (IPv4) and /64 (IPv6) slices, with dead-process reaping to prevent pool exhaustion.
- NatNetwork: lifecycle manager for a single build's namespace, veth pair, NAT rules, and DNS configuration.
- Integration with buildroot, package manager, backend, and util for consistent network-isolation decisions across all build paths.
- Orphaned resource cleanup (cleanup_orphaned_networks) to reclaim subnets, interfaces, and namespaces after killed processes.

Configuration options (site-defaults.cfg):
- network_isolation: 'loopback' | 'nat' | 'auto'
- network_veth_subnet_block: IPv4 CIDR block for subdivision
- network_veth_subnet_block_v6: optional IPv6 CIDR block
- network_fallback_dns: fallback DNS servers for loopback-only hosts

Includes unit tests for the SubnetPool allocator covering CIDR subdivision, lock-file parsing, PID liveness checks, allocate/release semantics, pool exhaustion, and dead-entry reaping.

Assisted-By: Qoder

Fixes: #NUMBER

> not deserve a release note.

### DIFF
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -82,8 +82,8 @@
 #config_opts['rpmbuild_networking'] = False
 
 # Network isolation mode.  Requires the host to have IPv4 and IPv6 forwarding
-# enabled (e.g. net.ipv4.ip_forward=1 and net.ipv6.conf.all.forwarding=1), which
-# mock will enable automatically when 'nat' or 'auto' is used.  Other system-wide
+# enabled (e.g. net.ipv4.ip_forward=1 and net.ipv6.conf.all.forwarding=1).
+# Mock checks these at startup and fails fast if they are not enabled.  Other system-wide
 # resources used by NAT mode (iptables rules, ip6tables rules, veth interfaces,
 # network namespaces and /run/mock-nat/subnet.lock) are removed when each build
 # command finishes; orphaned resources from killed mock processes are cleaned up
--- a/mock/docs/site-defaults.cfg
+++ b/mock/docs/site-defaults.cfg
@@ -80,6 +80,26 @@
 # If you're using isolation='nspawn', then by default networking will be turned
 # off for rpmbuild.  This helps ensure more reproducible builds.
 #config_opts['rpmbuild_networking'] = False
+
+# Network isolation mode.  Requires the host to have IPv4 and IPv6 forwarding
+# enabled (e.g. net.ipv4.ip_forward=1 and net.ipv6.conf.all.forwarding=1), which
+# mock will enable automatically when 'nat' or 'auto' is used.  Other system-wide
+# resources used by NAT mode (iptables rules, ip6tables rules, veth interfaces,
+# network namespaces and /run/mock-nat/subnet.lock) are removed when each build
+# command finishes; orphaned resources from killed mock processes are cleaned up
+# at the start of each Buildroot.
+#   'loopback' - isolated loopback-only network (default, current behavior)
+#   'shared'   - full host network (legacy, insecure)
+#   'nat'      - isolated namespace with veth + NAT/masquerading
+#   'auto'     - use 'nat' for commands that need network, 'loopback' otherwise
+#config_opts['network_isolation'] = 'loopback'
+# IPv4 CIDR block auto-subdivided into /30 subnets for NAT veth pairs.
+#config_opts['network_veth_subnet_block'] = '192.168.200.0/24'
+# IPv6 CIDR block auto-subdivided into /64 subnets.  Leave empty to disable IPv6.
+#config_opts['network_veth_subnet_block_v6'] = ''
+# Fallback DNS servers when the host's /etc/resolv.conf only has loopback entries.
+#config_opts['network_fallback_dns'] = ['8.8.8.8', '1.1.1.1']
+
 # Additional args for nspawn
 # suppress-sync is added only on system with systemd 250+ (RHEL9, Fedoras)
 # config_opts['nspawn_args'] = ['--capability=cap_ipc_lock']
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -60,6 +60,7 @@ class Commands(object):
         self.no_root_shells = config['no_root_shells']
 
         self.private_network = not config['rpmbuild_networking']
+        self.network_isolation = config.get('network_isolation', 'loopback')
         self.rpmbuild_noclean_option = None
 
         # on-demand buildroot properties
@@ -366,12 +367,23 @@ class Commands(object):
 
         try:
             self.state.start("shell")
+
+            # NAT network isolation for interactive shell
+            nat_network = None
+            if not self.private_network:
+                if self.network_isolation != 'loopback':
+                    from .network import create_nat_network
+                    nat_network = create_nat_network(
+                        self.config, log, self.buildroot.make_chroot_path(),
+                        util.USE_NSPAWN)
+
             ret = util.doshell(chrootPath=self.buildroot.make_chroot_path(),
                                environ=self.buildroot.env, uid=uid, gid=gid,
                                cwd=cwd,
                                nspawn_args=self.config.get("nspawn_args", []),
                                unshare_net=self.private_network,
-                               cmd=cmd)
+                               cmd=cmd,
+                               nat_network=nat_network)
         finally:
             log.debug("shell: unmounting all filesystems")
             self.state.finish("shell")
--- a/mock/py/mockbuild/buildroot.py
+++ b/mock/py/mockbuild/buildroot.py
@@ -145,6 +145,7 @@ class Buildroot(object):
         self._setup_nspawn_devicemapper_device()
         self._setup_nspawn_fuse_device()
         self._setup_nspawn_loop_devices()
+        self._cleanup_orphaned_nat_networks()
 
 
     def set_package_manager(self, fallback=None):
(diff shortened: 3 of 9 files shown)


xsuchy:
NAT is not a security measure. See
https://security.stackexchange.com/a/8773
https://www.f5.com/resources/white-papers/the-myth-of-network-address-translation-as-security
and many others. 

Honestly, I do not see a benefit in this.

jinzihao:
> NAT is not a security measure. See https://security.stackexchange.com/a/8773 https://www.f5.com/resources/white-papers/the-myth-of-network-address-translation-as-security and many others.
> 
> Honestly, I do not see a benefit in this.

I agree NAT isn't something like a firewall. However, using NAT here is to prevent direct access to host network interfaces, a rationale similar to the device isolation. Without NAT, the guest can directly inject packets and tcpdump on host network, effectively becoming an MITM.

jinzihao:
Some background: we did a security analysis for Koji, finding it's entirely possible to fully access (i.e., access as root) host devices and network interfaces from within the build environment. There's no need to exploit any CVE - it's more of a design and logic problem. This means any untrusted package can potentially take over your Koji builder. We believe a simple device and network isolation mechanism can plug this hole, without the need for a disposable VM or a full-fledged container like docker or podman.

praiskup:
>  This means any untrusted package can potentially take over your Koji builder.

Koji is a trusted platform.  Packagers that we trust build from packages that are transitively trusted.  

Check the Mock Scope: https://rpm-software-management.github.io/mock/

> > Mock is neither container nor VM. Mock does some isolation, but it does not aim to be fully isolated.

If we want containers -- mock itself  [can be run in container](https://rpm-software-management.github.io/mock/#mock-inside-podman-fedora-toolbox-or-docker-container) or in a disposable VM (check copr (security) [model](https://docs.pagure.org/copr.copr/developer_documentation/architecture.html)).

Also note, there's the [Hermetic build feature](https://rpm-software-management.github.io/mock/feature-hermetic-builds.html) (== you run in container with --network=none)

jinzihao:
@praiskup Thanks a lot for the valuable comment! As I understand it, Mock has evolved from using a plain chroot to systemd-nspawn-based isolation, which, as you said, "does some isolation". Interestingly, a correctly configured systemd-nspawn actually offers the same level of security as docker or podman - ultimately they're wrappers of kernel capabilities. They're of course weaker than a VM, but at a fraction of the setup and maintenance cost.

Yes, I did look into mock-in-container and Copr earlier. If we have a low-cost, low-friction way for container-level security, that would fall within the comfort zone of a small-to-medium sized team maintaining a Koji system - as a security researcher, that's what I heard back from our engineering team. Luckily, we already have the isolation infrastructure in mock, we only need to harden it a little and we're there.

praiskup:
If not wrapped by a sandbox, Mock escalates privileges to root at some level. A notable event (if a bootstrap image is not used) is the installation of the bootstrap chroot. DNF from the host environment is run as root to install the minimal RPM buildroot and its BuildRequires. Therefore, you need to trust  *that package set* at least.

Considering that running Mock-inside-a-container is trivial these days (for example, Konflux does this), maintaining container-in-Mock use cases seems like a maintenance nightmare. Moreover, keep in mind that OpenShift Kubernetes clusters typically do not allow running containers inside containers..

jinzihao:
> If not wrapped by a sandbox, Mock escalates privileges to root at some level. A notable event (if a bootstrap image is not used) is the installation of the bootstrap chroot. DNF from the host environment is run as root to install the minimal RPM buildroot and its BuildRequires. Therefore, you need to trust _that package set_ at least.

We did notice this in a fairly rigorous security analysis, but quickly moved on to other steps in the build process. It's obviously fine to trust your host-side packages, as they cannot be contaminated by packages built by Koji. Simply speaking, it lies strictly outside the security boundary.

> Considering that running Mock-inside-a-container is trivial these days (for example, Konflux does this), maintaining container-in-Mock use cases seems like a maintenance nightmare. Moreover, keep in mind that OpenShift Kubernetes clusters typically do not allow running containers inside containers..

Well, we already have a container-in-mock. We have a chroot-based container since very early days of mock. And we have a systemd-nspawn-based container for years. I used to believe in the myth that you either use a big-name container like podman, or you have no security at all. But I learned a lot by carefully analyzing how mock works - you just need to correctly configure systemd-nspawn, which mock is already 90% done.

praiskup:
> We did notice this in a fairly rigorous security analysis, but quickly moved on to other steps in the build process. It's obviously fine to trust your host-side packages, as they cannot be contaminated by packages built by Koji.

Mock does the bootstrap chroot installation using _host DNF_ and _RPMs from Koji_.

jinzihao:
I mean at the very beginning of a build process, dnf, rpm and other essential packages are installed into the environment directly using host-side dnf. Those are trusted and should be trusted. At later stages, they facilitate custom package installation that happens inside isolated environments.

praiskup:
I'm not sure what you mean. The RPMs—both the initial set and the packages installed later as build requirements—are installed in exactly the same way, using DNF _from the bootstrap chroot_. The bootstrap chroot itself, however, is created using the host's DNF and `/bin/rpm`.

Both the bootstrap chroot and the buildroot are installed from the same (= Koji) repository, the so-called "buildroot."

jinzihao:
Yes, that's exactly what we see. At some stages, mock executes as root inside the isolated environment, which include dnf-installing build requirements. From a security perspective, root-inside-container is perfectly fine as long as there's adequate isolation - that's the assurance a modern-day container would give you.

praiskup:
No, please double-check how bootstrap chroot installation works.

jinzihao:
After some investigation, I see why I haven't seen the bootstrap chroot installation you mentioned - our production system uses a bootstrap image to speed up things, which in fact closes an important security gap here. But you're right to point out this - if we use the traditional bootstrapping process, the first step is to use host-side dnf to install python3-dnf, python3-dnf-plugins-core and their dependencies from Koji's repo, which means we need to trust that set of packages. But even under this setup, what we need to trust is a small, fixed set of packages, as opposed to any arbitrary package.

jinzihao:
As I understand it, mock invokes dnf 2 or 3 times during a build.

Case 1 - traditional bootstrap: isolation helps in step 2 and 3, leaves step 1 unprotected - need to trust a list of packages.

1. `dnf install python3-dnf python3-dnf-plugins-core`: using host-side dnf in host environment to install a fixed list of packages.
2. `dnf install @build`: using guest-side dnf (obtained from step 1) in an isolated environment to install a koji-specified list of packages.
3. `dnf builddep xxx.src.rpm`: using guest-side dnf (obtained from step 1) in an isolated environment to install a package-specified list of packages.

Case 2 - with bootstrap image: skips step 1, fully isolated - need to trust nothing.

Case 3 - without bootstrap: skips step 1 and moves step 2 and 3 to host environment, no isolation - need to trust any package.

praiskup:
> Builds are executed inside isolated network namespaces with controlled outbound internet access via NAT.
 
I don't see this part, can you better explain how this is done?

I agree with @xsuchy, I don't see the benefit of this, just a lot of maintenance work.  In the era of containers, we should make our life easier by enclosing mock by container, not attempt to reimplement container tooling.

jinzihao:
> > Builds are executed inside isolated network namespaces with controlled outbound internet access via NAT.
> 
> I don't see this part, can you better explain how this is done?

Look at util.py, you can find a `nat_network` argument is added to `do_with_status` and `doshell`, which call into `ChildPreExec.__init__`, passing the `nat_network` and its associated `nat_ns_path`.

Inside `ChildPreExec`, you can find an if-else added to `__call__`. The preexisting behavior is to call into `condUnshareNet`, which either leaves a loopback interface inside the container, or does nothing and leaves the host network interfaces as-is. What I've added here is a third case - enter the pre-created network namespace with `setns`, and configure the guest-side IP addresses and routes with `configure_container_ns`. The configuration process follows a way similar to `condUnshareNet`.

Before diving into the namespace thing, please look at `create_nat_network` in network.py, which creates a `NatNetwork` instance and calls its `setup` function. Note that there're three call sites of `create_nat_network`, namely `Commands.shell`, `BuildRoot.doChroot` and `_PackageManager._execute_mounted`, immediately before calling into `util.doshell`, `util.do_with_status` or `util.do`, which use the network namespace as I just explained.

Now, let's see how `NatNetwork.setup` sets up the namespace for them to use. It creates one using `pyroute2.netns`, creates a virtual network interface, assigns it an IP address, and sets up the iptables forwarding rules. (You'll find this pretty familiar if you've ever turned your Raspberry Pi into a home router.) The end result is a `ns_path`, which is an identifier of this network environment, available for consumption by `setns`. Hence, the loop is closed.

> I agree with @xsuchy, I don't see the benefit of this, just a lot of maintenance work. In the era of containers, we should make our life easier by enclosing mock by container, not attempt to reimplement container tooling.

Honestly speaking, you're very close to the finish line in terms of isolation. However, if you quit the race here, the systemd-nspawn-based isolation introduced in mock 1.4.1, which you've been actively maintaining, together with a sophisticated bootstrapping process, actually offers zero security benefit. It's in fact very carefully designed, and I appreciate it a lot. However, isolation is an all-or-nothing game - you cannot pray some attackers don't know the exact techniques I used to break out of this. I wish I could complete that with a smaller pull request, but we have to use NAT when we need to enable network connectivity while hiding real devices from the guest, so conceptually it's not that complicated.

praiskup:
> if you quit the race here

There's never been any race, since there's no isolation promise.  There are better "dedicated" tools to container isolation, namely Podman.

> systemd-nspawn-based isolation introduced in mock 1.4.1, ... , actually offers zero security benefit.

Our team isn't consistent on this, but I _personally_ dislike the nspawn support in Mock because it gives our users a false promise and potentially multiplies the code's execution paths.  If I could choose, I'd deprecate it, and drop.

jinzihao:
> > systemd-nspawn-based isolation introduced in mock 1.4.1, ... , actually offers zero security benefit.
> 
> Our team isn't consistent on this, but I _personally_ dislike the nspawn support in Mock because it gives our users a false promise and potentially multiplies the code's execution paths. If I could choose, I'd deprecate it, and drop.

I'm not philosophically against this - as a security reseacher, I know many past problems were solved by the deprecation and eventual drop of insecure features. However, as you can see all major browsers still haven't dropped support for http without the s, there can be practical difficulties. Anyway, unlike http, we have an option to fix mock without introducing a breaking change.

xsuchy:
I will close this. It will complicate the code and design and will not bring anything useful.
I am sorry.
To avoid wasting time in coding, I recommend opening an issue first and only creating an MR when there is consensus that such a feature is wanted.

jinzihao:
I do believe this worths more discussion. Let's move to https://github.com/rpm-software-management/mock/issues/1777.

PR #1776, opened 2026-07-13
fix: resolve #1559 — Add new --isolation= mode starting virtual machines

### LINKED ISSUE #1559: Hermetic builds should start a VM instead of a container
Truly hermetic builds require complete isolation from the host, including removal of the influence of the host kernel and system properties. In order to do so, Mock needs to learn how to boot up a lightweight virtual machine, similar to what `obs-build` does in this scenario (and to note, this is the _default_ for `obs-build`).

This would essentially mean adding a third runtime backend (in addition to `simple` and `nspawn`) for KVM, and maybe also a fourth one for plain QEMU (for foreign arch builds).

The code from `obs-build` is fairly straightforward:
* KVM: https://github.com/openSUSE/obs-build/blob/master/build-vm-kvm
* QEMU: https://github.com/openSUSE/obs-build/blob/master/build-vm-qemu

This would also resolve several problems we have with container-based builds where host properties leak and break things (like #952, #691, #1100, #1222, and so on).

Nam0101:
## Summary

fix: resolve #1559 — Add new --isolation= mode starting virtual machines

## Problem

**Severity**: `High` | **File**: `mock/py/mockbuild/isolation.py`

Current isolation limited to simple/nspawn. Need registry entry points for kvm/qemu so Buildroot and CLI resolve new modes without hardcoding.

## Solution

Add IsolationMode constants KVM/QEMU. Map mode string → backend class factory. Fail early if mode needs /dev/kvm or qemu-system-* missing. Keep simple/nspawn paths untouched.

## Changes

- `mock/py/mockbuild/isolation.py` (new)
- `mock/py/mockbuild/config.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
_Note: this change was drafted with AI assistance and reviewed locally before submission._

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -102,6 +102,11 @@ def setup_default_config_opts():
     #    config_opts['nspawn_args'] += ['--suppress-sync=yes']
     config_opts['use_container_host_hostname'] = True
 
+    config_opts['vm_image'] = None
+    config_opts['vm_memory'] = None
+    config_opts['vm_cpus'] = None
+    config_opts['qemu_path'] = None
+
     config_opts['use_bootstrap'] = True
     config_opts['use_bootstrap_image'] = True
     config_opts['bootstrap_image'] = 'fedora:latest'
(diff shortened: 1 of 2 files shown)


praiskup:
Seems like this does nothing, just AI generated nonsense.  Going to close by the end of the week.

Issue #894, opened 2022-04-13
Mock compatibility with DNF5 (and the new microdnf)


xsuchy:
This is mostly a note for myself:

We should test the new microdnf with Mock:
https://fedoraproject.org/wiki/Changes/MajorUpgradeOfMicrodnf
https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-unstable/.

Tracking the missing parts:

- [x] DNF5 [fails to order the initial --installroot transaction properly](https://bugzilla.redhat.com/show_bug.cgi?id=2164378)
- [x] https://github.com/rpm-software-management/dnf5/issues/248
- [x] https://github.com/rpm-software-management/dnf5/issues/249
- [x] https://github.com/rpm-software-management/dnf5/issues/254
- [x] https://github.com/rpm-software-management/dnf5/issues/461 (work-around exists)
- [ ] https://github.com/rpm-software-management/dnf5/issues/112 (blocks cross-arch builds)
- [ ] https://github.com/rpm-software-management/dnf5/issues/146

xsuchy:
This is now mostly about DNF5. FYI we want to do this in the upcoming 3 months.
@j-mracek FYI

praiskup:
See also: https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5

praiskup:
Can be tested with:
```
$ cat ~/.config/mock/fedora-rawhide-dnf5-x86_64.cfg 
config_opts['target_arch'] = 'x86_64'
config_opts['legal_host_arches'] = ('x86_64',)
include('templates/fedora-rawhide.tpl')
config_opts["root"] = "dnf5-test"
config_opts["package_manager"] = "dnf5"
config_opts["dnf5_avoid_opts"] = {"builddep": ["--allowerasing"]}
```

Then `mock -r fedora-rawhide-dnf5-x86_64 <your SRPM>`.

The system /bin/dnf (v4) installs dnf5 into a bootstrap chroot. If you want to test
with DNF5 on host (DNF5 installs even the buildroot), install dnf5 package from [this copr](https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-unstable/).

praiskup:
That ^^ comment was meant to be in the corresponding PR #1027, sorry.

praiskup:
PR #1027 deserves a review, please.  I'll add an integration test for dnf5, but still.

[...]

PR #1778, opened 2026-07-16
No --allowerasing for `dnf5 list`

praiskup:
Relates to: https://github.com/rpm-software-management/mock/issues/894#issuecomment-4982505976


xsuchy (approved):
+1

hroncok:
Just curious, at what point would it make sense to only pass this option explicitly when invoking the install command?

praiskup:
@hroncok someone needs to do that change :face_holding_back_tears: :shrug:

PR #1781, opened 2026-07-20
Fix buildrootlock for OSH

### LINKED ISSUE #1758: buildroot_lock plugin crashes when bootstrap image is unavailable
## Problem

When `--calculate-build-dependencies` is used (which auto-enables the `buildroot_lock` plugin), mock crashes with an unhandled `subprocess.CalledProcessError` if the configured bootstrap image is not locally available — even when bootstrap was not used for the build.

## Reproducer

```bash
# Get a small SRPM
dnf download --source basesystem

# Ensure the bootstrap image is not cached in root's podman storage
sudo podman rmi registry.fedoraproject.org/fedora:42 2>/dev/null || true

# Use the real shipped Fedora 42 config, just disable bootstrap.
# The bug is that buildroot_lock still tries to inspect the bootstrap
# image even though bootstrap was never used.
cat > /tmp/test-buildroot-lock.cfg <<'EOF'
include('/etc/mock/fedora-42-x86_64.cfg')
config_opts['root'] = 'test-buildroot-lock'
config_opts['use_bootstrap'] = False
config_opts['use_bootstrap_image'] = False
config_opts['bootstrap_image_ready'] = False
EOF

# --calculate-build-dependencies auto-enables the buildroot_lock plugin
mock -r /tmp/test-buildroot-lock.cfg --calculate-build-dependencies basesystem-*.src.rpm
```

The build itself succeeds, but mock crashes in the `postdeps` hook when the `buildroot_lock` plugin attempts `podman image inspect` on the bootstrap image (`registry.fedoraproject.org/fedora:42`) that was never pulled.

## Error

```
ERROR: Command '['podman', 'image', 'inspect', '--format',
  '{"pull_digest": "{{ .Digest }}", "id": "{{.Id}}", "architecture": "{{ .Architecture }}"}',
  'registry.fedoraproject.org/fedora:42']' returned non-zero exit status 125.
  File "/usr/libexec/mock/mock", line 1132, in <module>
    exitStatus = main()
  ...
  File ".../mockbuild/plugins/buildroot_lock.py", line 110, in produce_lockfile
    data["bootstrap"] = podman.inspect_hermetic_metadata()
  File ".../mockbuild/podman.py", line 268, in inspect_hermetic_metadata
    res = subprocess.run(cmd, env=self.buildroot.env,
                         stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                         check=True)
subprocess.CalledProcessError: Command '['podman', 'image', 'inspect', ...,
  'registry.fedoraproject.org/fedora:42']' returned non-zero exit status 125.
```

## Analysis

It looks like two things contribute to this:

1. **`bootstrap_image` always exists in config** — it defaults to `'fedora:latest'` (config.py L105), so the condition `if "bootstrap_image" in data["config"]` in buildroot_lock.py L102 evaluates to `True` even when bootstrap wasn't actually used for the build.

2. **Exception type mismatch** — `inspect_hermetic_metadata()` in podman.py calls `subprocess.run(..., check=True)` which raises `subprocess.CalledProcessError` on failure, but the handler in buildroot_lock.py only catches `PodmanError`:

   ```python
   # buildroot_lock.py L107-113
   try:
       podman = Podman(self.buildroot, data["config"]["bootstrap_image"])
       data["bootstrap"] = podman.inspect_hermetic_metadata()
       data["bootstrap"]["image_digest"] = podman.get_oci_digest()
   except PodmanError:
       data["bootstrap"] = {}
   ```

   It seems like `get_oci_digest()` already wraps its errors in `PodmanError`, but `inspect_hermetic_metadata()` does not. This looks similar to the issue fixed for image pulls in ef179b2f ("podman: bail clearly when image pull fails").

## Thoughts on possible fixes

Not sure what the best approach is here — a few ideas, but I'd appreciate guidance from the maintainers:

- The condition in buildroot_lock.py could check `bootstrap_image_ready` rather than just the presence of `bootstrap_image` in config, since the former actually indicates whether the image was prepared.
- `inspect_hermetic_metadata()` could wrap `CalledProcessError` in `PodmanError`, similar to how `get_oci_digest()` already does (and the fix in ef179b2f for image pulls).
- The except clause in buildroot_lock.py could be broadened to also catch `CalledProcessError`.

These aren't mutually exclusive, and there may be a better approach I'm not seeing.

---

*Assisted-By: Claude Code (claude-opus-4-6)*

praiskup:
Fixes: #1758

### DIFF
--- a/mock/py/mockbuild/config.py
+++ b/mock/py/mockbuild/config.py
@@ -718,7 +718,8 @@ def set_config_opts_per_cmdline(config_opts, options, args):
 
     config_opts["calculatedeps"] = options.calculatedeps
     if config_opts["calculatedeps"]:
-        config_opts["plugin_conf"]["buildroot_lock_enable"] = True
+        if "buildroot_lock" not in options.disabled_plugins:
+            config_opts["plugin_conf"]["buildroot_lock_enable"] = True
 
     if options.buildroot_image:  # --buildroot-image option
         if os.path.exists(options.buildroot_image):
--- a/mock/py/mockbuild/plugins/buildroot_lock.py
+++ b/mock/py/mockbuild/plugins/buildroot_lock.py
@@ -99,10 +99,10 @@ class BuildrootLockfile:
                     if cfg_option in self.buildroot.config:
                         data["config"][cfg_option] = self.buildroot.config[cfg_option]
 
-                if "bootstrap_image" in data["config"]:
-                    # Optional object, only if bootstrap image used (we still
-                    # produce lockfiles even if these are useless for hermetic
-                    # builds).
+                # The 'bootstrap' field is omitted when 'use_bootstrap_image'
+                # is False.  Consequently, we cannot perform a subsequent
+                # hermetic build using this lockfile.
+                if self.buildroot.config["use_bootstrap_image"]:
                     with self.buildroot.uid_manager.elevated_privileges():
                         try:
                             podman = Podman(self.buildroot,
--- a/mock/tests/test_buildroot_lock.py
+++ b/mock/tests/test_buildroot_lock.py
@@ -94,7 +94,8 @@ def _mock_vars(rpm_out, repoquery_out):
             (repoquery_out, None),
         ])
     )
-    buildroot.config = EXPECTED_OUTPUT['config']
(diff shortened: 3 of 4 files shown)


FrostyX (approved):
I didn't test but the change looks reasonable to me

PR #1783, opened 2026-07-21
Document oci_platform_map as a temporary workaround

praiskup:
The oci_platform_map feature (commit 577135) works around Podman not being able to auto-detect x86_64 sub-architecture variants in OCI images.  We do not want this workaround to be used unintentionally on non-AlmaLinux hosts — for example, if a Fedora v3 host builds an AlmaLinux v2 target, we want to pull the default (v3) bootstrap image, not force a v2 pull via --platform.

Move the oci_platform_map population into a new build-time-generated almalinux-platforms.tpl that is included by all AlmaLinux templates. On AlmaLinux hosts, mock-core-configs.spec populates this file with the sub-architecture mapping; on other hosts it stays empty.  The default in config.py is now an empty dict.

All related code is marked as temporary, referencing PR #1733. Remove once Podman/Skopeo gain native sub-architecture resolution, or keep as long as we need to handle sub-architecture variants (which may be permanent).

Assisted-By: Claude (claude-opus-4-6)
Relates: #1733

### DIFF
--- a/mock-core-configs/etc/mock/templates/almalinux-10.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-10.tpl
@@ -7,6 +7,8 @@ config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'quay.io/almalinuxorg/almalinux:10'
 config_opts['bootstrap_image_ready'] = True
 
+include('templates/almalinux-platforms.tpl')
+
 # deal with special handling for x86_64_v2 variant
 config_opts['mirrorlist_arch'] = "{% if repo_arch == 'x86_64_v2' %}?arch=x86_64_v2{% endif %}"
 config_opts['baseurl_arch'] = "{% if repo_arch == 'x86_64_v2' %}x86_64_v2{% else %}$basearch{% endif %}"
--- a/mock-core-configs/etc/mock/templates/almalinux-8.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-8.tpl
@@ -6,6 +6,8 @@ config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'quay.io/almalinuxorg/almalinux:8'
 config_opts['bootstrap_image_ready'] = True
 
+include('templates/almalinux-platforms.tpl')
+
 config_opts['dnf.conf'] = """
 [main]
 keepcache=1
--- a/mock-core-configs/etc/mock/templates/almalinux-9.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-9.tpl
@@ -6,6 +6,8 @@ config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'quay.io/almalinuxorg/almalinux:9'
 config_opts['bootstrap_image_ready'] = True
 
+include('templates/almalinux-platforms.tpl')
+
 config_opts['dnf.conf'] = """
 [main]
 keepcache=1
--- a/mock-core-configs/etc/mock/templates/almalinux-kitten-10.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-kitten-10.tpl
@@ -7,6 +7,8 @@ config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['bootstrap_image'] = 'quay.io/almalinuxorg/almalinux:10-kitten'
 config_opts['bootstrap_image_ready'] = True
 
+include('templates/almalinux-platforms.tpl')
+
 # deal with special handling for x86_64_v2 variant
 config_opts['mirrorlist_arch'] = "{% if repo_arch == 'x86_64_v2' %}?arch=x86_64_v2{% endif %}"
 config_opts['baseurl_arch'] = "{% if repo_arch == 'x86_64_v2' %}x86_64_v2{% else %}$basearch{% endif %}"
--- a/mock-core-configs/etc/mock/templates/almalinux-platforms.tpl
+++ b/mock-core-configs/etc/mock/templates/almalinux-platforms.tpl
@@ -1,0 +1,9 @@
+# Temporary workaround for Podman not resolving x86_64 sub-architecture
+# variants in OCI images.  See the discussion in PR #1733 and PR #1783:
+# https://github.com/rpm-software-management/mock/pull/1733
+# https://github.com/rpm-software-management/mock/pull/1783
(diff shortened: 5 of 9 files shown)


praiskup:
@andrewlukoshko would you mind reviewing this?

andrewlukoshko:
@praiskup 
So basically EPEL or Fedora mock package won't allow to build AlmaLinux 10 x86_64_v2 packages on x86_64_v2 hardware. This PR makes it only possible if mock package is built for AlmaLinux but we don't build it.

If we were building our own mock, we'd simply apply the necessary patch ourselves. But we want anyone to be able to build a package for x86_64_v2.

Conan-Kudo:
This change does not make sense. A Fedora host can be v1, v2, or v3 in hardware, and where's the related demand to Podman and Skopeo to support subarch detection? Is there even a plan for them to support it?

You can't call something temporary unless they've agreed to fix it.

Conan-Kudo:
(We also need eventually to add support to Mock to block RHEL on incompatible hosts, because right now there's no way to tell because the images are not published properly for that.)

praiskup:
> So basically EPEL or Fedora mock package won't allow to build AlmaLinux 10 x86_64_v2 packages on x86_64_v2 hardware. 

~~Why so?  It just uses the default, system native, podman image.~~ I got you were talking about v2 HW now.

praiskup:
> But we want anyone to be able to build a package for x86_64_v2.

Here's the point.  In Fedora Copr we use Fedora 44 VMs on modern x86 hardware.  There's no point Mock falling back to v2 images for v2 Alma builds.

praiskup:
> This change does not make sense.

No, from a Mock architectural perspective, the previous change doesn't make sense. There we are just working around: a) broken Podman/Skopeo, and b) broken Alma image labeling (knowing Podman is broken, layout should be adjusted). I just don't want to upset anyone by simply reverting it and forcing the Alma people to iterate on b) (which is, IMHO, the best move here anyway).

praiskup:
Key reasons why I think this patch should be merged:

- Non-v2 hosts continue using the default, hopefully **native**, OCI images for bootstrap regardless of whether they are building for v2, v3, or ppc64le, Alma or any other. (This is the most important point)
- Alma uses the preferred configuration by default.
- Anyone else running on a v2 host can adjust the configuration files to match the host they're running Mock on.

Please take another look.  Well, we can avoid doing the %if fork for Alma, and keep it opt-in for everybody.

Conan-Kudo:
> > But we want anyone to be able to build a package for x86_64_v2.
> 
> Here's the point. In Fedora Copr we use Fedora 44 VMs on modern x86 hardware. There's no point Mock falling back to v2 images for v2 Alma builds.

Well, if we're working off the premise that we want to run in the target environment for the build environment, yes there is.

praiskup:
We are **not** discussing build environment here.  We are discussing bootstrap environment that must contain semantically compatible DNF/RPM stack with the build environment (because otherwise we risk invalid transaction resolution, incompatible feature sets, etc.).  Bootstrap is "native" these days.

praiskup:
@andrewlukoshko Can we revisit this discussion real quick? I'd like to wrap up a new release ASAP.

It would be great if we could move forward without the specfile changes—would you be open to approving that? Everyone on _v2 (including Alma) can configure it. Let me know your preference!

Conan-Kudo:
> We are **not** discussing build environment here. We are discussing bootstrap environment that must contain semantically compatible DNF/RPM stack with the build environment (because otherwise we risk invalid transaction resolution, incompatible feature sets, etc.). Bootstrap is "native" these days.

Yes, but because Mock doesn't know how to identify x86_64 variants, it needs to choose the right image when the host is x86_64. Likewise, we need aliases on v3 system images that "look like" v1 systems so that they can't run on ineligible systems. That functionality is missing right now and needs to exist in order to do the right thing.

Conan-Kudo:
On some level I somewhat agree about bootstrap needing to be native, but without extra support logic, we can't really do that properly either.

praiskup:
Thanks for checking. The design is to use a native bootstrap unless specified otherwise. This works fine for most current users, and it wouldn't be wise to introduce a change for them (i.e., enforcing a v2 bootstrap image on modern machines).

This patch is a compromise that helps v2 users (requiring an opt-in configuration, as a work-around for broken Podman) while leaving all other users unaffected. If you could give it a formal review, that would be awesome!

praiskup:
I'd love to hear your thoughts on the Alma fix in the spec file, @andrewlukoshko :pray:—it basically turns this PR into a no-op for AlmaLinux. However, if you could configure these things in your local x86_64_v2 infra instead, that would be much cleaner.

mock-core-configs/mock-core-configs.spec
```diff
@@ -50,6 +50,23 @@ Rocky Linux and various other specific or combined chroots.
 
 
 %build
+# Temporary workaround for Podman not resolving x86_64 sub-architecture
+# variants in OCI images (PR #1733).  On AlmaLinux hosts we populate
+# almalinux-platforms.tpl with the oci_platform_map so that podman pull
+# uses --platform for sub-arch images.  On other hosts the file stays
+# empty so we always pull the default (host-native) bootstrap image.
+%if 0%{?almalinux}
+cat > etc/mock/templates/almalinux-platforms.tpl <<'EOF'
```
xsuchy: I believe that this change (the whole change in spec file) is not needed.

I agree with the change of moving the `config_opts['oci_platform_map']`. But the content should be there unconditionally. When somebody builds for Alma on Fedora, it should be used as well. The bootstrap chroot should have as closest architecture to the final chroot as possible. Even if that means slower building. But that is a decision of the Alma leads.

xsuchy:
> You can't call something temporary unless they've agreed to fix it.

I understand the "temporary" as "until this is fixed in Podman". Once fixed there, it should not be needed here.

mock-core-configs/mock-core-configs.spec
praiskup: > for Alma on Fedora, it should be used as well

That's actually not true.  Unless the Fedora box is v2 (which it most probably isn't).

praiskup:
We had quite a long discussion with @xsuchy about this—thank you for your time!

While I still disagree with this direction—given that 99% of users are on newer hardware, and still may need to build v2 RPMs, making v2 image downloads unnecessary (and total antipattern)—I respect that Alma configuration is not my responsibility and will honor @andrewlukoshko's wish. I am going to keep oci_platform_map unconditionally hard-coded in the AlmaLinux template(s). Consequently, builds of Alma v2 in Fedora Copr (and everywhere else) will be affected.

xsuchy (approved):
+1

PR #1789, opened 2026-07-30
configs: Fixes for Fedora RISC-V

andreabolognani:
Add the configuration for Fedora 45 on riscv64, which was missing, and fix the branching script so that it will handle the architecture correctly in future invocations.


kashyapc:
FWIW, both changes look good to me.  Thank you!

andreabolognani:
The CI issue might have been caused by the infra instability that Fedora was experiencing yesterday.

praiskup (approved):

andreabolognani:
@praiskup thanks!

PR #1801, opened 2026-08-21
system_monitor: extend the sampler with more data

nikromen:
Old implementation looked up a single nspawn scope, collecting mainly
rpmbuild -bb part. Now it injects a dedicated --slice for the nspawn
container (plus IO accounting anchor to monitor IO), and samples the
whole build (including bootstrap, chroot init, dep resoluition, build
and check).

Also the sampler was extended to collect a lot more data.

Relates #1243

### DIFF
--- a/mock/py/mockbuild/backend.py
+++ b/mock/py/mockbuild/backend.py
@@ -751,7 +751,14 @@ class Commands(object):
         if not calculatedeps:
             checkdeps = dynamic_buildrequires and self.bootstrap_buildroot is not None
             rpmbuild.run_build(mode, checkdeps=checkdeps)
-            rpmbuild.run_separate_check()
+            if rpmbuild.use_separate_check:
+                # these hooks let system_monitor tell %check apart from %build/%install;
+                # only fired when %check actually runs as its own separate phase
+                try:
+                    self.plugins.call_hooks('precheck')
+                    rpmbuild.run_separate_check()
+                finally:
+                    self.plugins.call_hooks('postcheck')
 
         results = glob.glob(bd_out + '/RPMS/*.rpm')
         results += glob.glob(bd_out + '/SRPMS/*.rpm')
--- a/mock/py/mockbuild/plugins/system_monitor.py
+++ b/mock/py/mockbuild/plugins/system_monitor.py
@@ -1,190 +1,176 @@
 # License: GPL2 or later see COPYING
 """
-Track various system statistics during the build phase.
-- Report maximum allocated memory in total and for the "biggest" process
-- Report maximum swap usage
+Sample resource usage continuously throughout the whole build,
+tagging each sample with the build phase active at the time, and
+stream the raw samples into the result dir.
 
-The plugin requires systemd-nspawn as build container runner
+The actual sampling happens in a separate `mock-system-monitor-collector`
+process (see mock-system-monitor-collector.py). This plugin only figures out
+which cgroups to monitor plus is creating named nspawn slice.
 """
 
-import threading
-import os
 import json
-import backoff
+import os
+import subprocess
+import threading
 
+from mockbuild import util
 from mockbuild.trace_decorator import getLog
-from mockbuild.util import get_machinectl_uuid, _safe_check_output, USE_NSPAWN
 
 requires_api_version = "1.1"
+# TODO: would be nice to have monitoring for bootstrap as well
 run_in_bootstrap = False
+
+_COLLECTOR_ARGV = ["mock-system-monitor-collector"]
+
 
 def init(plugins, conf, buildroot):
     """ Plugin entry point """
     SystemMonitor(plugins, conf, buildroot)
 
-class SystemMonitor:
-    """ Main plugin class """
 
-    def get_top_process_info(self, scope_path):
-        """Finds the process in the cgroup with the highest RSS.
+def _own_cgroup_path():
+    # return the cgroupfs path of the current process cgroup as
+    # absolute path (no self-reference). This is where mock itself
+    # is running: builddeps, chroot init, ...
+    try:
+        with open("/proc/self/cgroup", 'r', encoding="utf-8") as file:
+            for line in file:
+                # example line: 0::/some/path
+                if line.startswith("0::"):
+                    relative = line.strip().split(":", 2)[2]
+                    return "/sys/fs/cgroup" + relative
+    except OSError as e:
+        getLog().debug("SYSMON: could not read own cgroup: %s", e)
 
-        Args:
-            scope_path (str): Path to the cgroup scope directory.
+    return None
 
-        Returns:
-            tuple[int, str]: A tuple containing the RSS of the top process in bytes
-                and its command line.
-        """
-        procs_path = os.path.join(scope_path, "cgroup.procs")
-        max_rss = 0
-        max_cmdline = ""
 
-        try:
-            if not os.path.exists(procs_path):
-                return (0, "unknown")
-            with open(procs_path, 'r', encoding="utf-8") as f:
-                pids = f.read().split()
+def _setup_cgroup():
+    cgroup_path = _own_cgroup_path()
+    if not cgroup_path or not os.path.isdir(cgroup_path):
+        return None
 
-            for pid in pids:
-                try:
-                    with open(f"/proc/{pid}/statm", 'r', encoding="utf-8") as sm:
-                        data = sm.read().split()
-                        if not data:
-                            continue
-                        # RSS in pages * PAGE_SIZE bytes
-                        rss_bytes = int(data[1]) * os.sysconf('SC_PAGE_SIZE')
+    return cgroup_path
 
-                    if rss_bytes > max_rss:
-                        max_rss = rss_bytes
-                        with open(f"/proc/{pid}/cmdline", 'r', encoding="utf-8") as cmd:
-                            max_cmdline = cmd.read().replace('\0', ' ').strip()
-                        if not max_cmdline:
-                            max_cmdline = f"{pid}"
 
-                except (FileNotFoundError, ProcessLookupError, IndexError):
-                    continue
-        except OSError as e:
-            getLog().error("SYSMON: Error: %s", e)
+def _inject_nspawn_slice(config, slice_name):
+    # without a dedicated slice, the actual build (running inside
+    # the nspawn container) would be invisible to us - monitoring
+    # only mock's own cgroup would be misleading, not useful.
+    # This is where the building happens.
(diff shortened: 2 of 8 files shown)


nikromen:
overall stats on project hyprpicker seems promising on VM with 2cpu:

```
{
    "phase": "overall",
    "duration_seconds": 42.07338738441467,
    "peak_rss_gb": 0.7924156188964844,
    "swap_peak_gb": 0.0,
    "mean_rss_gb": 0.33376711065118964,
    "p95_rss_gb": 0.7165908813476562,
    "sustained_high_rss_pct": 0.13636363636363635,
    "cpu_user_seconds": 43.749225,
    "cpu_system_seconds": 3.328693,
    "cpu_pct_mean": 1.0678,
    "cpu_pct_max": 1.978,
    "io_read_gb": 0.0,
    "io_write_gb": 0.0,
    "io_read_iops_total": 0,
    "io_write_iops_total": 0,
    "psi_cpu_total_usec": 1273944,
    "psi_mem_total_usec": 0,
    "psi_io_total_usec": 0,
    "sample_count": 22,
    "sample_interval_seconds": 2
}
```

I/O seems to be not working rn, but otherwise the approach looks promising

mock/py/mockbuild/plugins/system_monitor.py
```diff
@@ -1,190 +1,483 @@
...
+def _stop_io_accounting_anchor(proc):
+    if proc is None:
+        return
+
+    # just kill the sleep infinity
+    try:
+        proc.kill()
+        proc.wait(timeout=5)
+    except (OSError, subprocess.TimeoutExpired):
+        pass
```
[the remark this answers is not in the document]
nikromen: --scope execs directly into the target command (same PID, no fork+wait). Killing it does terminate the scope:

```bash
[jkyjovsk@fedora ~]$ systemd-run --user --scope -- sleep 60 & p=$!; sleep 1; ps -o pid,cmd -p $p; kill -9 $p; sleep 1; ps -p $p
[1] 55508
Running as unit: run-p55508-i30926.scope; invocation ID: 740062d66274476da21e589d5fcf4f36
    PID CMD
  55508 /usr/bin/sleep 60
[1]+  Killed                     systemd-run --user --scope -- sleep 60
    PID TTY          TIME CMD
[jkyjovsk@fedora ~]$
```

nikromen:
I am satisfied witth the metrics that the plugin produces, last commit's docs was generated with AI so that need check from my side later, however I think this is good to go for review

mock/py/mockbuild/backend.py
```diff
@@ -751,7 +751,14 @@ def rebuild_package(self, spec_path, dynamic_buildrequires):
...
             rpmbuild.run_build(mode, checkdeps=checkdeps)
-            rpmbuild.run_separate_check()
+            if rpmbuild.use_separate_check:
+                # these hooks let system_monitor tell %check apart from %build/%install;
+                # only fired when %check actually runs as its own separate phase
+                try:
+                    self.plugins.call_hooks('precheck')
+                    rpmbuild.run_separate_check()
+                finally:
+                    self.plugins.call_hooks('postcheck')
```
praiskup, line 761: What's this for?

mock/py/mockbuild/backend.py
```diff
@@ -751,7 +751,14 @@ def rebuild_package(self, spec_path, dynamic_buildrequires):
         if not calculatedeps:
             checkdeps = dynamic_buildrequires and self.bootstrap_buildroot is not None
             rpmbuild.run_build(mode, checkdeps=checkdeps)
-            rpmbuild.run_separate_check()
+            if rpmbuild.use_separate_check:
+                # these hooks let system_monitor tell %check apart from %build/%install;
+                # only fired when %check actually runs as its own separate phase
+                try:
+                    self.plugins.call_hooks('precheck')
```
praiskup, line 758: Is it wise to have a hook that is only ever called when separate check is activated?  If you really need this, please document the reasons (these are not obivous).

praiskup:
Do we need to hardwire the monitoring logic directly in a plugin?  Heck, this is a great level of generated code, and too long to read properly now.

Imagine a minimalistic SystemMonitor plugin periodically executing a python binary, ideally without `root` :pray:

- no input
- output is json with `key => value` pairs
- mock just executes the script, gathers the output and combines it into a larger dict??  or catenates it into a predefined file text file no need to wait till the end to write the file, but in Copr we have also a logic for streamed JSON, memory consumption)
- users can provide their own monitoring logic (according to the specified api)
- if there's any traceback, be there -- build continues, we are just loosing the snapshot

nikromen:
> Do we need to hardwire the monitoring logic directly in a plugin?

no, not all of it.

> and too long to read properly now.

yes I see that the final version of plugin had too much in it. I'll split it into multiple PR phases then... this would be the core one: creating --slice for this mock process and reading correct cgroup files at right locations; periodically dumping info into samples.jsonl. Next steps would just add more data from the cgroup; summarizing the samples; extending it with IO samples (this one was tricky).

currently the plugin writes swap, CPU, RAM usage (and pid count) as a core.

> ideally without root

yes that works

> no input

not sure what that means? the plugin doesn't take any input either currenlty?

> output is json with key => value pairs

yes

> mock just executes the script, gathers the output and combines it into a larger dict?

basically how it is done, just in a small difference: it's in-process function call on every tick (a background thread), not separately executed binary - same as the original implementation

> Copr we have also a logic for streamed JSON

no need in here, it will be few MB at most even for extremely long builds

> users can provide their own monitoring logic

I think this would be hard to implement? the current plugin implementation lacks this as well

> if there's any traceback, be there -- build continues, we are just loosing the snapshot

most of the work is done in python background thread (_run) - if that fail, the thread just dies quietly and build continues. _finalize is risky too due to IO, so I wrapped it in try-except. The only part that could possibly crash is SystemMonitor.__init__

mock/py/mockbuild/backend.py
nikromen: The plugin itself tags in which phase it was measuring. This was added to be able to distinguish check phase from others, otherwise %check samples would show up as build.

Fair point though - if you'd rather not create a hook in core that exists for one plugin's benefit, I am fine with dropping whis - worst case the plugin just tags %check phase as build (which is a shame since tagging rest of phases works)

nikromen:
one more note - it;s fair to say that the idea/mechanics for tracking this data is almost same with the original implementation.... this ended up looking more like a rewrite mainly for those reasons:

- the original captured nspawn scope once via machinectl, which only works for single nspawn infocation - a real issue, since mock execs nspawn instance per command. That is replaced with mock own cgroup + persistend name `--slice`. This tracks every nspawn invocations for the whole build, thus we are not losing any data.
- the reading logic needed (and will need) generic helpers instead of one-off code per field - otherwise if we want track more metrics it would be bloated
- also switched to reading cgroup directly via `/proc/self/cgroup`, instead of tracing it -> no dbus/systemctl query and no polling loop needed then, since kernel exposes the file already

but the approach remains basically the same: measuring something through cgroups, injecting something into `nspawn_args` to make it possible, running it as background thread and hooking via add_hook

praiskup:
> basically how it is done, just in a small difference: it's in-process function call on every tick (a background thread), not separately executed binary - same as the original implementation

Since we significantly rewrite now anyway, it would be nice to make it a self standing script right now. '

> not sure what that means? the plugin doesn't take any input either currenlty?

I just underlined that the script should expect no input.  So you get it right.

> no need in here, it will be few MB at most even for extremely long builds

The memory consumption is one thing, but the major benefit is that you get some data even though the script dies in in the middle of writing (because, I don't know, OOM?).  For that reason, I would prefer some CSV format over JSON anyway.

> most of the work is done in python background thread (_run) - if that fail, the thread just dies quietly and build continues.

And I don't quite like that approach, to be honest (e.g., sys._exit()).  And then the benefits like "I can concifure my own data collector..., list of monitored fields.... etc.".

> it;s fair to say that the idea/mechanics for tracking this data is almost same with the original implementation

I don't disagree.  But it's fair to let the original author know, @yaneti  -- feedback is welcome.

yaneti:
...
> > it;s fair to say that the idea/mechanics for tracking this data is almost same with the original implementation
> 
> I don't disagree. But it's fair to let the original author know, @yaneti -- feedback is welcome.

Glad someone else is extending the concept.  I am not really married to anything about the original design :) it was just a first attempt at getting the most pertinent information that I needed at the time.

nikromen:
@praiskup I rewrite it into minimalistic plugin that executes the monitoring script as agreed

> I would prefer some CSV format over JSON anyway.

I still think JSONL is better: JSONL already survives partial writes the same as CSV, CSV wouldn't improve that. Also it adds self-describing keys we don't need a fixed column schema across the files to hold as a context - this will be future proof as more metrics will be added. So we don't risk column shift, misalign or silently wrong data. With JSONL each line just has whatever fields it has, no ambiguity. But if you really insist on CSV, I'l change that

also these still needs to be inside the plugin:
```py
        cgroup_path = _setup_cgroup()
        if not cgroup_path:
            getLog().warning("SYSMON: could not determine own cgroup, "
                              "resource monitoring disabled")
            return

        slice_name = f"mocksysmon{os.getpid()}.slice"
        if not _inject_nspawn_slice(buildroot.config, slice_name):
            getLog().warning("SYSMON: systemd-nspawn does not support --slice, "
                              "resource monitoring disabled")
            return
```

since we need to read mock's own `/proc/self/cgroup` - it can be easily discovered on its own instead of complicated tracing of the correct mock cgroup in `/proc/fs/cgroup`... these will contain everything mock does (builddep, init, ...)

the named slice needs to be injected to nspawn args before any nspawn command, so that still needs to be in the plugin. This way we can collect through this cgroup directory what is hapenning during all of the rpmbuild phases.

nikromen:
@yaneti this PR is missing https://github.com/rpm-software-management/mock/blob/main/mock/py/mockbuild/plugins/system_monitor.py#L28 functionality, but I'll add it in next PR phases (to make this review and other reviews easier). I just want to ask whether you are using this? Or anyone using this metric? Or it was just an experiment

yaneti:
@nikromen  Finding the process with the maximum RSS during the build was the main reason for coming up with the plugin.  It makes it easy to know that there is absolutely no reason to try a build on e.g. a 8G RAM builder, when a single rust process takes 16G+

nikromen:
understood, I will add it later then. I was just wondering whether it was leftover after perhaps adding more accurate kernel's absolute memory peak from `memory.peak`, since max memory usage from 1 process would be typically less than that, so I was wondering whether it was still needed. But I see your point that you find this information useful

Issue #1805, opened 2026-08-27
Unbreq plugin fails when building f45 `mesa`


luk1337:
Repro steps:
fedpkg clone mesa
cd mesa
fedpkg mockbuild -- --enable-plugin=unbreq

Traceback:
```
Traceback (most recent call last):
  File "/usr/libexec/mock/mock", line 1140, in <module>
    exitStatus = main()
  File "/usr/libexec/mock/mock", line 892, in main
    result = run_command(options, args, config_opts, commands, buildroot)
  File "/usr/libexec/mock/mock", line 1014, in run_command
    mockbuild.rebuild.do_rebuild(config_opts, commands, buildroot, options, srpms)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.15/site-packages/mockbuild/rebuild.py", line 84, in do_rebuild
    rebuild_generic(srpms, commands, buildroot, config_opts, cmd=build,
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                    post=post_build, clean=clean)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.15/site-packages/mockbuild/rebuild.py", line 25, in rebuild_generic
    ret = cmd(item)
  File "/usr/lib/python3.15/site-packages/mockbuild/rebuild.py", line 66, in build
    commands.build(srpm, spec=options.spec)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.15/site-packages/mockbuild/backend.py", line 315, in build
    results = self.rebuild_package(spec_path, dynamic_buildreqs)
  File "/usr/lib/python3.15/site-packages/mockbuild/backend.py", line 736, in rebuild_package
    self.plugins.call_hooks('postdeps')
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/lib/python3.15/site-packages/mockbuild/plugin.py", line 81, in call_hooks
    hook(*args, **kwargs)
    ~~~~^^^^^^^^^^^^^^^^^
[...]

PR #1806, opened 2026-09-01
Fix logic errors in unbreq plugin

mkoncek:
These were found when inspecting https://github.com/rpm-software-management/mock/issues/1805 and are not related to that issue.

I believe these fixes don't need a release note because they are difficult to describe and we could simply say that we fixed generic logic errors.


xsuchy:
+1

